PR #5, opened 2017-10-26
Fix response size for error messages in system_report case.

timopollmeier:
In the gsad.c:handle_request function an invalid response size was used
 in cases where get_system_report_omp does not set it.
Now the string length is used in these cases.


bjoernricks:
@timopollmeier is this already fixed in master?

PR #9, opened 2017-10-27
Trashcan added

swaterkamp:
The trashcan page is added with corresponding new api and icons.

### DIFF
--- a/src/html/classic/ng/src/gmp/commands/trashcan.js
+++ b/src/html/classic/ng/src/gmp/commands/trashcan.js
@@ -1,0 +1,186 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+ *
+ * Copyright:
+ * Copyright (C) 2017 Greenbone Networks GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import {HttpCommand, register_command} from '../command.js';
+import {map} from '../utils.js';
+
+import CollectionList from '../collection/collectionlist.js';
+
+import Agent from '../models/agent.js';
+import Alert from '../models/alert.js';
+import Scanconfig from '../models/scanconfig.js';
+import Credential from '../models/credential.js';
+import Filter from '../models/filter.js';
+import Group from '../models/group.js';
+import Note from '../models/note.js';
+import Override from '../models/override.js';
+import Permission from '../models/permission.js';
+import PortList from '../models/portlist.js';
+import ReportFormat from '../models/reportformat.js';
+import Role from '../models/role.js';
+import Scanner from '../models/scanner.js';
+import Schedule from '../models/schedule.js';
+import Tag from '../models/tag.js';
+import Target from '../models/target.js';
+import Task from '../models/task.js';
+
+class Trashcan extends HttpCommand {
+
+  restore({id}) {
+    const data = {
+      cmd: 'restore',
+      target_id: id,
+    };
+    return this.httpPost(data);
+  }
+
+  delete({id, entity_type}) {
+    const command = 'delete_trash_' + entity_type;
+    const type_id = entity_type + '_id';
+    const data = {
+      cmd: [command],
+      [type_id]: id,
+    };
+    return this.httpPost(data);
+  }
+
+  empty() {
+    return this.httpPost({cmd: 'empty_trashcan'});
+  }
+
+  get() {
+    return this.httpGet({cmd: 'get_trash'}).then(response => {
+      const trash_data = response.data.get_trash;
+
+      const agent_list = new CollectionList(
+        {
+          entries: map(trash_data.get_agents_response.agent,
+          model => new Agent(model)),
+        });
+      const alert_list = new CollectionList(
+        {
+          entries: map(trash_data.get_alerts_response.alert,
+          model => new Alert(model)),
+        });
+      const scan_config_list = new CollectionList(
+        {
+          entries: map(trash_data.get_configs_response.config,
+          model => new Scanconfig(model)),
+        });
+      const credential_list = new CollectionList(
+        {
+          entries: map(trash_data.get_credentials_response.credential,
+          model => new Credential(model)),
+        });
+      const filter_list = new CollectionList(
+        {
+          entries: map(trash_data.get_filters_response.filter,
+          model => new Filter(model)),
+        });
+      const group_list = new CollectionList(
+        {
+          entries: map(trash_data.get_groups_response.group,
+          model => new Group(model)),
+        });
+      const note_list = new CollectionList(
+        {
+          entries: map(trash_data.get_notes_response.note,
+          model => new Note(model)),
+        });
+      const override_list = new CollectionList(
+        {
+          entries: map(trash_data.get_overrides_response.override,
+          model => new Override(model)),
+        });
+      const permission_list = new CollectionList(
+        {
+          entries: map(trash_data.get_permissions_response.permission,
+          model => new Permission(model)),
+        });
+      const port_list_list = new CollectionList(
+        {
+          entries: map(trash_data.get_port_lists_response.port_list,
+          model => new PortList(model)),
+        });
+      const report_format_list = new CollectionList(
+        {
+          entries: map(trash_data.get_report_formats_response.report_format,
+          model => new ReportFormat(model)),
+        });
+      const role_list = new CollectionList(
+        {
+          entries: map(trash_data.get_roles_response.role,
+          model => new Role(model)),
+        });
+      const scanner_list = new CollectionList(
+        {
+          entries: map(trash_data.get_scanners_response.scanner,
+          model => new Scanner(model)),
+        });
+      const schedule_list = new CollectionList(
+        {
+          entries: map(trash_data.get_schedules_response.schedule,
+          model => new Schedule(model)),
+        });
+      const tag_list = new CollectionList(
+        {
+          entries: map(trash_data.get_tags_response.tag,
+          model => new Tag(model)),
+        });
+      const target_list = new CollectionList(
+        {
+          entries: map(trash_data.get_targets_response.target,
+          model => new Target(model)),
+        });
+      const task_list = new CollectionList(
+        {
+          entries: map(trash_data.get_tasks_response.task,
+          model => new Task(model)),
+        });
+
+      return response.setData({
+        agent_list,
+        alert_list,
+        scan_config_list,
+        credential_list,
+        filter_list,
+        group_list,
+        note_list,
+        override_list,
+        permission_list,
+        port_list_list,
+        report_format_list,
+        role_list,
+        scanner_list,
+        schedule_list,
+        tag_list,
+        target_list,
+        task_list,
+      });
+    });
+  }
+}
+
+register_command('trashcan', Trashcan);
+
+// vim: set ts=2 sw=2 tw=80:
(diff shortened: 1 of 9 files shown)


src/html/classic/ng/src/web/pages/extras/trashactions.js
```diff
@@ -0,0 +1,181 @@
...
+import {is_defined} from 'gmp/utils.js';
+import PropTypes from '../../utils/proptypes.js';
+
+const TrashActions = ({
+    entity,
+    onEntityDelete,
+    onEntityRestore,
+  }) => {
+
+    const check_by_type = {
```
bjoernricks: Please move the check_by_type object out of TashActions and pass the entity as a parameter to the functions. They object is generated on each render and should only be generated once because it is static.

bjoernricks: Please use arrow functions instead.
```javascript
agent: entity => {
  return {restorable, deletable: entity.isInUse()};
},
```

src/html/classic/ng/src/web/pages/extras/trashactions.js
```diff
@@ -0,0 +1,181 @@
...
+  return (
+    <TableData>
+      <IconDivider
+        align={['center', 'center']}
+        grow
+      >
+        <RestoreIcon
+          name="restore"
+          value={entity}
+          {...mode().restoremode}
```
bjoernricks: The mode function is called twice here but you only use one of its returned values. You should either assign its return value to a variable or split it into two different functions one for getting the restoremode and one for getting the deletemode.

src/html/classic/ng/src/web/pages/extras/trashactions.js
```diff
@@ -0,0 +1,181 @@
...
+        const restorable = !entity.schedule.isInTrash() &&
+          !entity.target.isInTrash() &&
+          !entity.config.isInTrash() &&
+          !entity.scanner.isInTrash() &&
+          !entity.alerts.some(alert => alert.isInTrash());
+        return {restorable, deletable: false};
+      },
+    };
+
+    function mode() {
```
bjoernricks: As above. The function is created for each render. So move it above the component and pass entity as a parameter.

src/html/classic/ng/src/web/pages/extras/trashactions.js
bjoernricks: I would split this into two checks. One for checking if the entity is restoreable and one for checking if the entity is deletable.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+import TargetsTable from '../targets/table.js';
+import TasksTable from '../tasks/table.js';
+
+import Table from '../../components/table/strippedtable.js';
+import TableBody from '../../components/table/body.js';
+import TableData from '../../components/table/data.js';
+import TableRow from '../../components/table/row.js';
+import TableHead from '../../components/table/head.js';
+import TableHeader from '../../components/table/header.js';
+import {Col} from 'glamorous';
```
bjoernricks: Please move the glamorous import to the top after the react import.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+* You should have received a copy of the GNU General Public License
+* along with this program; if not, write to the Free Software
+* Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import React from 'react';
+
+import _ from 'gmp/locale.js';
+
+import TrashActions from './trashactions.js';
```
bjoernricks: Please move the import to the bottom as the last import. Generic pattern for imports is currently:
first react, afterwards external components, then gmp imports then web imports, last imports from same current dir.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+
+import InnerLink from '../../components/link/innerlink.js';
+import LinkTarget from '../../components/link/target.js';
+
+import Button from '../../components/form/button.js';
+import HelpIcon from '../../components/icon/helpicon.js';
+
+import Loading from '../../components/loading/loading.js';
+
+import {map, is_defined} from 'gmp/utils.js';
```
bjoernricks: Move to the other gmp imports.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+    let scanner_list;
+    let schedule_list;
+    let tag_list;
+    let target_list;
+    let task_list;
+
+    if (!is_defined(trash_data)) {
+      return <Loading/>;
+    }
+    if (is_defined(trash_data)) {
```
bjoernricks: You can remove this if check. The render method is already returned if trash_data is not defined. Therefore this if condition is always true.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+      this.getTrash();
+  }
+
+  getTrash() {
+    const {gmp} = this.props;
+    const data = Promise.all([
+      gmp.trashcan.readTrashcan().then(response => {
+        const trash_data = response.data;
+        const contents_table = this.createContentsTable(trash_data);
+        this.setState({trash_data, contents_table});
```
bjoernricks: Please don't put a react element into the state. The react element should only be created in the render method.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+        </TableData>
+      </TableRow>
+    );
+  }
+
+  createContentsTable(trash) {
+    return (
+      <TableBody>
+        {this.createContentRow('Agents',
+          trash.get_agents_response.agent_count.toString())}
```
bjoernricks: You should improve your Command class that the response already contains a better data structure. You need the counts therefore it should only contain the counts here: tash.get_xyz_response is unnecessary. tash.agent_count would be the better api. Maybe even name it tash.agents_count.

bjoernricks: Btw. toString() isn't needed at all. It's already a string. Also converting from number to string wouldn't be necessary here. And converting to string would be better with '' + value

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
bjoernricks: Creating the CollectionLists shouldn't be done in the render method. The render method should only create and render react components. Creating the CollectionsLists better belongs to the command class.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
```diff
@@ -0,0 +1,493 @@
...
+      </Layout>
+    );
+  }
+};
+
+Trashcan.propTypes = {
+  gmp: PropTypes.gmp.isRequired,
+};
+
+export default withContext({
```
bjoernricks: you can replace this with withGmp HOC now. I have added a withGmp HOC some weeks ago.
```javascript
export default withGmp(Tashcan);
```

src/html/classic/ng/src/web/components/bar/menubar.js
```diff
@@ -229,8 +229,8 @@ const MenuBar = (props, {gmp, capabilities}) => {
         </Menu>
         <Menu title={_('Extras')}>
           <MenuEntry
-            legacy
             title={_('Trashcan')}
+            to="trashcan"
             cmd="get_trash"/>
```
bjoernricks: cmd="get_tash" property can be removed

src/html/classic/ng/src/gmp/commands/trashcan.js
```diff
@@ -0,0 +1,61 @@
...
+    const command = 'delete_trash_' + entity_type;
+    const type_id = entity_type + '_id';
+    const data = {
+      cmd: [command],
+      [type_id]: id,
+    };
+    return this.httpPost(data);
+  }
+
+  emptyTrashcan() {
```
bjoernricks: Maybe rename it to only empty() ? The api will be gmp.trashcan.empty() which is nicer then gmp.tashcan.emptyTashcan()

Maybe also gmp.tashcan.flush ?

src/html/classic/ng/src/gmp/commands/trashcan.js
```diff
@@ -0,0 +1,61 @@
...
+    return this.httpPost({cmd: 'empty_trashcan'});
+  }
+
+  readTrashcan() {
```
bjoernricks: I haven't used a readXyz api name yet. Therefore IMHO only get would be better here. gmp.tashcan.get()

src/html/classic/ng/src/web/pages/extras/trashactions.js
bjoernricks: Also the name mode is a bit confusing because it is about props for the icons. I would expect something like get_delete_props(entity) and get_restore_props(entity) instead of mode. If you still would like to use one function get_delete_restore_props(entity) would be a good name. The api could be
```javascript
const {deleteprops, restorepros} = get_delete_restore_props(entity);
return (
...
  <RestoreIcon
    ...
    {...restoreprops}
  />
...
);
```

bjoernricks:
Summarized in the main your code is nice. It only needs some adjustments.

src/html/classic/ng/src/gmp/commands/trashcan.js
swaterkamp: Changed it to "empty()" in order to stay closer to the original naming of "empty_trashcan" http-command.

src/html/classic/ng/src/gmp/commands/trashcan.js
swaterkamp: The api is now gmp.trashcan.get()

src/html/classic/ng/src/web/components/bar/menubar.js
swaterkamp: Done

src/html/classic/ng/src/web/pages/extras/trashactions.js
swaterkamp: Moved check_by_type out of TrashActions, unscrambled the logics for "deletable", and rewrote to use arrow functions.

src/html/classic/ng/src/web/pages/extras/trashactions.js
swaterkamp: Moved it out of TrashActions. This made it necessary, though, to also pass onEntityDelete and onEntityRestore as parameters.

src/html/classic/ng/src/web/pages/extras/trashactions.js
swaterkamp: I renamed the function but sticked to the single function solution. Both checks have to be performed anyway and by keeping both within one function, the source code can be shorter.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: Resorting done

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: done

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: done

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: createContentsTable() is now a function which gets called from the render() method, without the detour through the state.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: Done

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: If-check is removed and the creation of the CollectionLists is done in the command class now.

src/html/classic/ng/src/web/pages/extras/trashcanpage.js
swaterkamp: Done

bjoernricks:
Nice thanks!

PR #10, opened 2017-10-28
Add a State component

bjoernricks:
The new State component uses the render props pattern to store a state.
The state can be changed by calling onValueChange(value, name).
Afterwards value will be stored in the passed state e.g. state[name] ===
value.

Example for using the State component:
```jsx
<State>
{({
  state,
  onValueChange,
}) => {
  const {myvalue} = state;
  return (
    <TextField
      value={myvalue}
      name="myvalue"
      onChange={onValueChange}
    />
    <button onClick={() => onSave(state)} />
  );
}}
</State>
```
With the state component it will not be necessary to re-implement the
state handling in components using forms.


swaterkamp:
Looks like a good idea! Maybe the example can be improved to be fully working (e.g. add a button).

PR #40, opened 2017-10-31
Move ng to a top level directory

bjoernricks:
ng can be served independently from gsad daemon (as it is done with the
development server already). Therefore move ng to the top level
directory and the sources for gsad into a new gsad directory. This
separates ng and gsad cleanly.

With this change the strange src/html/classic/ng/src path is removed
too.


CMakeLists.txt
```diff
@@ -555,19 +175,11 @@ set (CMAKE_C_FLAGS              "${CMAKE_C_FLAGS} -Wall -Wformat -Wformat-securi
 set (CMAKE_C_FLAGS_DEBUG        "${CMAKE_C_FLAGS_DEBUG} -Werror")
 set (CMAKE_C_FLAGS_RELEASE      "${CMAKE_C_FLAGS_RELEASE} ${HARDENING_FLAGS}")
 
+add_subdirectory (gsad)
```
wiegandm: The 'gsad' subdirectory should be ignored when SKIP_SRC is defined. Otherwise all dependencies of gsad will be required to build documentation or packages.

bjoernricks: Ah good to know! I've just moved the check for SKIP_SRC to the new gsad/CMakeLists.txt. Wasn't aware of this problem. I'll update the PR.

bjoernricks:
Just updated the PR for the latest changes in master.

PR #55, opened 2017-11-04
Actually activate Loading component

swaterkamp:
Set <Loading loading={true}> to actually show the loading component,
when the page is loading. So many loadings...


swaterkamp:
I'm totally in favor of removing the loading prop. Or at least to set the default to loading=true. Maybe in some weird use-case we need to set loading to false manually.

bjoernricks:
Obsolete with #56

PR #56, opened 2017-11-04
Make the Loading component look nice

swaterkamp:
Use some CSS animation to show a rotating circle in green(bone) colors

### DIFF
--- a/ng/src/web/components/loading/loading.js
+++ b/ng/src/web/components/loading/loading.js
@@ -1,6 +1,7 @@
 /* Greenbone Security Assistant
  *
  * Authors:
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  * Björn Ricks <bjoern.ricks@greenbone.net>
  *
  * Copyright:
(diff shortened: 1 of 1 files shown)


src/html/classic/ng/src/web/components/loading/loading.js
```diff
@@ -23,18 +24,40 @@
...
+    width: '120px',
+    height: '120px',
+    fontWeight: 'bold',
+    animation: `${css.keyframes({
+      '0%': {transform: `rotate(0deg)`},
+      '100%': {transform: `rotate(360deg)`},
+    })} 2s linear infinite`,
+});
+
+const Center = glamorous.div({
```
bjoernricks: Hmm maybe it would be nice to use Layout here like
```jsx
const StyledLayout = glamorous(Layout)({
  width: '100%',
  paddingTop: '100px',   
}):

...

<StyledLayout align="center">
  <Loading/>
</StyledLayout>
```

src/html/classic/ng/src/web/components/loading/loading.js
```diff
@@ -23,18 +24,40 @@
...
+  display: '-webkit-flex',
+  display: 'flex',
+  width: '100%',
```
bjoernricks: wouldn't flexGrow: 1 be better here? i am not sure about the difference

src/html/classic/ng/src/web/components/loading/loading.js
```diff
@@ -23,18 +24,40 @@
...
+  justifyContent: 'center',
+  paddingTop: '100px',
```
bjoernricks: Feels a bit wrong to use a "random" padding-top of 100px. Where does it come from?

PR #65, opened 2017-11-08
Update the German translations for GSA 7.0

timopollmeier:
The labels in the "Duration Timeline" chart and a few other elements were lacking German translations,
 which should be fixed with these changes.

### DIFF
--- a/src/html/classic/js/gsa_chart_helpers.js
+++ b/src/html/classic/js/gsa_chart_helpers.js
@@ -1525,10 +1525,9 @@
   gch.field_name = function(field, type) {
     switch (field.toLowerCase()) {
       case 'c_count':
-        return gsa._('total {{resource_type_plural}}',
+        return gsa._('total {{- resource_type_plural}}',
             {
               resource_type_plural: gch.resource_type_name_plural(type),
-              interpolation: {escape: false},
             });
       case 'count':
         return gch.resource_type_name_plural(type);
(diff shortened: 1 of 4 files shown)


src/html/classic/js/gsa_chart_helpers.js
```diff
@@ -1587,25 +1591,37 @@
...
+          = label + gch.field_name(info.column ? info.column : info.stat,
+                                   info.type);
+
     if (include_stat) {
       switch (info.stat) {
         case 'min':
-          label = label + 'min. ';
+          label = label + gsa._('min. {{field_name}}',
+                                {field_name: field_name,
+                                 interpolation: {escape: false}});
```
bjoernricks: After looking at https://www.i18next.com/interpolation.html did they changed `escape` to `escapeValue`? I have always used `_('min. {{- field_name}}')` until now. Did you check if this works?

bjoernricks: Just saw `escape` expects a function https://www.i18next.com/interpolation.html#all-interpolation-options. Therefore your changes shouldn't work atm.

timopollmeier:
I didn't see any problems with the current use cases but I've changed it to use the "-" prefix
 for unescaped variables.

bjoernricks:
Thanks. Next time it would be nice to update your pull request by changing the history with git rebase -i instead of adding additional fix commits.

PR #81, opened 2017-11-10
Do not store filt_id in last_filt_ids if it is -2 (GSA master)

timopollmeier:
The value -2 should be ignored like 0 when storing the filt_id in the
last_filt_ids field of the current user credentials.
Otherwise it would cause problems where pages revert to a built in
filter selecting all items when they should not.


timopollmeier:
I've added macros for the special filt_id values to GSA and also to manager (SVN r29965).

bjoernricks:
Really really nice! Thank you so much!

PR #82, opened 2017-11-13
Ldap authentication

swaterkamp:
The LDAP authentication page is added with a corresponding new API to save auth settings and the possibility to get certificate info from cmd: auth_settings, that where not returned until now.

### DIFF
--- a/ng/src/web/pages/ldap/ldappage.js
+++ b/ng/src/web/pages/ldap/ldappage.js
@@ -1,0 +1,235 @@
+/* Greenbone Security Assistant
+*
+* Authors:
+* Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+*
+* Copyright:
+* Copyright (C) 2017 Greenbone Networks GmbH
+*
+* This program is free software; you can redistribute it and/or
+* modify it under the terms of the GNU General Public License
+* as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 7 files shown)


ng/src/web/pages/ldap/ldappage.js
```diff
@@ -0,0 +1,239 @@
...
+              <FileField
+                name="certificate"
+                onChange={this.handleValueChange}
+              />
+            </Layout>
+          </FormGroup>
+          <Layout align="center">
+            <Button
+              width="auto"
+              onClick={() => this.handleSaveSettings()}
```
bjoernricks: Not required to use a arrow function here anymore. this.handleSaveSettings is already bound.

ng/src/web/pages/ldap/ldappage.js
```diff
@@ -0,0 +1,239 @@
...
+
+  load() {
+    this.getLdapAuth()
+    .then(this.setState({loading: false}));
+  }
+
+  getLdapAuth() {
+    const {gmp} = this.props;
+    const auth_data = gmp.user.currentAuthSettings().then(response => {
+      const data = response.data.getEntries()[1][1]; // LDAP
```
bjoernricks:
```javascript
data.get('method:ldap_connect')
```

PR #85, opened 2017-11-13
Radius authentication

swaterkamp:
Add new RADIUS authentication page with corresponding API updates etc.

### DIFF
--- a/ng/src/web/pages/radius/radiuspage.js
+++ b/ng/src/web/pages/radius/radiuspage.js
@@ -1,0 +1,171 @@
+/* Greenbone Security Assistant
+*
+* Authors:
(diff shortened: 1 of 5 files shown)


ng/src/web/pages/radius/radiuspage.js
```diff
@@ -0,0 +1,171 @@
...
+  constructor() {
+    super();
+    this.state = {
+      enable: '',
+      radiushost: '',
+      radiuskey: '',
+      loading: 'true',
+    }
+
+    this.getRadiusAuth = this.getRadiusAuth.bind(this);
```
bjoernricks: Just a hint: It's not required to bind this.getRadiusAuth because it is not passed as a handler.

PR #88, opened 2017-11-14
About page

swaterkamp:
Add new About page

### DIFF
--- a/ng/src/web/pages/help/about.js
+++ b/ng/src/web/pages/help/about.js
@@ -1,0 +1,125 @@
+/* Greenbone Security Assistant
+*
+* Authors:
+* Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+*
+* Copyright:
+* Copyright (C) 2017 Greenbone Networks GmbH
+*
+* This program is free software; you can redistribute it and/or
+* modify it under the terms of the GNU General Public License
+* as published by the Free Software Foundation; either version 2
+* of the License, or (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


ng/src/web/pages/help/about.js
```diff
@@ -0,0 +1,133 @@
...
+import ExternalLink from '../../components/link/externallink.js';
+import Img from '../../components/img/img.js';
+
+import Layout from '../../components/layout/layout.js';
+import Section from '../../components/section/section.js';
+
+const StyledLayout = glamorous(Layout) ({
+  margin: '0 auto',
+  maxWidth: '1100px',
+  display: 'flex',
```
bjoernricks: A Layout already has `display: flex`. Also it has properties to set wrap and the alignments. Just remove display, flexFlow and justifyContent.

Instead use it like
```jsx
<StyledLayout
  wrap
  align="center"
>
 ....
</StyledLayout>
```
in the render method at line 69.

ng/src/web/pages/help/about.js
```diff
@@ -0,0 +1,133 @@
...
+  '@media screen and (max-width: 800px)': {
+    marginRight: "0px",
+  }
+});
+
+const ImageBlock = glamorous.div({
+  maxWidth: '400px',
+})
+
+class About extends React.Component {
```
bjoernricks: No need to have a class component here. You can use a function component instead because you don't have any handlers and also no state.

ng/src/web/pages/help/about.js
```diff
@@ -0,0 +1,133 @@
...
+
+  constructor(...args) {
```
bjoernricks: Just a hint: the constructor here is unnecessary.

PR #110, opened 2017-11-20
Show busy indicator on save button while loading

swaterkamp:
Note: The button has not been set to "disabled" explicitly. I tested clicking it several times while it was in "loading-mode" and it did not create new queries in the network-tab in the browser's Dev-Tools.

### DIFF
--- a/ng/src/web/components/dialog/savedialog.js
+++ b/ng/src/web/components/dialog/savedialog.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  *
  * Copyright:
  * Copyright (C) 2017 Greenbone Networks GmbH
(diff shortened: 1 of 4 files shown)


ng/src/web/components/dialog/savedialog.js
```diff
@@ -116,6 +126,7 @@ class SaveDialogContent extends React.Component {
             </ScrollableContent>
             <DialogFooter
               title={_('Save')}
+              loading={this.state.loading}
               onClick={() => this.handleSaveClick(state)}
```
bjoernricks: Could you change the onClick handler to
```jsx
onclick={() => !loading && this.handleSaveClick(state)}
```
?

Then the handler will not be called when loading is set to true.

bjoernricks: As an alternative you could add a check for loading to handleSaveClick e.g.
```javascript
if (onSave && !this.state.loading) {
```

swaterkamp: I went for the alternative solution.

PR #113, opened 2017-11-21
Fix array-mapping for reports-array

swaterkamp:
Use own gmp/utils.js map() to map through array of reports. The previous
usage of Array.prototype.map() threw a "map is not a function" error.

### DIFF
--- a/ng/src/gmp/commands/performance.js
+++ b/ng/src/gmp/commands/performance.js
@@ -20,7 +20,10 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-import {is_defined} from '../utils.js';
+import {
+  is_defined,
+  is_array,
+} from '../utils.js';
 
 import HttpCommand from './http.js';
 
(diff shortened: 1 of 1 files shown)


bjoernricks:
Oh this commit is also included in #114

bjoernricks:
So we have to merge this one first.

swaterkamp:
So basically what you would do is returning an array from perfomance.js like
```javascript
return response.setData([reports]);
```
instead of
```javascript
return response.setData(reports);
```
to make sure the Array.prototype.map() can map through single-item responses?

bjoernricks:
```javascript
return response.setData(is_array(reports) ? reports : [reports]);
```
would fit here imho. is_array can be imported from gmp/utils.js.

```javascript
return response.setData(map(reports, report => report));
```
would also be fine. I really would like to ensure that the consumer (our preference page) will always get an array.

Some background: Only getting a single item in this case is a special behavior of our xml2js library. I have already used another xml to js library that always returns arrays independent of having only one element of a kind, which is much nicer and more consistent. Also when we switch to another backend which may be GraphQL or REST based we will also always get an array.

PR #114, opened 2017-11-21
Performance Page and Icon Menu

swaterkamp:


ng/src/web/components/menu/iconmenu.js
```diff
@@ -35,10 +35,11 @@ import withIconSize from '../icon/withIconSize.js';
 
 const Div = glamorous.div({
   position: 'relative',
+  marginTop: '-3px',
```
bjoernricks: Could you clarify why it is -3px? The value isn't obvious to me.

swaterkamp: When the menu opened upon mouse-over on the icon, the gap between the icon and the menu was 3px. Basically, I did a trial-and-error until the gap was closed.

bjoernricks: Where does this gap come from? In the past the gap was sometime there and sometimes not. Depending on the used browser and it's version. Therefore I really would like to know the reason for the gap. At the moment I don't think this is the right fix. It only fixes the symptoms.

swaterkamp: I tested in Firefox 57.0 and Chrome 62.0. In both browsers the gap is shown when the menu opens. However, the clickable area seems to be bigger in chrome (reaching the menu) than in Firefox (even being about 1-2 px smaller than the icon itself). As far as I understood it, Firefox and Chrome calculate the sizes differently of spans and divs around the elements in question here. Why and how exactly is not obvious to me.

bjoernricks: Yes that's exactly my observation too. I have tried to fix this in the current (old) web page version several times. It broke many times due to different browsers and different versions.

swaterkamp: My best guess would be differences between browsers and their box model.

bjoernricks: If you still would like to add the -3px margin could you please add a meaningful comment to the css property. Would be nice to be able to remember the reason for adding the damn margin in the future ;-)

swaterkamp: I'd rather not add the margin, but this is the only fix atm that I can think of. If you are fast enough, you can bridge the gap before FF notices that :hover ended, but otherweise you will not be able to e.g. open the "New Container Task" dialog.
I added a short comment about the fix.

swaterkamp: Just came to my mind: Is it now possible that in older browsers the -3px margin causes an overlap of menu and icon? You said that it was dependent on the browser version whether the margin appeared or not. This would be a design peeve, but functionality in newer browsers seems to be more important.

bjoernricks: We should really not care about old brothers other then ie11. We'll only support latest Chrome, latest Firefox, latest Edge and IE11. IE11 is a burden but to many people are still using it.

bjoernricks: I had several issues with the icon menu and the margin. I tried to fix that at least twice and it seems still broken. In ng I have seen already some additional weird behavior where the icon menu was on the left side of the icon. But this wasn't always reproducible.

Therefore you can decide how we are going to manager this. I'll approve the PR.

bjoernricks:
Just saw you have to update this PR first to get rid of the "Fix array-mapping for reports-array" commit.

ng/src/web/components/menu/iconmenu.js
swaterkamp: I would like to have a look at this again. I only got the 3px gap in Chrome and FF and no other weird behavior. As the gap doesn't seem to be the only problem, the unexpected behavior might have some reasons other than the box model or so. If I don't find a solution within a couple of hours, we should merge this PR to at least fix the gap.

bjoernricks:
Could you please drop the "Fix array-mapping for reports-array" from this PR. It's a left over and already merged in another PR with a different commit id.

PR #116, opened 2017-11-22
Confirmation Dialog for ExternalLink

swaterkamp:
External links now feature a disclaimer/confirmation dialog and users are required to confirm that they are leaving GSA and opening an external website.

### DIFF
--- a/ng/src/web/components/dialog/linkconfirmationdialog.js
+++ b/ng/src/web/components/dialog/linkconfirmationdialog.js
@@ -1,0 +1,154 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
+ * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+ *
+ * Copyright:
+ * Copyright (C) 2017 Greenbone Networks GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import React from 'react';
+import glamorous from 'glamorous';
+
+import _ from 'gmp/locale.js';
+
+import PropTypes from '../../utils/proptypes.js';
+
+import Dialog from '../dialog/dialog.js';
+import DialogContent from '../dialog/content.js';
+import DialogTitle from '../dialog/title.js';
+import ScrollableContent from '../dialog/scrollablecontent.js';
+
+import Button from '../dialog/button.js';
+
+import Layout from '../../components/layout/layout.js';
+
+const StyledLayout = glamorous(Layout)({
+  justifyContent: 'space-between',
+  borderWidth: '1px 0 0 0',
+  borderStyle: 'solid',
+  borderColor: '#ddd',
+  marginTop: '15px',
+  padding: '10px 15px 10px 15px',
+});
+
+class ConfirmationDialogContent extends React.Component {
+
+  constructor(...args) {
+    super(...args);
+
+    this.handleResume = this.handleResume.bind(this);
+  }
+
+  handleResume() {
+    const {
+      onResumeClick,
+    } = this.props;
+
+    if (onResumeClick) {
+      onResumeClick();
+    }
+    this.props.close();
+  }
+
+  render() {
+    const {
+      moveprops,
+      text,
+      title,
+    } = this.props;
+
+    return (
+      <DialogContent>
+        <DialogTitle
+          title={title}
+          onCloseClick={this.props.close}
+          {...moveprops}
+        />
+        <ScrollableContent>
+          {text}
+        </ScrollableContent>
+        <StyledLayout >
+          <Button
+            onClick={this.props.close}
+            title={_('Cancel')}
+          >
+            {_('Cancel')}
+          </Button>
+          <Button
+            onClick={this.handleResume}
+            title={_('Follow Link')}
+          >
+            {_('Follow Link')}
+          </Button>
+        </StyledLayout>
+      </DialogContent>
+    );
+  }
+}
+
+ConfirmationDialogContent.propTypes = {
+  close: PropTypes.func.isRequired,
+  moveprops: PropTypes.object,
+  text: PropTypes.string,
+  title: PropTypes.string.isRequired,
+  onResumeClick: PropTypes.func.isRequired,
+};
+
+const LinkConfirmationDialog = ({
+  width,
+  text,
+  title,
+  visible,
+  onClose,
+  onResumeClick,
+}) => {
+  return (
+    <Dialog
+      visible={visible}
+      width={width}
+      onClose={onClose}
+    >
+      {({
+        close,
+        getMoveProps,
+      }) => (
+        <ConfirmationDialogContent
+          close={close}
+          moveprops={getMoveProps()}
+          text={text}
+          title={title}
+          onResumeClick={onResumeClick}
+        />
+      )}
+    </Dialog>
+  );
+};
+
+LinkConfirmationDialog.propTypes = {
+  text: PropTypes.string,
+  title: PropTypes.string.isRequired,
+  visible: PropTypes.bool.isRequired,
+  width: PropTypes.string,
+  onClose: PropTypes.func.isRequired,
+  onResumeClick: PropTypes.func.isRequired,
+};
+
+export default LinkConfirmationDialog;
+
+// vim: set ts=2 sw=2 tw=80:
--- a/ng/src/web/components/link/externallink.js
+++ b/ng/src/web/components/link/externallink.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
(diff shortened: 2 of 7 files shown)


ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,176 @@
...
+      {({
+        close,
+        getMoveProps,
+      }) => (
+        <ConfirmationDialogContent
+          close={close}
+          moveprops={getMoveProps()}
+          text={text}
+          title={title}
+          onResumeClick={() => onResumeClick(to)}
```
bjoernricks: Something is wrong with the onResumeClick handler. I guess you did mix something. Here it would be better to pass the to prop also to ConfirmationDialogContent and not to create a arrow handler function.

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,176 @@
...
+class ConfirmationDialogContent extends React.Component {
+
+  constructor(...args) {
+    super(...args);
+
+    this.handleResume = this.handleResume.bind(this);
+    this.handleClose = this.handleClose.bind(this);
+  }
+
+  handleResume(url) {
```
bjoernricks: As mentioned in the other comment I guess here is something wrong. Where does the url param come from? At the moment it is always undefined. I guess you would like to get passed the to prop as the url param. In this case forwared the to prop to the ConfirmationDialogContent and get it from this.props directly instead of using a parameter in the handler.

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,176 @@
...
+      width={width}
+      onClose={onClose}
+    >
+      {({
+        close,
+        getMoveProps,
+      }) => (
+        <ConfirmationDialogContent
+          close={close}
+          moveprops={getMoveProps()}
```
bjoernricks: I really don't like this moveprops thing here but I also don't have a better solution at the moment. Doesn't feel right.

bjoernricks: I guess i have to rethink the getMoveProps stuff at the Dialog to get a better solution

swaterkamp:
During updating of the commits there were some merge conflicts. Resolving those ended up deleting your comments on the original commit. Sorry!

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,178 @@
...
+
+  render() {
+    let {
+      moveprops,
+      text,
+      title,
+    } = this.props;
+
+    // display slashes correctly, which have been escaped before
+    text = text.replace(/&#x2F;/g, '/');
```
bjoernricks: Now I understand why you are adding this :-) It's not necessary. I'll point you to the code how to fix it in another comment.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,139 @@
...
+  render() {
+    const {
+      children,
+      className,
+      dialogvisible,
+      to,
+      ...props
+    } = this.state;
+    const dialogtitle = _('You are leaving GSA');
+    const dialogtext = _('This dialog will open a new window for {{to}} ' +
```
bjoernricks: Just use `{{- to}}` here to fix the url escaping issue. For details see https://www.i18next.com/interpolation.html#unescape

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,139 @@
...
+    }
+
+    className = classes(className, 'external-link');
+
+    this.setState({
+      children,
+      className,
+      to,
+      ...props,
+    });
```
bjoernricks: Don't put the props into the state. Thats not necessary.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,139 @@
...
+  }
+
+  handleCloseDialog() {
+    this.setState({
+      dialogvisible: false,
+    });
+  }
+
+  handleOpenLink(url) {
+    window.open(url, '_blank', 'resizable=yes', 'noopener');
```
bjoernricks: Just saw you don't even have to pass the to prop to the LinkConfirmationDialog. Just use it in the handleOpenLink handler directly.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,139 @@
...
+
+    this.state = {
+      dialogvisible: false,
+    };
+
+    this.handleClick = this.handleClick.bind(this);
+    this.handleCloseDialog = this.handleCloseDialog.bind(this);
+  }
+
+  componentDidMount() {
```
bjoernricks: You don't need componentDidMount.

ng/src/web/components/link/externallink.js
bjoernricks: Don't call this method load because it isn't loading anything. Also this method is completely unnecessary because the all variables are derived from the props. Therefore put this code into render and don't set any state.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,123 @@
...
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
- */
+*
```
bjoernricks: Please don't change the indentation of the copyright header.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,123 @@
...
+  }
+
+  render() {
+    const {
+      dialogvisible,
+    } = this.state;
+
+    let {
+      children,
+      className,
```
bjoernricks: We should get rid of adding a className and therefore of calling classes too. It's not necessary anymore. Since using glamorous this has become obsolete.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,123 @@
...
+          href={to}
+          className={className}
+          onClick={this.handleClick}
+        >
+          {children}
+        </a>
+        <LinkConfirmationDialog
+          visible={dialogvisible}
+          onClose={this.handleCloseDialog}
+          onResumeClick={() => this.handleOpenLink(to)}
```
bjoernricks: Not necessary to create an arrow function to pass the to prop to the handler. The handler can use this.props.to instead.

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,123 @@
...
+    this.setState({
+      dialogvisible: false,
+    });
+  }
+
+  handleOpenLink(url) {
+    window.open(url, '_blank', 'resizable=yes', 'noopener');
+  }
+
+  openDialog() {
```
bjoernricks: If openDialog isn't used anywhere beside handleClick you may put the setState call directly into handleClick

ng/src/web/components/link/externallink.js
```diff
@@ -1,57 +1,123 @@
...
+  render() {
+    const {
+      dialogvisible,
+    } = this.state;
+
+    let {
+      children,
+      className,
+      to,
+      href,
```
bjoernricks: We should mark href as deprecated and change all usage of ExternalLink to set to instead of href prop.

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,170 @@
...
+import DialogContent from '../dialog/content.js';
+import DialogTitle from '../dialog/title.js';
+import ScrollableContent from '../dialog/scrollablecontent.js';
+
+import Button from '../dialog/button.js';
+
+import Layout from '../../components/layout/layout.js';
+
+const StyledLayout = glamorous(Layout)({
+  display: 'flex',
```
bjoernricks: Not necessary. Layout always sets display: flex.

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,170 @@
...
+  justifyContent: 'space-between',
```
bjoernricks: you could call `<StyledLayout align="space-between">` instead of setting the css prop directly.

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,170 @@
...
+    this.handleClose = this.handleClose.bind(this);
+  }
+
+  handleResume() {
+    const {
+      onResumeClick,
+    } = this.props;
+
+    if (onResumeClick) {
+      const promise = onResumeClick();
```
bjoernricks: Do you use this promise pattern anywhere? If not just remove this code path.

ng/src/web/components/dialog/linkconfirmationdialog.js
```diff
@@ -0,0 +1,170 @@
...
+          () => this.handleClose()
+        );
+      }
+      else {
+        this.handleClose();
+      }
+    }
+  }
+
+  handleClose() {
```
bjoernricks: handleClose is unnecessary. Just use `<DialogTitle ... onCloseClick={this.props.close}>`

ng/src/web/components/dialog/linkconfirmationdialog.js
bjoernricks: If this promise based pattern isn't used ConfirmationDialogContent can be a stateless/function component or even better directly integrated into LinkConfirmationDialog

ng/src/web/components/link/externallink.js
swaterkamp: Sorry, didn't mean to...

ng/src/web/components/dialog/linkconfirmationdialog.js
swaterkamp: Is this a good idea? When I define it in the CSS properties I have all the styling in one place, easy to find. If I split it up into some CSS definitions at the beginning of the code and some props where StyledLayout is rendered, maintenance seems harder.

PR #121, opened 2017-11-23
Dialog cleanups

swaterkamp:
Margins are added to all dialog elements via Dividers.
Alert dialogs are committed separately, to keep an overview of which files belong to alerts and which to other dialogs.

### DIFF
--- a/ng/src/web/components/form/checkbox.js
+++ b/ng/src/web/components/form/checkbox.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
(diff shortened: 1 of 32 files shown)


bjoernricks:
I have the feeling that the form elements are a bit broken in its rendering. They are much older then dividers or even the layout component. They are nearly a 1:1 copy from the old form elements which again are derived from the twitter bootstrap elements (https://getbootstrap.com/docs/3.3/css/#forms).

It doesn't feel right to add another quirk to fix the behavior. This will someday bite us again. But nevertheless I'll add it and take a deeper look when I am back next week.

bjoernricks:
Btw. would be nice to split the coding style cleanups and real changes into different commits next time. It's faster to review then.

Issue #57, opened 2017-11-06
Fix rendering of ovaldef and cve list tables


bjoernricks:
Currently for being able to render divided rows all rows of these lists are put into an tbody element, because before React 16 it wasn't possible to return several elements from a components render method.

This has the side effect that the details are added as a tr child of the table without a tbody in between. Therefore react issues a warning when opening the details.

By using https://reactjs.org/docs/react-component.html#fragments it should be fixed to only have one tbody for all rows.

bjoernricks:
Totally forgot to mention we can fix this by using 
```jsx
<React.Fragment>...</React.Fragment>
```
only the new syntax <> for Fragments needs a babel plugin not yet provided by create-react-app.

PR #125, opened 2017-11-28
Ovaldef and cve tables

swaterkamp:
Concerning issue #57.

### DIFF
--- a/ng/src/web/components/table/stripedtable.js
+++ b/ng/src/web/components/table/stripedtable.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  *
  * Copyright:
  * Copyright (C) 2016 - 2017 Greenbone Networks GmbH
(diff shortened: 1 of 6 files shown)


ng/src/web/components/table/stripedtable.js
```diff
@@ -43,7 +44,42 @@ const StripedTable = glamorous(Table)({
...
+          '& > tbody:only-of-type > tr:nth-of-type(4n-1)']: {
+            background: '#EEEEEE',
+          },
+          ['& > tbody:not(:only-of-type):hover, ' +
+          '& > tbody:only-of-type > tr:hover,' +
+          '& > tbody:only-of-type > tr:hover + tr']: {
+            background: '#DDDDDD',
+          },
+          // FIXME hovering groups don't fit when mouse over even rows
+          ['& > tbody:only-of-type > tr:nth-of-type(even):hover,' +
```
swaterkamp: The coloring of the grouped rows only works, when the first column is hovered, that means effectively hovering every even row. When hovering an odd row (i.e. the descriptions) the grouping breaks and the next even row is colored together with the currently hovered odd row.
Without the surrounding TableBody I can't find a solution to group an even row together with the PREVIOUS one. Any suggestions are highly appreciated!

ng/src/web/components/table/stripedtable.js
```diff
@@ -43,7 +44,42 @@ const StripedTable = glamorous(Table)({
...
     },
   },
-});
+},
+  ({doubleRow}) => {
+    if (doubleRow) {
+      return {
+        // FIXME when details are toggled, coloring is scrambled up, due to
+        // adding just one more tr and messing with the groups of two
+        '@media screen': {
```
swaterkamp: Despite the issue described in the comment for line 71:
The coloring of groups of rows breaks, once the details for an entity are toggled. As the details are effectively added as a single tr, grouping with even/odd doesn't work anymore. A simple, yet "ugly" solution would be to add an empty table row with each toggled details row.
So we would need a reset of the row counter right after toggled details or another way of selecting the correct rows. Right now, I don't have a solution for this, but I wanted to show the changes already and maybe get some feedback on how to fix this.

swaterkamp: Or we go back to the rows each wrapped in a tbody. In that case the details row should also get a tbody around it, which should solve all problems.

swaterkamp:
As the branch keeps getting more and more out of date, I'm going to close the PR for now until there is a good solution for using fragments and the coloring issue for double rows.

PR #135, opened 2017-12-01
Agent installer download

bjoernricks:
Add an api for downloading an agent installer.

### DIFF
--- a/ng/src/web/pages/agents/component.js
+++ b/ng/src/web/pages/agents/component.js
@@ -41,6 +41,7 @@ class AgentComponent extends React.Component {
   constructor(...args) {
     super(...args);
 
+    this.handleDownloadInstaller = this.handleDownloadInstaller.bind(this);
     this.handleVerifyAgent = this.handleVerifyAgent.bind(this);
     this.openAgentDialog = this.openAgentDialog.bind(this);
   }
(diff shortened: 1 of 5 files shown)


ng/src/web/pages/agents/component.js
```diff
@@ -51,6 +52,15 @@ class AgentComponent extends React.Component {
     gmp.agent.verify(agent).then(onVerified, onVerifyError);
   }
 
+  handleDownloadInstaller(agent) {
+    const {gmp, onInstallerDownloadError, onInstallerDownloaded} = this.props;
+
+    return gmp.agent.downloadInstaller(agent).then(response => {
+      const filename = 'agent-' + agent.id + '-installer';
```
swaterkamp: I'd use agent.name for a meaningful filename. Two agents can't have the same name, so filenames are unique anyway.

ng/src/web/pages/agents/detailspage.js
```diff
@@ -101,6 +102,12 @@ const ToolBarIcons = withCapabilities(({
           onClick={onAgentVerifyClick}
         />
       }
+      <Icon
+        img="agent.svg"
+        value={entity}
+        title={_('Download Agent installer package')}
```
swaterkamp: I'd go for lower case "agent", because for the user it's really just a noun, not the component.

PR #137, opened 2017-12-01
Credential download refactoring

bjoernricks:
Don't use LegacyLinks for downloading credentials anymore.


ng/src/web/pages/credentials/detailspage.js
```diff
@@ -210,13 +217,16 @@ const Page = ({
       onDeleteError={onError}
       onDownloaded={onDownloaded}
       onDownloadError={onError}
+      onInstallerDownloaded={onDownloaded}
+      onInstallerDownloadErro={onError}
```
swaterkamp: Should be onInstallerDownloadError={onError}

PR #146, opened 2017-12-04
Fix position of details-toolbar in Firefox

swaterkamp:
The flexbox around the toolbar needed an explicit command to justify its
contents at the start (Firefox only).

### DIFF
--- a/ng/src/web/entity/page.js
+++ b/ng/src/web/entity/page.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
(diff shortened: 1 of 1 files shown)


ng/src/web/entity/page.js
```diff
@@ -179,15 +185,15 @@ class EntityPage extends React.Component {
     }
 
     return (
-      <Layout flex="column" grow="1">
+      <StyledLayout flex="column" grow="1">
```
bjoernricks: Just set align="start" see https://github.com/greenbone/gsa/blob/master/ng/src/web/components/layout/withLayout.js instead of creating a new component. If you are using Layout don't override flex layout based css by creating a new parent component.

PR #152, opened 2017-12-05
Resizable dialogs

swaterkamp:
The three dialog types are given the option to resize. This includes dialogs using the withDialog() HOC, SaveDialogs, and LinkConfirmationDialogs.

### DIFF
--- a/ng/src/web/components/dialog/dialog.js
+++ b/ng/src/web/components/dialog/dialog.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  *
  * Copyright:
  * Copyright (C) 2016 - 2017 Greenbone Networks GmbH
@@ -29,8 +30,12 @@ import PropTypes from '../../utils/proptypes.js';
 
 import DialogContainer from './container.js';
 import DialogOverlay from './overlay.js';
+import Resizer from './resizer.js';
 
 const DEFAULT_DIALOG_WIDTH = '800px';
+const DEFAULT_DIALOG_HEIGHT = '550px';
+const DEFAULT_DIALOG_MIN_HEIGHT = 250;
+const DEFAULT_DIALOG_MIN_WIDTH = 450;
 
 class Dialog extends React.Component {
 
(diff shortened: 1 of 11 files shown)


ng/src/web/components/dialog/dialog.js
```diff
@@ -22,15 +23,31 @@
...
 import DialogOverlay from './overlay.js';
 
 const DEFAULT_DIALOG_WIDTH = '800px';
+const DEFAULT_DIALOG_HEIGHT = '550px';
+const DEFAULT_DIALOG_MIN_HEIGHT = 250;
+const DEFAULT_DIALOG_MIN_WIDTH = 450;
+
+const Resizer = glamorous.div({
+  position: 'absolute',
+  bottom: '.4em',
```
bjoernricks: The resizer is used at dialog.js and withDialog.js so it should be put into an own js module (dialog/resizer.js) and imported in both.

ng/src/web/components/dialog/dialog.js
```diff
@@ -89,53 +110,105 @@ class Dialog extends React.Component {
...
+    if (!resizable) {
+      height = 'auto';
+    }
+
     if (!visible) {
       return null;
     }
 
+    if (is_defined(setWidth)) {
+      width = setWidth;
```
bjoernricks: IMHO this doesn't work correctly. If the width is set on a dialog resizing the width won't be possible anymore.

To get the expected behavior just set the width in componentDidMount:
```javascript
componentDidMount() {
    document.addEventListener('keydown', this.onKeyDown);

    this.setState({width: this.props.width});
}
```

Afterwards only use the width from the state. Then a user is able to set an initial width when the dialog is rendered the first time. Afterwards changes on the width prop aren't considered anymore. If we need this also in future componentWillReceiveProps has to be implemented. But I don't think so at the moment.

ng/src/web/components/dialog/dialog.js
```diff
@@ -145,20 +218,26 @@ class Dialog extends React.Component {
...
           width={width}
+          height={height}
           posX={posX}
           posY={posY}
           innerRef={ref => this.dialog = ref}>
           {children({
             close: this.handleClose,
             getMoveProps: this.getMoveProps,
           })}
+          {resizable && <Resizer
```
bjoernricks: Please put the <Resizer on the next line
```jsx
{resizable &&
  <Resizer onMouseDown={this.onMouseDownResize} />
}
```

IMHO this is easier to read.

ng/src/web/components/dialog/dialog.js
```diff
@@ -145,20 +218,26 @@ class Dialog extends React.Component {
...
+            onMouseDown={this.onMouseDownResize}
+          />}
         </DialogContainer>
       </DialogOverlay>
     );
   }
 };
 
 Dialog.propTypes = {
+  height: PropTypes.string,
```
bjoernricks: Should this really be only a string or can it be numberOrNumberString from proptypes.js?

PR #181, opened 2017-12-14
Add sorting indicator to table columns

swaterkamp:
A small icon is shown to indicate which column is sorted and in which direction.

### DIFF
--- a/ng/src/web/components/table/head.js
+++ b/ng/src/web/components/table/head.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  *
  * Copyright:
  * Copyright (C) 2016 - 2017 Greenbone Networks GmbH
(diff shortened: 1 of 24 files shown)


ng/src/web/components/table/head.js
```diff
@@ -38,20 +40,48 @@ const TableHead = ({
     className,
     colSpan,
     rowSpan,
-    sortby,
+    sortDir,
+    filterSortBy,
```
bjoernricks: Could you please rename it to sortOrder. For me it is not consistent to name it filterSortBy and also have sortDir which comes from the filter too.

ng/src/web/components/table/head.js
```diff
@@ -38,20 +40,48 @@ const TableHead = ({
...
+    sortBy,
     width,
     onSortChange,
     ...other
   }) => {
+
+  const sortSymbol = sortDirection => {
```
bjoernricks: Any reason why you did put this into a function? It can just be a variable because the param sortDirection is static and provided via props already.

ng/src/web/entities/table.js
```diff
@@ -106,6 +106,13 @@ class EntitiesTable extends React.Component {
 
     const filterstring = is_defined(filter) ? filter.toFilterString() : '';
 
+    let sortDir;
+    let filterSortBy;
+    if (is_defined(filter)) {
```
bjoernricks: Either put also line 107 into the if clause or use the ternary operator also for sortDir and filterSortBy/sortDirection

PR #209, opened 2017-12-20
Remove its face from gsad

bjoernricks:
The additonal face/webpage its isn't used anymore especially not in ng.
Therefore it can be removed from gsad completely.


bjoernricks:
I have added several people to the reviewers list to get them notified about the removal of its.

PR #220, opened 2017-12-21
Move static content from gsad to ng

bjoernricks:
The static content is now only used in ng.


ng/public/img/.directory
```diff
@@ -0,0 +1,4 @@
+[Dolphin]
```
wiegandm: Looks like a stray file added in error, please remove.

PR #221, opened 2017-12-21
Make the XML transformation more flexible

bjoernricks:
Currently we are using https://github.com/abdmob/x2js for xml to js transformations. This library has several disadvantages:

* it's not importable via import statement
* it returns arrays or objects for xml elements (it can be possible that an array is expected because the xml element may have several children)
* it hasn't been updated for a while
* it's license is Apache v2 which isn't compatible with GPLv2

In future I would like to be able to switch to https://github.com/Leonidas-from-XIV/node-xml2js 
This change should be doable step by step. Therefore the xml transformation must be more flexible and adjustable per command class.


bjoernricks:
Would be nice to merge this commit via "Merge pull request" at the end to get a reference to this pr in git log.

PR #229, opened 2017-12-28
Link to the manual instead of the built-in help system

swaterkamp:


bjoernricks:
Sorry this PR needs some additional work. I don't think adding special props for specific usecases to an existing component is the right approach. For this usecase a new separated component should  be created. But I need to take a deeper look at the sources before making another statement.

PR #253, opened 2018-01-15
Copying update

bjoernricks:
Update COPYING file after ng got splitted from gsad and several files got removed from the repo.


COPYING
```diff
@@ -9,82 +9,43 @@ or any later version (GNU GPLv2+). Some files have a different, but
 compatible licenses (BSD 3-clause and MIT).
```
wiegandm: Reference to BSD-3 should be removed here as well.

bjoernricks:
As discussed with Michael just added missing license headers and updated the COPYING file.

PR #256, opened 2018-01-15
Resolve one call of xsl_transform and cleanup.

janowagner:


gsad/src/gsad_http.c
```diff
@@ -465,38 +465,78 @@ char *
...
+  gchar *html;
+
+  html = g_strdup_printf (
+    "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\" \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">"
+    "<html xmlns=\"http://www.w3.org/1999/xhtml\" lang=\"\">"
+    "<head>"
+      "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\"></meta>"
+      "<link rel=\"icon\" href=\"/img/favicon.gif\" type=\"image/gif\"></link>"
+      "<title>Greenbone Security Assistant</title>"
+      "<link rel=\"stylesheet\" type=\"text/css\" href=\"/css/gsa-base.css\"></link>"
```
bjoernricks: You can remove most of the css links and script tags. Especially the js files below aren't included in gsa(d) anymore.

gsad/src/gsad_http.c
```diff
@@ -465,38 +465,78 @@ char *
...
-                         "<token>%s</token>"
-                         "</gsad_response>",
-                         title,
-                         GSAD_VERSION,
-                         msg,
-                         "");
+  gchar *html;
+
+  html = g_strdup_printf (
+    "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\" \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">"
```
bjoernricks: Since we don't use xslt anymore we can use html5 here. Just use <!DOCTYPE html> and remove the xmlns from the <html> tag.

gsad/src/gsad_http.c
```diff
@@ -465,38 +465,78 @@ char *
...
+      "<script src=\"/js/i18next-xhr-0.5.3.js\" type=\"text/javascript\"></script>"
+      "<script src=\"/js/i18next-languagedetector-0.2.2.js\" type=\"text/javascript\"></script>"
+      "<script src=\"/js/moment.js\" type=\"text/javascript\"></script>"
+      "<script src=\"/js/moment-timezone-with-data.js\" type=\"text/javascript\"></script>"
+      "<script src=\"/js/greenbone-ui.js\" type=\"text/javascript\"></script>"
+      "<script src=\"/js/greenbone.js\" type=\"text/javascript\"></script>"
+      "<script src=\"/js/gsa_polyfill.js\" type=\"text/javascript\"></script>"
+    "</head>"
+    "<body>"
+    "<div class=\"gsa-main\"><div>"
```
bjoernricks: I really don't know if the css classes are still available. Maybe we just reduce this page body to something like
```html
<h1>Not found</h1>
<p>The requested content couldn't be found.</p>
```
without any css formatting.

gsad/src/gsad_http.c
janowagner: I intentionally copied over the original transformation output for documentation purpose.
Once the other calls of xsl_transform() are resolved, the HTML parts can be simplified
in conjunction with the actual removal of the then-unused css and js files.

PR #267, opened 2018-01-18
Set table column widths

swaterkamp:


ng/src/web/pages/scanners/row.js
```diff
@@ -60,7 +60,7 @@ const Actions = ({
   onScannerVerifyClick,
 }) => (
   <IconDivider
-    align={['start', 'center']}
+    align={['center', 'center']}
```
bjoernricks: This was by intention as already shown.

PR #269, opened 2018-01-18
Resolve all calls of xsl_transform().

janowagner:


gsad/src/gsad_http.c
```diff
@@ -1332,36 +1332,25 @@ login_xml (const gchar *message, const gchar *token, const gchar *time,
...
 
-  xml = login_xml (message, NULL, ctime_now, NULL, NULL, NULL);
-
-  if (xml_flag)
-    return xml;
-
-  res = xsl_transform (xml, response_data);
-  g_free (xml);
-  return res;
+  return (login_xml (message, NULL, ctime_now, NULL, NULL, NULL));
```
bjoernricks: This could even been improved further. There is no need to return the login xml anymore. Just returning a 200 OK without content should suffice.

gsad/src/gsad_http.c
```diff
@@ -1332,36 +1332,25 @@ login_xml (const gchar *message, const gchar *token, const gchar *time,
 }
 
 /**
- * @brief Removes user token and returns logout page or logout xml in case of
- * xml_flag is set to true
+ * @brief Removes user token and returns logout xml.
```
bjoernricks: Actually it's the login xml as also documented with `@return `

PR #275, opened 2018-01-19
Copying update

bjoernricks:
Try to get github to display our license.

It's a bit difficult to get this working with our files. Github checks via https://github.com/benbalter/licensee which license a repo has and displays the result. Adding several files with license or copying can prevent that if a different or no license can be found in these files. See https://help.github.com/articles/licensing-a-repository/#detecting-a-license

I am still not sure if gplv2 will be shown after this PR has been merged because licensee still checks COPYING


bjoernricks:
I am not merging this PR currently because we have to get rid of the MIT based files first.

PR #277, opened 2018-01-20
Removed handling of xml parameter and removed xsl transformation functions.

janowagner:


bjoernricks:
just a hint: next_page_url() is also unused with ng.

PR #283, opened 2018-01-22
Replace Select2

swaterkamp:
The old Select2 component is replaced by Select and MultiSelect respectively. The two new components are native React components and they replace the old jQuery version.


bjoernricks:
Next step would be to replace the options by settings the items prob. Most of the time this can be done by replacing 
```jsx
const xyz_opts = render_options(...)
<select>{xyz_opts}</select>
```
with
```jsx
const xyz_items = render_select_items(...)
<select items={xyz_items} />
```

ng/src/web/components/form/select.js
```diff
@@ -91,7 +91,7 @@ class Select extends React.Component {
       className,
       disabled = false,
       items,
-      menuPosition,
+      menuPosition = 'adjust',
```
bjoernricks: what's the reason for the change? i don't think that makes sense. it breaks e.g. the powerfilter

ng/src/web/components/form/multiselect.js
```diff
@@ -162,7 +164,8 @@ class MultiSelect extends React.Component {
       className,
       disabled = false,
       items,
-      menuPosition,
+      menuPosition = 'adjust',
```
bjoernricks: Please adjust your commit message of f1f6e9ae5d982a50475d1bda2cd8fdc3a247b546 or split the commit into to separated commits. Your are not mentioning the setting a new default for menuPosition.

ng/src/web/components/form/select.js
swaterkamp: For the MultiSelect it made sense to have a menu position that looks decent by default. I thought it to be a good idea to have that for Select as well. I didn't recognize this breaks the powerfilter.

PR #312, opened 2018-01-30
Fix loginpage for Internet Explorer

swaterkamp:
The aspect ratio of the Icon was broken in IE, so size calculation had
to be changed mainly relying in width instead of height.
Additionally, IE11 moved the "insecure"-message flush to the username
input field when the browser height was decreased. A spacer is only
added in IE to prevent this.

### DIFF
--- a/ng/src/web/pages/loginpage.js
+++ b/ng/src/web/pages/loginpage.js
@@ -112,7 +112,7 @@ class LoginForm extends React.Component {
     const {username, password} = this.state;
     const protocol_insecure = window.location.protocol !== 'https:';
     return (
-      <Layout flex="column">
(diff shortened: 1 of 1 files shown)


ng/src/web/pages/loginpage.js
```diff
@@ -176,8 +184,9 @@ LoginForm.propTypes = {
 };
 
 const GreenboneIcon = glamorous(GBIcon)({
+  minWidth: '60px',
+  width: '100%',
   minHeight: '60px',
-  maxHeight: '315px',
```
bjoernricks: Couldn't we set the height here instead of using a prop in line 282?

PR #318, opened 2018-01-31
Add test for HTTP header Host in redirects

timopollmeier:
The "Host" HTTP header used in redirects now has to refer to a known
host name or address, which by default includes 127.0.0.1, ::1,
localhost and either all addresses given by the --listen option or
all local interface addresses.
Additional host names or addresses can be given with the new command
line option --allow-header-host.

### DIFF
--- a/src/gsad.c
+++ b/src/gsad.c
@@ -3920,6 +3920,9 @@ validate_host_header (const char *host_header)
   else if (g_utf8_validate (host_header, -1, NULL) == FALSE)
     return 1;
 
+  /*
+   * Find brackets and colons for detecting IPv6 addresses and port.
+   */
   bracket_index = -1;
   colon_index = -1;
   for (char_index = strlen (host_header) - 1;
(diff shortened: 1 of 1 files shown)


src/gsad.c
```diff
@@ -3972,19 +4091,25 @@ send_redirect_to_urn (struct MHD_Connection *connection, const char *urn,
...
+        // Invalid UTF-8
+        send_response (connection,
+                       UTF8_ERROR_PAGE ("'Host' header"),
+                       MHD_HTTP_BAD_REQUEST, NULL,
+                       GSAD_CONTENT_TYPE_TEXT_HTML, NULL, 0);
+        return MHD_YES;
+      case 2:
+      default:
+        // Otherwise invalid
+        send_response (connection, BAD_REQUEST_PAGE, MHD_HTTP_NOT_ACCEPTABLE,
```
cfi-gb: For improved UX i wouldn't throw a bad request around if the user just accessed the system via a wrong Hostname/IP. If the accepted host headers are just misconfigured the user will never now directly why he gets a bad request when accessing the system.

A way better approach would be to throw a 403 / Forbidden and include a short note in the returned HTML document explaining how to configure the system in a way to be able to access it.

src/gsad.c
```diff
@@ -3828,6 +3833,120 @@ gsad_add_content_type_header (struct MHD_Response *response,
...
+  int ret;
+  int char_index, colon_index, bracket_index;
+  gchar *host;
+
+  if (host_header == NULL || strlen (host_header) == 0)
+    return 2;
+  else if (g_utf8_validate (host_header, -1, NULL) == FALSE)
+    return 1;
+
+  bracket_index = -1;
```
bjoernricks: I don't really understand this code in the for loop. Maybe it would be better to extract it into a function or at least at a comment.

src/gsad.c
```diff
@@ -3828,6 +3833,120 @@ gsad_add_content_type_header (struct MHD_Response *response,
...
+  colon_index = -1;
+  for (char_index = strlen (host_header) - 1;
```
bjoernricks: I don't really understand this code. Maybe you can extract it into another function or at least add some comment.

src/gsad.c
```diff
@@ -3828,6 +3833,120 @@ gsad_add_content_type_header (struct MHD_Response *response,
     }
 }
 
+/**
+ * @brief Add all local IP addresses to a GHashTable.
```
bjoernricks: This doesn't include localhost, 127.x.x.x and the ipv6 local addresses right? So the description is a bit confusing.
Maybe move adding the localhost addresses to the hashmap also into this function.

src/gsad.c
```diff
@@ -6402,6 +6538,14 @@ main (int argc, char **argv)
       exit (EXIT_FAILURE);
     }
 
+  /* Initialize addresses and accepted host headers for HTTP header */
```
bjoernricks: Adding the localhost addresses should be moved into a function.

src/gsad.c
bjoernricks: Thanks!!!

bjoernricks:
If @cfi-gb is fine with the additional changes I would merge it instantly.

cfi-gb:
@bjoernricks I'm still unsure about the ``MHD_HTTP_BAD_REQUEST`` vs. ``MHD_HTTP_FORBIDDEN`` in the case where the GSA is misconfigured / doesn't have the expected allowed header configured.

The text / html is IMHO fine for now. It might be linked to the online documentation in the future but i think that is out of the scope of this PR.

bjoernricks:
@cfi-gb choosing the http status code is difficult indeed. I would got for bad request because it is technically correct.

cfi-gb:
It seems CVE-2018-25016 has been assigned to this recently.

PR #345, opened 2018-02-05
Build compile_commands.json and ignore coverage reports

bjoernricks:


bjoernricks:
I have thought about setting the minimum required version but setting a variable to true if it doesn't exist in older versions is fine because the older versions will ignore it. so it doesn't harm for older cmake versions. and i am also not sure if it isn't included before 3.5.

PR #350, opened 2018-02-05
Fix adding tags

swaterkamp:
When adding a tag for an entity on its details page, port lists and scan
configs yielded an error message "invalid resource type"). The
TagsHandler set the resourceType for its render-method to the name of
the entity by default. Removing this default fixes the bug.


swaterkamp:
As it turns out, adding tags and their respective types is a bigger issue. For example: adding tags for NVTs yield errors in all cases. They just change from "must be valid resource type" (without this bugfix) to "internal error" (with this bugfix).

bjoernricks:
The resource type of nvts should be info

PR #356, opened 2018-02-06
Update testing an alert

bjoernricks:
Change gsad to return only 4xx/5xx responses in case of unexpected error when calling test_alert gmp command. 

Display error details in dialog if available.

Cleanup gsad code.


swaterkamp:
The changes of .js-files look reasonable to me. I can't say anything about the .c and .h, though.

PR #359, opened 2018-02-06
Return gsad_message if password is invalid

timopollmeier:
The "Edit My Settings" does not need the settings included in the
response when submitting the changes, so it will only return the error
message if the old password is invalid.
The HTTP return code has also been changed to 400 as it makes more sense
for an invalid user input than 500.


timopollmeier:
I have used the the HTTP code 400 for now because it is the current standard for invalid user input in GSA 7.0 and I'm not sure if it's a good idea to change that as part of a bugfix.

bjoernricks:
400 should be fine for gsa 7.0

Issue #17, opened 2017-10-30
Replace all css files with glamorous


bjoernricks:
We are using the CSS-in-JS solution glamorous which is much nicer to encapsulate styles in components. But there are still some components that use css files directly from the before glamorous era.

src/web/components/dashboard/css/dashboard.css
src/web/components/folding/css/folding.css
src/web/components/menu/css/menu.css
src/web/components/form/css/select2.css
src/web/components/form/css/form.css
src/web/components/form/css/button.css
src/web/components/form/css/datepicker.css
src/web/components/form/css/checkboxradio.css
src/web/components/form/css/spinner.css
src/web/components/form/css/formgroup.css
src/web/pages/users/css/row.css
src/web/wizard/css/wizard.css
src/web/css/app.css
src/web/css/gsa-base.css
src/web/css/gsa-style.css
src/web/css/gsa-login.css
src/web/entities/css/component.css
src/web/entities/css/footer.css

These files should be replaced by using glamorous instead as possible.

swaterkamp:
We reached a point were most of the CSS is replaced by glamorous now.

Here is an update about which .css are left:

src/web/components/dashboard/css/dashboard.css

src/web/components/form/css/form.css
src/web/components/form/css/formgroup.css

src/web/css/app.css
src/web/css/gsa-base.css


Within the react-datepicker package was a new css-file introduced for internal use.

The dashboard.css will be obsolete as soon as the new dashboard component works, I assume.
So what's really left is the form.css and formgroup.css, where we should find a new concept of the design in general, and the base-css for the SPA.

PR #364, opened 2018-02-07
Use glamorous for folding

swaterkamp:
Contributes to #17


ng/src/web/components/folding/folding.js
```diff
@@ -107,12 +116,16 @@ export const withFolding = (Component, defaults = {}) => {
...
-        style={new_style}
-        {...props}
-        onTransitionEnd={onFoldStepEnd}
-        onAnimationEnd={onFoldStepEnd}
-      />
+      <Div
+        {...new_style}
+      >
+        <Component
+          style={new_style}
```
bjoernricks: This prop should be removed from Component. It's being applied to the Div already

ng/src/web/components/folding/folding.js
```diff
@@ -107,12 +116,16 @@ export const withFolding = (Component, defaults = {}) => {
     }
 
     return (
-      <Component
-        style={new_style}
```
bjoernricks: The variable should be renamed. new_style isn't a good name anymore.

ng/src/web/components/folding/folding.js
```diff
@@ -107,12 +116,16 @@ export const withFolding = (Component, defaults = {}) => {
...
-        onAnimationEnd={onFoldStepEnd}
-      />
+      <Div
+        {...new_style}
+      >
+        <Component
+          style={new_style}
+          {...props}
+          onTransitionEnd={onFoldStepEnd}
+          onAnimationEnd={onFoldStepEnd}
```
bjoernricks: Maybe both handlers can be added to the Div instead. But I am not sure about that. Maybe Component isn't generic and is always some defined component which requires both handlers.

swaterkamp: I tried moving them into the Div and the last remaining instance, where folding is used, still functions fine.

Issue #28, opened 2017-10-30
Rewrite dialogs to use SaveDialog


bjoernricks:
The current dialogs are implemented using the withDialog HOC. The API to use these dialogs is based on react refs and the show() methods. refs are a bit problematic because they might already be removed when a promise of the dialog is fulfilled and this can cause serious bugs.

Therefore all dialogs using withDialog should be replaced with a version using the new SaveDialog component.

PR #375, opened 2018-02-08
Use SaveDialog for Agents

swaterkamp:
The withDialog HOC is replaced by using the SaveDialog component.

Contributes to #28

### DIFF
--- a/ng/src/web/pages/agents/dialog.js
+++ b/ng/src/web/pages/agents/dialog.js
@@ -2,9 +2,10 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  *
  * Copyright:
- * Copyright (C) 2017 Greenbone Networks GmbH
+ * Copyright (C) 2017 - 2018 Greenbone Networks GmbH
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


ng/src/web/pages/agents/dialog.js
```diff
@@ -28,81 +29,95 @@ import {is_defined} from 'gmp/utils';
...
+      }}
+    </SaveDialog>
   );
 };
 
 AgentDialog.propTypes = {
   agent: PropTypes.model,
-  name: PropTypes.string,
-  comment: PropTypes.string,
+  title: PropTypes.string,
```
bjoernricks: You don't set a default title anymore. Please either require the title or use
```javascript
const AgentDialog = ({
    agent,
    comment,
    name,
    title = _('New Alert'), // or something else
}) => {
```

ng/src/web/pages/agents/dialog.js
```diff
@@ -28,81 +29,95 @@ import {is_defined} from 'gmp/utils';
...
+  visible: PropTypes.bool,
+  onClose: PropTypes.func.isRequired,
+  onSave: PropTypes.func.isRequired,
   onValueChange: PropTypes.func,
```
bjoernricks: Drop onValueChange proptype. It isn't used anymore.

ng/src/web/pages/agents/dialog.js
```diff
@@ -28,81 +29,95 @@ import {is_defined} from 'gmp/utils';
...
 };
 
 
-export default withDialog({
-  title: _('New Alert'),
-  footer: _('Save'),
-  defaultState: {
-    comment: '',
-    name: _('Unnamed'),
```
bjoernricks: This is missing from the new implementation. You could set a default by using
```jsx
 <SaveDialog
     visible={visible}
     title={title}
     onClose={onClose}
     onSave={onSave}
     initialData={{name: _('Unnamed'), ...agent}}
>
```
Or even better set some variable like
```jsx
const defaults = {name: _('Unnamed')};
return (
 <SaveDialog
     visible={visible}
     title={title}
     onClose={onClose}
     onSave={onSave}
     initialData={{...defaults, ...agent}}
>
...
```
The second version is easier to extend later.

PR #383, opened 2018-02-09
SaveDialogs for Filters and Groups

swaterkamp:
Replace withDialog() HOC with SaveDialog.


ng/src/web/pages/filters/dialog.js
```diff
@@ -23,89 +24,114 @@
...
+    title = _('New Filter'),
     type,
     types,
-    onValueChange,
+    visible,
+    onClose,
+    onSave,
   }) => {
+  const defaults = {name: _('Unnamed'), term: ''};
+
```
bjoernricks: Maybe put the defaults as constant into the js module. This would avoid creating it on every render.

```jsx
const DEFAULTS = {name: _('Unnamed'), term: ''};

const FilterDialog = ({
...
```

Didn't thought about that last time.

ng/src/web/pages/filters/dialog.js
```diff
@@ -23,89 +24,114 @@
...
-          onChange={onValueChange}/>
-      </FormGroup>
+            <FormGroup title={_('Term')}>
+              <TextField
+                name="term"
+                grow="1"
+                value={state.term}
+                size="30"
+                onChange={onValueChange}/>
+            </FormGroup>
```
bjoernricks: Same as above. Maybe here are tabs included.

ng/src/web/pages/filters/dialog.js
```diff
@@ -23,89 +24,114 @@
...
   comment: PropTypes.string,
-  name: PropTypes.string,
+  filter: PropTypes.model,
   term: PropTypes.string,
+  title: PropTypes.string,
   type: PropTypes.string,
-  types: PropTypes.array.isRequired,
-  onValueChange: PropTypes.func,
+  types: PropTypes.array,
+  visible: PropTypes.bool,
```
bjoernricks: visible should be required because SaveDialog requires it.

ng/src/web/pages/groups/component.js
```diff
@@ -39,36 +40,42 @@ class GroupComponent extends React.Component {
...
+    this.closeGroupDialog = this.closeGroupDialog.bind(this);
     this.openGroupDialog = this.openGroupDialog.bind(this);
   }
 
   openGroupDialog(group) {
     const {gmp} = this.props;
 
+    let allUsers = [];
+    gmp.users.getAll().then(response => {
+      allUsers = response.data;
```
bjoernricks: You have to put allUsers into the state here otherwise it is currently always an empty array.

ng/src/web/pages/groups/component.js
```diff
@@ -39,36 +40,42 @@ class GroupComponent extends React.Component {
...
-        name,
-        comment,
-        users,
-      }, {
-        title: _('Edit Group {{name}}', group),
+
+      const title = _('Edit Group {{name}}', group);
+
+      this.setState({
+        allUsers,
```
bjoernricks: allUsers is [] here always.

ng/src/web/pages/groups/component.js
```diff
@@ -39,36 +40,42 @@ class GroupComponent extends React.Component {
...
+        dialogVisible: true,
+        group,
+        title,
       });
     }
     else {
-      this.dialog.show({});
+      this.setState({
+        allUsers,
```
bjoernricks: as above

ng/src/web/pages/groups/dialog.js
```diff
@@ -30,103 +31,117 @@ import Layout from '../../components/layout/layout.js';
...
-    onValueChange,
+    group,
+    title = _('New Group'),
+    visible,
+    onClose,
+    onSave,
   }) => {
+  const is_edit = is_defined(group);
+
+  const defaults = {name: _('Unnamed'), users: []};
```
bjoernricks: Could be moved out of the Dialog component into a constant at the module scope. Same as in the other dialog.

ng/src/web/pages/groups/dialog.js
```diff
@@ -30,103 +31,117 @@ import Layout from '../../components/layout/layout.js';
...
+                onChange={onValueChange}
+              >
+                {
+                  map(allUsers, user => {
+                    return (
+                      <option
+                        key={user.id}
+                        value={user.name}>
+                        {user.name}
+                      </option>
```
bjoernricks: Please create an items array instead of option elements. option elements will be converted into items internally. They are only provided for backwards compatibility.

ng/src/web/pages/groups/dialog.js
```diff
@@ -30,103 +31,117 @@ import Layout from '../../components/layout/layout.js';
...
-  comment: PropTypes.string,
+  allUsers: PropTypes.array,
   grant_full: PropTypes.oneOf(['0', '1']),
-  id: PropTypes.id,
-  name: PropTypes.string,
-  users: PropTypes.array,
-  onValueChange: PropTypes.func.isRequired,
+  group: PropTypes.model,
+  title: PropTypes.string,
+  visible: PropTypes.bool,
```
bjoernricks: visible is required here too.

ng/src/web/pages/filters/dialog.js
swaterkamp: I deleted all indentations completely and inserted new ones. Atom/Git doesn't recognize any changes. I don't know, where the indentations on GitHub are from.

ng/src/web/pages/filters/dialog.js
```diff
@@ -23,89 +24,114 @@
...
-        </Select>
-      </FormGroup>
-    </Layout>
+            <FormGroup title={_('Type')}>
+              <Select
+                name="type"
+                onChange={onValueChange}
+                value={state.type}>
+                {
+                  types.map(option => (
```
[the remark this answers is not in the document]
swaterkamp: I made dure that types is at least an empty array in filter/component.js
lines 94-96:
````
    if (!is_defined(types)) {
      types = [];
    };
````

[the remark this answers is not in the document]
Comment (bjoernricks) [reply ng/src/web/pages/filters/dialog.js]: If somebody uses the dialog without the component he won't know. Therefore it's best to either require types or ensure that it can't crash.

File: ng/src/web/pages/filters/dialog.js [hunk 11]
```diff
@@ -23,89 +24,110 @@
...
+        ...filter,
+        term,
+        type,
+      }}
+    >
+      {({
+        data: state,
+        onValueChange,
+      }) => {
+        const filterOptions = types.map(option => ({
```
bjoernricks: Would be better to move the filterOptions creation to the top before the first jsx statement. In that case it will be only re-created when the dialog is re-rendered. Currently it is re-created on each input value change which is very often.

ng/src/web/pages/groups/component.js
```diff
@@ -39,36 +40,41 @@ class GroupComponent extends React.Component {
...
-        users,
-      }, {
-        title: _('Edit Group {{name}}', group),
+
+      const title = _('Edit Group {{name}}', group);
+
+      this.setState({
+        dialogVisible: true,
+        group,
+        title,
```
bjoernricks: I (re-)thought about allUsers and we have to ensure the state doesn't contain any old and potentially protected data. Therefore we should ensure that allUsers get overriden in the state when opening the dialog. Therefore we should add it to the state here and below too. Then it will either contain the correct data for the current user or an empty array which will get filled later.

ng/src/web/pages/groups/dialog.js
```diff
@@ -30,103 +31,109 @@ import Layout from '../../components/layout/layout.js';
...
+      initialData={{
+        ...DEFAULTS,
+        ...group,
+      }}
+    >
+      {({
+        data: state,
+        onValueChange,
+      }) => {
+        const userOptions = map(allUsers, user => ({
```
bjoernricks: Should be moved up before the first jsx statement to avoid unnecessary re-creation too.

bjoernricks:
Thanks a lot for being patient and for considering all my remarks!

PR #386, opened 2018-02-09
Forbid caching except for static files.

timopollmeier:
For most requests the headers "Expires", "Cache-Control" and "Pragma"
will be added to the HTTP response to forbid caching.
The exception to this are index.html and other static files.


gsad/src/gsad_http.c
```diff
@@ -1035,6 +1039,25 @@ add_cors_headers (http_response_t *response)
...
+/**
+ * @brief Add header to manage caching to a HTTP response.
+ *
+ * @param[in]  response       The HTTP response to add the headers to.
+ * @param[in]  allow_caching  1 to allow caching, 0 to forbid.
+ */
+void
+add_caching_headers (http_response_t *response, int allow_caching)
+{
+  if (allow_caching == FALSE)
```
bjoernricks: Would be better to avoid this check here. Either call this function and add the header or don't call the function at all.

gsad/src/gsad_http.h
```diff
@@ -185,7 +185,7 @@ int send_response (http_connection_t *connection, const char *content,
                    int status_code, const gchar *sid,
                    content_type_t content_type,
                    const char *content_disposition,
-                   size_t content_length);
+                   size_t content_length, int allow_caching);
```
bjoernricks: Please don't add a param for caching to all functions. It make the functions signature only longer and calls more and more unreadable.

gsad/src/gsad_http_handler.c
```diff
@@ -308,7 +308,7 @@ handle_validate (http_connection_t *connection, const char * method,
   if (&url[0] == NULL)
     {
       send_response (connection, BAD_REQUEST_PAGE, MHD_HTTP_NOT_ACCEPTABLE,
-                     NULL, GSAD_CONTENT_TYPE_TEXT_HTML, NULL, 0);
+                     NULL, GSAD_CONTENT_TYPE_TEXT_HTML, NULL, 0, 0);
```
bjoernricks: e.g. this is nearly unreadable without knowing the function signature of send_response. It won't get better with another int param.

timopollmeier:
Something went wrong merging the latest state of the master branch here so I created a new PR: #549

PR #392, opened 2018-02-14
Use react-popper for Select and MultiSelect

swaterkamp:
Opened drop-downs from Select and MultiSelect now _pop_ out of context
and can thereby break the "overflow:hidden" from the dialog contents
(just as DatePicker does, which uses react-popper as well).
While making these changes, another issue came up, which I couldn't
solve: Spinners and some icons are not overlayed by the drop-down but
show through.
Using react-popper seems to be a better basis for fixing the issues with
(Multi)Select, though.


bjoernricks:
I have found some issues with (react-)popper. Also I have a split feeling about it and can't say that I understand its docu. It feels to complicated for this simple use case.

swaterkamp:
Issue was solved with #434

PR #415, opened 2018-02-24
Add simple test for rendering the whole app

bjoernricks:


bjoernricks:
Not ready yet. Test is still complaining about jquery. It should be added when the dashboards are converted to react.

PR #425, opened 2018-03-01
New Schedule chart component

bjoernricks:
Add a Schedule chart formerly known as Gantt chart.


ng/src/web/components/chart/schedule.js
```diff
@@ -0,0 +1,358 @@
...
+      stopColor={Theme.lightGreen}
+      stopOpacity="0.1"
+    />
+  </LinearGradient>
+);
+
+const fillGradientUrl = `url(#${FILL_GRADIENT_ID})`;
+
+const TIRANGLE_WIDTH = 20;
+
```
swaterkamp: Please change to TRIANGLE_WIDTH. Same for the 5th line below.

PR #438, opened 2018-03-07
Remove css

swaterkamp:
Part of #17


ng/src/web/components/form/formgroup.js
```diff
@@ -48,38 +90,33 @@ const FormGroup = ({
...
   return (
-    <Layout
+    <FormGroupLayout
       flex
       align={['start', 'center']}
-      className={className}>
-      {title}
-      <Layout
+      className={className}
+      >
```
bjoernricks: Could you remove two spaces?

ng/src/web/components/form/formgroup.js
```diff
@@ -48,38 +90,33 @@ const FormGroup = ({
...
-  }
   return (
-    <Layout
+    <FormGroupLayout
       flex
       align={['start', 'center']}
-      className={className}>
-      {title}
-      <Layout
+      className={className}
```
bjoernricks: Is passing the className to FormGroup still necessary?

ng/src/web/components/form/formgroup.js
```diff
@@ -48,38 +90,33 @@ const FormGroup = ({
...
+          {title}
+        </Title>
+      }
+      <FormGroupContent
         {...other}
         flex={flex}
-        className={css}>
+        offset={offset}
+        size={size}
+        >
```
bjoernricks: Here are also two spaces to much for indentation.

ng/src/web/components/form/formgroup.js
```diff
@@ -22,14 +23,55 @@
...
+  paddingLeft: '10px',
+  paddingRight: '10px',
+} : null,
+  ({offset}) => is_defined(offset) ? {
+    marginLeft: COLUMNS[parse_int(offset) - 1],
+  } : null,
+);
 
 const FormGroup = ({
     children,
```
bjoernricks: The indentation of the props of FormGroup could be fixed too.

ng/src/web/components/form/textarea.js
```diff
@@ -22,20 +23,41 @@
...
+  padding: '1px 8px',
+}, ({disabled}) =>
+    disabled ? {cursor: 'not-allowed'} : null,
+    ({disabled, readonly}) => readonly || disabled ? {
+      backgroundColor: Theme.dialogGray,
+      opacity: 1,
+    } : null,
+  );
+
 const TextAreaComponent = ({className, value = '', ...props}) => {
```
bjoernricks: IMHO the TextAreadComponent is unnecessary now. You could remove it and use StyledTextAread instead.

ng/src/web/pages/tasks/dialog.js
```diff
@@ -362,8 +362,8 @@ const TaskDialog = ({
             <Layout
               flex="column"
               offset="2"
-              className={classes('offset-container',
-                state.in_assets === YES_VALUE ? '' : 'disabled')}>
+              className={'offset-container ' +
```
bjoernricks: I don't think this is better then before. Maybe we should introduce a new OffsetLayout class? The offset-container css class should be removed in future too.

ng/src/web/components/icon/greenboneicon.js
```diff
@@ -24,14 +24,13 @@
...
 import _ from 'gmp/locale.js';
-import {classes} from 'gmp/utils';
 
 import PropTypes from '../../utils/proptypes.js';
 
 import Icon from './icon.js';
 
 const GreenboneIcon = ({className, size = 'default', ...props}) => {
-  className = classes(className, 'greenbone-icon');
+  className = className + ' greenbone-icon';
```
bjoernricks: Just remove className completely. We shouldn't add class names anymore. Only for debugging purposes in glamorous components.

bjoernricks: It doesn't make sense to replace classses helper function by string concatenation. classes is still better and easier to use then using plain strings.

ng/src/web/components/form/submitbutton.js
```diff
@@ -23,14 +23,12 @@
...
 
-import {classes} from 'gmp/utils';
-
 import PropTypes from '../../utils/proptypes.js';
 
 import Button from './button.js';
 
 const SubmitButton = ({className, ...other}) => {
-  className = classes('button-block', className);
+  className = className + ' button-block';
```
bjoernricks: Same here. Replacing classes doesn't make sense.

ng/src/web/components/form/selectelements.js
```diff
@@ -46,6 +46,11 @@ export const Box = glamorous.div({
 } : null);
 
 export const Input = glamorous.input({
+  /* use font and line settings from parents not from browser default */
+  fontFamily: 'inherit',
+  fontSize: 'inherit',
+  lineHeight: 'inherit',
+
```
bjoernricks: Just curious why is this necessary?

swaterkamp: The old .css had those three settings for all ```<input>```, ```<textarea>``` and ```<select>``` elements. To not lose these, I added them here.

bjoernricks: Yeah but the original css was derived from bootstrap and relied on other css formattings. I am really not sure if that is necessary at all anymore. Even I am unsure about the original purpose.

swaterkamp: Leaving it out didn't make an obvious difference for the formgroups whose function I checked during programming. But then again, I didn't check all dialogs and use cases of the formgroups especially not with changed fonts etc. I can easily just delete it. Those settings are copied to some other files of this PR as well. Should I also delete those?

ng/src/web/components/form/formgroup.js
swaterkamp: It has been during some changes I made before, but in the end, it's just a leftover. I removed it now.

PR #455, opened 2018-03-13
Dialog clean-up

swaterkamp:


ng/src/web/pages/schedules/dialog.js
```diff
@@ -70,48 +63,32 @@ const TimeUnitSelect = props => {
...
+  schedule,
+  timezone = 'UTC',
+  title = _('New Schedule'),
+  visible = true,
+  onClose,
+  onSave,
+}) => {
+
+  const data = {
+    ...DEFAULTS,
```
bjoernricks: why is DEFAULTS still necessary? Why is name not a prop of ScheduleDialog?

swaterkamp: It hasn't been before, I think, and so I just adopted that. Come to think of it, it clearly can be a prop.

ng/src/web/pages/tags/component.js
```diff
@@ -167,25 +167,35 @@ class TagComponent extends React.Component {
     if (is_defined(tag)) {
       const {resource = {}} = tag;
       this.setState({
+        active: tag.active,
+        comment: tag.comment,
         dialogVisible: true,
+        fixed: tag.fixed,
```
bjoernricks: I don't think fixed is a property of the tag. I guess fixed is used for displaying the dialog with a fixed part. Could you check this? Maybe I am wrong.

ng/src/web/pages/schedules/dialog.js
```diff
@@ -79,13 +76,14 @@ const ScheduleDialog = ({
 }) => {
 
   const data = {
-    ...DEFAULTS,
     ...schedule,
```
bjoernricks: Are there any properties beside id used from the schedule not already passed via props to ScheduleDialog? If not maybe we can avoid passing the schedule to the ScheduleDialog at all by adding id to the props.

ng/src/web/pages/filters/dialog.js
```diff
@@ -61,8 +60,8 @@ const FilterDialog = ({
       onClose={onClose}
       onSave={onSave}
       defaultValues={{
-        ...DEFAULTS,
         ...filter,
```
bjoernricks: Same as for the ScheduleDialog. I suppose only id is used from the filter here. Then passing the id explicitly would be better.

ng/src/web/pages/tags/component.js
swaterkamp: That seems to be the case, indeed. I changed that and don't put fixed into the state anymore.

ng/src/web/pages/schedules/dialog.js
swaterkamp: It seems to be the id only, so I deleted passing the whole schedule and replaced it with just the id.

ng/src/web/pages/filters/dialog.js
swaterkamp: In this case, id and comment were used from the filter. It still seems more efficient to just pass those two props instead of the whole filter, so I changed it.

PR #474, opened 2018-03-23
Add handling of Git Revisions in CMakeLists.txt.

jjnicola:


bjoernricks:
@wiegandm ah you are right. Did only look for includes. But in that case I really would have get this code reworked. It contains several lines of duplicated code.

CMakeLists.txt
```diff
@@ -40,6 +40,32 @@ if (NOT CMAKE_BUILD_TYPE)
   set (CMAKE_BUILD_TYPE Debug)
 endif (NOT CMAKE_BUILD_TYPE)
 
+## Retrieve git revision (at configure time)
+find_program (GIT_EXECUTABLE git DOC "git command line client")
```
bjoernricks: Would be better to use FindGit instead (https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindGit.cmake)

wiegandm:
@bjoernricks Note that the code at the moment is used both at configuration time (to populate `CPACK_PACKAGE_VERSION` and establish `GIT_REVISION` is retrievable) and at build time (by having a target which is always build if `GIT_REVISION` was retrievable and creates the header file used by the C code.

The revisions gathered at configuration and build time may or may not deviate depending on the work flow of the developer.

I'm definitely not opposed to refactoring, but the dual use of the current functionality should be kept in mind.

PR #485, opened 2018-03-28
Add a component to calculate height and width automatically

bjoernricks:
The new component can be used to provide the available width and height
for a sub-component.

### DIFF
--- a/ng/src/web/components/layout/autosize.js
+++ b/ng/src/web/components/layout/autosize.js
@@ -1,0 +1,103 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
+ * Björn Ricks <bjoern.ricks@greenbone.net>
+ *
+ * Copyright:
+ * Copyright (C) 2018 Greenbone Networks GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


ng/src/web/components/layout/autosize.js
```diff
@@ -0,0 +1,106 @@
...
+ * to a child function. width and height are re-calculated on each render of
+ * AutoSize.
+ *
+ * This component uses the render props pattern.
+ */
+class AutoSize extends React.Component {
+
+  static propTypes = {
+    children: PropTypes.func.isRequired,
+    initial: PropTypes.bool,
```
swaterkamp: What does this PropType do? It doesn't seem to be used in the component.

ng/src/web/components/layout/autosize.js
```diff
@@ -0,0 +1,106 @@
...
+    };
+  }
+
+  render() {
+    const {children} = this.props;
+
+    const {width, height} = this.getSize();
+
+    // only call children if height and width are defined
+    const shouldRun = is_defined(height) && is_defined(width);
```
swaterkamp: Could you rename this variable? "Run" seems a bit odd at this place. Maybe shouldCallChildren or something.

PR #487, opened 2018-03-28
Theme font settings

bjoernricks:


ng/src/web/utils/theme.js
```diff
@@ -55,6 +55,11 @@ const Theme = {
     higher: 1,
     default: 0,
   },
+
+  Font: {
+    default: 'small Verdana, sans-serif',
```
swaterkamp: As "small Verdana" doesn't seem to exist and is renamed to "Verdana" in the browser by default, I would exclude the "small" from the definition in Theme.

PR #494, opened 2018-03-29
Fix creating/deleting port ranges for better UX

swaterkamp:


ng/src/web/pages/portlists/dialog.js
```diff
@@ -52,11 +52,12 @@ const PortListsDialog = ({
   name = _('Unnamed'),
   port_list,
   port_range = 'T:1-5,7,9,U:1-3,5,7,9',
+  tmp_port_ranges = [],
```
bjoernricks: Could you rename tmp_port_ranges to only port_ranges? The dialog doesn't know if the port_ranges are temporary  :smiley:

ng/src/web/pages/portlists/dialog.js
```diff
@@ -76,8 +72,11 @@ const PortListsDialog = ({
   );
 
   const data = {
-    ...DEFAULTS,
     ...port_list,
```
bjoernricks: Would be better to pass the necessary values from  port_list explicitly. But I guess we only need the id which also should be put in the defaultValues object instead into the values object.

ng/src/web/pages/portlists/dialog.js
```diff
@@ -123,25 +126,30 @@ const PortListsDialog = ({
                       name="from_file"
                       value="0"
                       onChange={onValueChange}
-                      checked={from_file !== '1'}/>
+                      checked={from_file !== '1'}
```
bjoernricks: Could you add a constant variable  for this, like from_file === FROM_FILE?

ng/src/web/pages/portlists/portrangedialog.js
```diff
@@ -37,21 +37,18 @@ import TextField from '../../components/form/textfield.js';
...
+  port_type = 'tcp',
   title = _('New Port Range'),
   visible,
   onClose,
   onSave,
 }) => {
 
   const data = {
-    ...DEFAULTS,
     ...port_list,
```
bjoernricks: Same as in PortListDialog. The properties used from port_list should be explicitly defined.

ng/src/web/pages/portlists/component.js
```diff
@@ -66,16 +73,21 @@ class PortListComponent extends React.Component {
...
       gmp.portlist.get(entity).then(response => {
         const port_list = response.data;
+        this.created_port_ranges = [];
+        this.deleted_port_ranges = [];
         this.setState({
           comment: port_list.comment,
           port_list,
           name: port_list.name,
           portListDialogVisible: true,
+          tmp_port_ranges: port_list.port_ranges,
```
bjoernricks: I would also rename this variable to port_port ranges even if the are only temporary because this is the list with the visible port ranges which should be saved and used at the end.

ng/src/web/pages/portlists/component.js
```diff
@@ -146,6 +146,105 @@ class PortListComponent extends React.Component {
...
+    const {tmp_port_ranges} = this.state;
+    const {
+      port_range_end,
+      port_range_start,
+      port_type,
+    } = values;
+
+    // reject port ranges with missing values
+    if (!port_range_start || !port_range_end) {
+      return Promise.reject(new Error(_('The port range needs numerical ' +
```
bjoernricks: I am not 100% sure if Promise.reject is necessary. It could be possible that `throw new  Error(...) ` is enough.

bjoernricks:
Additionally I would update the commit message for 80732fbc515d2ada854431186886c0538ffce943 in something like "Update PortList command createPortRange to return an action result"

ng/src/web/pages/portlists/component.js
swaterkamp: I tried throwing a new Error before, but that didn't work, so it seems we need to catch the rejected promise.

bjoernricks: Just did a quick look at the corresponding SaveDialog code. Indeed return Promise.reject(new Error is necessary to work as expected.

ng/src/web/pages/portlists/dialog.js
swaterkamp: The dialog doesn't know that, but then again djafodnic would be an equally senseful variablename for the dialog ;) I was aiming at letting the reader know, what this variable is used for and which larger context stands behind it.

ng/src/web/pages/portlists/component.js
```diff
@@ -90,6 +91,7 @@ class PortListComponent extends React.Component {
       this.deleted_port_ranges = [];
       this.setState({
         comment: undefined,
+        id: undefined,
         name: undefined,
         port_list: undefined,
```
bjoernricks: Is the port_list object still used? If not it can be removed from the state.

ng/src/web/pages/portlists/component.js
```diff
@@ -299,6 +302,7 @@ class PortListComponent extends React.Component {
             {portListDialogVisible &&
               <PortListsDialog
                 comment={comment}
+                id={id}
                 name={name}
                 port_list={port_list}
```
bjoernricks: I guess port_list isn't used in the PortListDialog anymore and can be safely removed

ng/src/web/pages/portlists/dialog.js
```diff
@@ -147,7 +149,7 @@ const PortListsDialog = ({
                       name="from_file"
                       value="1"
```
bjoernricks: should better use value={FROM_FILE}

ng/src/web/pages/portlists/component.js
swaterkamp: It's still in use to check whether the dialog is_edit, and the newrangeicon (ll. 70) seems to get more data from the port_list than just the id.

PR #541, opened 2018-04-13
Fix yAxis labels for schedule charts

swaterkamp:
Labels that are too long will not be cut off by the display anymore.
Instead the chart will be moved to the right to make space for longer
labels. If the label exceeds a MAX_LABEL_LENGTH = 25 it will be
shortened.

### DIFF
--- a/ng/src/web/components/chart/schedule.js
+++ b/ng/src/web/components/chart/schedule.js
@@ -2,6 +2,7 @@
  *
  * Authors:
  * Björn Ricks <bjoern.ricks@greenbone.net>
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
  *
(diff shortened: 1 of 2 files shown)


ng/src/web/components/chart/schedule.js
```diff
@@ -50,6 +52,15 @@ const margin = {
   left: 60,
 };
 
+const MAX_LABEL_LENGTH = 25;
+
+const tickFormat = val => {
+  if (val.toString().length > MAX_LABEL_LENGTH) {
```
bjoernricks: the check is not required. shorten only cuts the text if it is longer then the second param see https://github.com/greenbone/gsa/blob/master/ng/src/gmp/utils/string.js#L38

PR #542, opened 2018-04-15
Update module documentation.

janowagner:


gsad/doc/gsad.8.xml
```diff
@@ -28,26 +28,18 @@
...
-      various management functionalities to handle and organise scan
-      results.
-    </p>
-
-    <p>
-      The Greenbone Security Assistant Daemon (gsad) is a web
-      application that connects to the Greenbone Vulnerability Manager to provide for a
-      full-featured user interface for vulnerability management.
+      The Greenbone Security Assistant Daemon (gsad) is a stand-alone web server
+      that acts as the layer between the single page application Greenbone
```
bjoernricks: Could we say front-end application instead of single page application? For me single page application is only a technical detail.

PR #568, opened 2018-04-24
Fix multiple column sort in detached charts

timopollmeier:
To make sorting by multiple columns work in detached charts, also
generate the attributes data-sort-fields, data-sort-stats and
data-sort-orders from the request parameters, not just their singular
versions data-sort-field, data-sort-stat and data-sort-order.

### DIFF
--- a/src/html/classic/omp.xsl
+++ b/src/html/classic/omp.xsl
@@ -9648,6 +9648,30 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
     </xsl:if>
   </xsl:variable>
 
+  <xsl:variable name="sort_fields">
+    <xsl:if test="/envelope/params/_param[starts-with (name, 'sort_fields:')]">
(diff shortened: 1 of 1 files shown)


src/html/classic/omp.xsl
```diff
@@ -9694,6 +9718,9 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
       data-sort-field="{/envelope/params/sort_field}"
       data-sort-stat="{/envelope/params/sort_stat}"
       data-sort-order="{/envelope/params/sort_order}"
```
bjoernricks: @timopollmeier shouldn't these old data-sort properties be removed?

timopollmeier: These are still used by some charts that only require sorting by one field.

bjoernricks: Ah overlooked the s at the end of the properties. Thanks for clarification!

PR #586, opened 2018-04-27
Update help pages

timopollmeier:
This PR is for various updates to the help pages, including:
* Updating the copyright and version on the about page
* Updating the German help pages


timopollmeier:
The commit for the German help page wasn't pushed yet when this was merged, so that part is actually done in #587

PR #588, opened 2018-04-27
OVAL Definitions Dashboard

swaterkamp:


ng/src/web/components/dashboard2/display/utils.js
```diff
@@ -137,4 +137,22 @@ export const qodTypeColorScale = scaleOrdinal()
...
+export const OVAL_CLASS_TYPES = {
+  compliance: _('Compliance'),
+  inventory: _('Inventory'),
+  miscellaneous: _('Miscellaneous'),
+  patch: _('Patch'),
+  vulnerability: _('Vulnerability'),
+};
+
+export const ovalClassColorScale = scaleOrdinal()
+  .domain(Object.keys(OVAL_CLASS_TYPES))
```
bjoernricks: don't forget to add an `import 'core-js/fn/object/keys'` statement for ie 11 support

swaterkamp: It's included already because of using the QOD_TYPES

PR #668, opened 2018-05-16
Add option to enable support for coverage analysis

wiegandm:
This commit adds a new CMake option called `ENABLE_COVERAGE`. When
activated, this option will cause the `--coverage` option to passed to
the C compiler. This option is used to compile and link code
instrumented for coverage analysis.

Running code built with `--coverage` will result is coverage data and
notes being recorded and stored in the build directory. The recorded
data can then be consumed by `gcov` and related tools.

The `ENABLE_COVERAGE` option is only supported for the `Debug` build
type as compiler optimizations performed for other build types would
render the coverage measurements less meaningful.


bjoernricks:
I am going to close this PR because gsad will not get any documentation in future.

wiegandm:
> I am going to close this PR because gsad will not get any documentation in future.

Note that this PR was about code coverage, not documentation. But I do accept your point that measuring the C test coverage does not make sense if no unit tests for the C code are forthcoming.

PR #675, opened 2018-05-17
Donut labels

swaterkamp:


ng/src/web/components/chart/donut3d.js
```diff
@@ -303,75 +391,100 @@ class Donut3DChart extends React.Component {
       innerRadiusX,
       innerRadiusY,
     };
+    this.allLabels = [];
```
bjoernricks: just use `const allLabels = []`

ng/src/web/components/chart/donut3d.js
```diff
@@ -303,75 +391,100 @@ class Donut3DChart extends React.Component {
...
-                        </Label>
-                      </Group>
-                    )}
-                  </ToolTip>
-                );
-              }}
-            </Pie> :
+                        {({targetRef, hide, show}) => {
+                          if (angleAbs > MIN_ANGLE_FOR_LABELS) {
+                            this.allLabels.push({
```
bjoernricks: Use allLabels instead of this.allLabels here

ng/src/web/components/chart/donut3d.js
```diff
@@ -303,75 +391,100 @@ class Donut3DChart extends React.Component {
...
+                      </ToolTip>
+                    );
+                }}
+              </Pie>
+              <Group
+                top={centerY}
+                left={centerX}
+              >
+                <AllLabels
+                  allLabels={this.allLabels}
```
bjoernricks: Use allLabels instead of this.allLabels

ng/src/web/components/chart/donut3d.js
```diff
@@ -224,13 +230,37 @@ EmptyDonut.propTypes = {
...
+};
+
 class Donut3DChart extends React.Component {
 
   constructor(...args) {
     super(...args);
 
     this.legendRef = React.createRef();
 
+    this.allLabels = [];
```
bjoernricks: Can be removed by using const allLabels in the render method.

PR #694, opened 2018-05-24
Fix FilterTypes for SecInfos

swaterkamp:
All SecInfo items use the filtertype 'info' instead of having an own
type such as e.g. 'nvt'.


bjoernricks:
I am not sure if we should keep NVTS_FILTERS_FILTER, CVES_FILTER_FILTER, ... variables and instead just use INFO_FILTERS_FILTER.

PR #701, opened 2018-05-25
Multiple Permissions Dialog

swaterkamp:


ng/src/web/pages/permissions/multipledialog.js
```diff
@@ -56,11 +56,11 @@ const MultiplePermissionDialog = ({
   group_id,
   groups = [],
   include_related,
-  permission,
+  permission = 'read',
   related = [],
```
swaterkamp: The redundancy of these defaults with the DEFAULTS constant was recognized later on and is removed in the "Update coding style" commit 2aa82a997404867ec6d753f7da5447e4e68e40e9

ng/src/web/pages/permissions/multipledialog.js
```diff
@@ -64,7 +64,7 @@ const MultiplePermissionDialog = ({
   title = _('Create Multiple Permissions'),
   user_id,
   users = [],
-  visible,
+  visible = true,
```
bjoernricks: Can be removed. It should be obsolete now.

PR #708, opened 2018-06-04
Add CI job for checking documentation coverage of gsad

davidak:
This commit adds a CI job which generates the source code documentation
with Doxygen as XML and then analyzes the documentation coverage with
doxy-coverage.

Errors reported during the Doxygen run and the coverage results
themselves are stored as build artifacts.

Should the job name be more precise?


bjoernricks:
Adding a coverage job for gsad doesn't make sense imho. gsad itself will not get any documentation because its questionable if gsad will survive the next year.

bjoernricks:
Same as #668

PR #709, opened 2018-06-04
Remove support for severity class OpenVAS Classic

mattmundell:
This severity class is no longer used provided by Manager.


mattmundell:
Requires GVM PR #123.

PR #760, opened 2018-06-29
added systemd units

GIJack:
Systemd UNIX for gsad.

gsad.default gets moved to /etc/default/gsad

gsad.logrotate goes to /etc/logrotate.d/gsad

systemd service is self explanatory.

### DIFF
--- a/systemd/gsad.default
+++ b/systemd/gsad.default
@@ -1,0 +1,35 @@
+OPTIONS=""
+
+#
+# The address the Greenbone Security Assistant will listen on.
+#
+
+GSA_ADDRESS=127.0.0.1
+
+#
+# The port the Greenbone Security Assistant will listen on.
+#
+
+GSA_PORT=9392
+R_PORT=80 # redirect from port
+
+# NOT USED BY DEFAULT IN OpenVAS 9
+#
+# The address the OpenVAS Manager is listening on.
+#
+
+#MANAGER_ADDRESS=127.0.0.1
+
+#
+# The port the OpenVAS Manager is listening on.
+#
+
+#MANAGER_PORT=9390
+
+## Determines ciphers that GSAD will use with TLS and their order.
+# Work around with gnutls prorities not working
+GNUTLS_PRIORITIES="NORMAL"
+
+# This is the new standard. DOES NOT WORK BECASUE OF BUG
+#GNUTLS_PRIORITIES="NONE:+SECURE256:+SECURE192:+SUITEB192:+VERS-TLS1.2"
+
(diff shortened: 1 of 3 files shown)


bjoernricks:
@janowagner the systemd PRs have been created on my recommendation. I have talked to the GOS team before and they haven't found any objections against including systemd service files. Even the GSF guys are using self created service files.

So I am highly in favor to include systemd service files for all of our daemons to get upstream service file references.

bjoernricks:
The discussion about service files is still ongoing at https://github.com/greenbone/gvm/pull/142

shibumi:
Hi, I would like to get this merged. I am the Arch Linux Maintainer for OpenVAS. I really don't think that this is just a 'packaging aspect'. One reason why the linux distributions have adopted `systemd` was that init scripts would be the same across all distributions. Furthermore upstream is now able to manage these systemd service files. If you include systemd service files in your project you can:

1. r**educe bug reports**, because you know that OpenVAS is started on all systems the same way. If something non default happens, you can close the bug report and refer to the systemd service files that are supported by upstream.
2. **harden your services**. You are able to harden your security application against malicious behavior using cgroups and namespaces (both supported by the linux kernel). If all distributions would do harden the service files by themselves you can expect more work for your team, because of more bug reports in the future regarding system startup or weird behavior of your application.

bjoernricks:
@shibumi As with the gsa maintainer hat on I am completely in favor of merging service files for gsa. If someone is willing to update this PR (or to create a new one), put the service files under the gsad directory and install them via cmake to the standard dirs in the prefix I am going to merge the pr.

cfi-gb:
@shibumi @bjoernricks 

Keep in mind that the systemd files provided within this PR is still distro-specific as there is no such directory like (from gsad.service):

> /etc/sysconfig

available on Debian/Ubuntu based systems. Furthermore running the gsad daemon as root is probably also highly discouraged.

bjoernricks:
> @shibumi @bjoernricks
> 
> Keep in mind that the systemd files provided within this PR is still distro-specific as there is no such directory like (from gsad.service):
> 
> > /etc/sysconfig
> 
> available on Debian/Ubuntu based systems. Furthermore running the gsad daemon as root is probably also highly discouraged.

Yeah that's why I wanted to get a more generalized service file. Not sure if it can fit for all distros but we should have a strong opinion here.

GIJack:
On some systems its /etc/sysconfig, on some systems its /etc/default/. Either case, the file contents would be the exact same, and the path can be edited by system maintainers.

bjoernricks:
I've added a replacement for this PR in https://github.com/greenbone/gsa/pull/1486

#1486 is not intended as a "final" version. Just as a start to unify the service and config files. Feel free to create additional pull requests to improve the files.

PR #765, opened 2018-07-03
Add result hostname to list pages

timopollmeier:
This adds the hostname to the "Host" column of the results list and the
report details and a new filter option for it.
The column header now has two subcolumns: IP and Hostname to easily sort
by both.

### DIFF
--- a/gsa/src/web/pages/results/row.js
+++ b/gsa/src/web/pages/results/row.js
@@ -23,6 +23,8 @@
 
 import React from 'react';
 
+import glamorous from 'glamorous';
+
 import {longDate} from 'gmp/locale/date';
 
 import {is_defined} from 'gmp/utils/identity';
@@ -42,6 +44,10 @@ import TableRow from '../../components/table/row.js';
 import TableData from '../../components/table/data.js';
 
 import ResultDelta from './delta.js';
+
+const Hostname = glamorous.span({
+  fontSize: '10px',
+});
 
 const Row = ({
   actions,
@@ -83,7 +89,7 @@ const Row = ({
       <TableData flex align="center">
         {entity.qod.value} %
       </TableData>
-      <TableData flex align="center">
+      <TableData colSpan="2">
         <DetailsLink
           type="host"
           id={host.id}
@@ -91,6 +97,9 @@ const Row = ({
         >
           {host.name}
         </DetailsLink>
+        {host.hostname &&
+          <Hostname>({host.hostname})</Hostname>
+        }
       </TableData>
       <TableData>
         {entity.port}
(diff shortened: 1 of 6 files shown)


gsa/src/web/pages/results/row.js
```diff
@@ -83,14 +89,17 @@ const Row = ({
...
-      <TableData flex align="center">
+      <TableData colSpan="2">
         <DetailsLink
           type="host"
           id={host.id}
           textOnly={!links}
         >
           {host.name}
         </DetailsLink>
+        {host.hostname &&
```
bjoernricks: You have defined `host.hostname` in models/results.js to be always a string. So this check should be `host.hostname.length > 0`

gsa/src/web/pages/results/row.js
```diff
@@ -43,6 +45,10 @@ import TableData from '../../components/table/data.js';
 
 import ResultDelta from './delta.js';
 
+const Hostname = glamorous.span({
+  fontSize: '10px',
```
bjoernricks: What's the reason behind this fontSize? should hostname be smaller then other columns in the row? it's better to avoid fixed font sizes normally.

gsa/src/web/pages/results/row.js
```diff
@@ -46,7 +46,7 @@ import TableData from '../../components/table/data.js';
 import ResultDelta from './delta.js';
 
 const Hostname = glamorous.span({
-  fontSize: '10px',
+  fontSize: '0.8em',
```
swaterkamp: Instead of decreasing font size, you could use gmp/utils/string.js's shorten() to cut off names that are too long.

gsa/src/web/pages/results/row.js
```diff
@@ -60,6 +56,16 @@ const Row = ({
   const {host} = entity;
   const shown_name = is_defined(entity.name) ? entity.name : entity.nvt.oid;
   const has_tags = is_defined(entity.nvt) && is_defined(entity.nvt.tags);
+  var hostname_elem;
+  if (host.hostname.length > 40)
```
bjoernricks: You don't have to check the length here. Just render
```jsx
<TableRow>
....
<span title={host.hostname}>
  ({shorten (host.hostname, 40)})
</span>
```
below.
`<span>...</span>` is the same as `<span title={undefined}>....</span>`
and `shorten(undefined)` returns an empty string. Thus your conditions here can be omitted.

gsa/src/web/pages/results/row.js
```diff
@@ -103,7 +94,11 @@ const Row = ({
         >
           {host.name}
         </DetailsLink>
-        {hostname_elem}
+        {host.hostname.length > 0
```
bjoernricks: You could still remove this condition but it's fine.

PR #778, opened 2018-07-05
Fix bulk actions delete and export

swaterkamp:
In order to fully function the previously deleted process_bulk in gsad needs to be restored and renamed to bulk_export.

### DIFF
--- a/gsa/src/gmp/commands/entities.js
+++ b/gsa/src/gmp/commands/entities.js
@@ -141,8 +141,8 @@ class EntitiesCommand extends HttpCommand {
     // FIXME change gmp to allow deletion by filter
     let deleted;
     return this.get({filter}).then(entities => {
-      deleted = entities.getEntries();
-      return this.delete(entities, extra_params);
+      deleted = entities.data;
+      return this.delete(deleted, extra_params);
     }).then(response => response.setData(deleted));
   }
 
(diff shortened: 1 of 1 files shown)


gsa/src/gmp/commands/entities.js
```diff
@@ -120,7 +120,8 @@ class EntitiesCommand extends HttpCommand {
   }
 
   delete(entities, extra_params) {
-    return this.deleteByIds(map(entities, entity => entity.id), extra_params)
```
bjoernricks: The old code was correct. `entities` should be an Array of ids.

gsa/src/gmp/commands/entities.js
```diff
@@ -141,7 +142,7 @@ class EntitiesCommand extends HttpCommand {
     // FIXME change gmp to allow deletion by filter
     let deleted;
     return this.get({filter}).then(entities => {
-      deleted = entities.getEntries();
+      deleted = entities.data;
       return this.delete(entities, extra_params);
```
bjoernricks: should be
```js
return this.delete(deleted, extra_params);
```
now

gsa/src/gmp/commands/entities.js
```diff
@@ -92,12 +92,12 @@ class EntitiesCommand extends HttpCommand {
   }
 
   export(entities) {
-    return this.exportByIds(map(entities, entity => entity.id));
+    return this.exportByIds(map(entities.data, entity => entity.id));
```
bjoernricks: Same as delete. `entities` should still be an Array of ids. The calling code should be changed instead.

PR #780, opened 2018-07-06
Code cleanup

bjoernricks:


gsad/src/gsad_gmp.c
```diff
@@ -423,8 +414,8 @@ filter_exists (gvm_connection_t *connection, const char *filt_id)
   entity_t entity;
 
   if (filt_id == NULL
-      || strcmp (filt_id, "0")
-      || strcmp (filt_id, "-2"))
+      || str_equal (filt_id, "0")
+      || str_equal (filt_id, "-2"))
```
timopollmeier: I think you forgot to negate the results here.
Also, you could use FILT_ID_NONE and FILT_ID_USER_SETTING here.

gsad/src/gsad_http.c
```diff
@@ -804,22 +806,22 @@ attach_remove_sid (http_response_t *response, const gchar *sid)
 {
   if (sid)
     {
-      if (strcmp (sid, "0"))
+      if (str_equal (sid, "0"))
```
timopollmeier: I think there's another negation missing here.

gsad/src/gsad_gmp.c
```diff
@@ -2014,12 +1598,12 @@ get_many (gvm_connection_t *connection, const char *type,
           else
             filt_id = FILT_ID_USER_SETTING;
         }
-      else if ((strcmp (filter, "sort=nvt") == 0)
-               && ((strcmp (type, "note") == 0)
-                   || (strcmp (type, "override") == 0)))
+      else if (str_equal (filter, "sort=nvt")
+               && (str_equal (type, "note")
+                   || str_equal (type, "override") == 0))
```
timopollmeier: It looks like you accidentally left the == 0 in here.

gsad/src/gsad_gmp.c
```diff
@@ -2014,12 +1598,12 @@ get_many (gvm_connection_t *connection, const char *type,
...
         filt_id = FILT_ID_USER_SETTING;
-      else if ((strcmp (filter, "apply_overrides=1") == 0)
-               && (strcmp (type, "task") == 0))
+      else if (str_equal (filter, "apply_overrides=1")
+               && str_equal (type, "task") == 0)
```
timopollmeier: Same here

gsad/src/gsad_gmp.c
bjoernricks: Thanks for the hint! Didn't know we introduced those defines already. Nice! I'll check the condition.

gsad/src/gsad_http.c
bjoernricks: If have changed the sequence of the clauses. Should be fine here.

gsad/src/gsad_gmp.c
bjoernricks: You are right. I'll fix it.

gsad/src/gsad_gmp.c
bjoernricks: Regarding the condition. IMHO it was broken before and my change fixed it. Can you take another look?

gsad/src/gsad_gmp.c
bjoernricks: All fixed now. Btw. I am not sure if all conditions in this path are correct. The code is confusing to me

gsad/src/gsad_gmp.c
timopollmeier: I've checked the context again and it looks like the condition was wrong.

PR #794, opened 2018-07-13
Improve Tags handling for entity types and entities loading

bjoernricks:
Improve several tag related components for being able to load the entities from the store at the end.
Also update the components to allow to change the entity type naming scheme more easily in future.

### DIFF
--- a/gsa/src/web/entities/tagsdialog.js
+++ b/gsa/src/web/entities/tagsdialog.js
@@ -62,7 +62,7 @@ class TagsDialog extends React.Component {
       resourceType,
     } = this.props;
 
-    gmp[pluralizeType(normalizeType(resourceType))].getAll({filter})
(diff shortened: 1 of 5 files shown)


gsa/src/web/entities/tagsdialog.js
```diff
@@ -125,7 +120,12 @@ class TagsDialog extends React.Component {
             <FormGroup title={_('Comment')} titleSize="4">
               {comment}
             </FormGroup>
-            {noticeText}
+            {entitiesCount >= ENTITIES_THRESHOLD &&
+              <span>
+                {_('Please note that assigning a tag to a large number of ' +
+                  'resources may take several minutes.')}
```
swaterkamp: Please rename resources to items to be consistent with other wordings.

PR #800, opened 2018-07-16
Update url for http gmp api calls

bjoernricks:
Use /gmp?token=... instead of the old /omp?token=... now.


bjoernricks:
We don't have to wait for secinfo to be updated. The link won't work with ng anyway.

Issue #643, opened 2018-05-09
Migrate to other CSS-in-JS library


bjoernricks:
It seems glamorous will be deprecated https://github.com/paypal/glamorous/issues/419
So we need to investigate howto migrate to another CSS-in-JS solution like https://emotion.sh/

bjoernricks:
The other possible solution is https://github.com/styled-components/styled-components

TejasQ:
I wrote a codemod that'll automatically migrate your codebase to emotion. Maybe it could help you? https://github.com/TejasQ/babel-plugin-glamorous-to-emotion

bjoernricks:
Disadvantages of styled-components:
* Object API (`styled.div({...})`) is new in 3.3.0 and undocumented https://github.com/styled-components/styled-components/pull/1732
* I am not sure if returning an object with styles from a function is working as expected currently
https://github.com/styled-components/styled-components/pull/1820
* It hasn't an api to filter props => all props are put into the dom
* It's not possible to define an extra classname without using [attrs](https://www.styled-components.com/docs/basics#attaching-additional-props) or an extra component to filter these props
* Difficult to understand the codebase

Advantages:
* Great community
* Referencing of components in styles https://www.styled-components.com/docs/advanced#referring-to-other-components
* Only one package to add as dependency
* [tagged-template-literals](https://www.styled-components.com/docs/advanced#tagged-template-literals) API

swaterkamp:
After some considerations I'd opt for styled-components over emotion. I can't get rid of the feeling that emotion will face the same fate as glamorous after a while, despite it's increasing prevalence. Styled-Components just seems to be the more stable choice and it should be able to provide everything we need one way or the other.

PR #852, opened 2018-08-03
Use styled-components for the menu

swaterkamp:
The Menu is changed in design and styled-components is used instead of glamorous.

Some minor fixes are also performed for Badge and logging color as well as translatability for the TitleBar.

This is the first commit regarding issue #643.


bjoernricks:
I guess with this PR we can also get rid of sticky.js?

swaterkamp:
Yes, it will probably be removed in the next PR.

PR #860, opened 2018-08-08
Fix ResourceList and UserSelection for Tags

swaterkamp:
- Fix proptype warning
- Fix selection of bulk action targets with checkboxes
- update Tags model
- update ResourceList (new component, improved loading)
- update Tags detailspage
- camelCase and GSA-type in tag model
- withRouter() for EntityContainer

### DIFF
--- a/gsa/src/gmp/models/tag.js
+++ b/gsa/src/gmp/models/tag.js
@@ -25,7 +25,6 @@
 import Model from '../model.js';
 
 import {isDefined} from 'gmp/utils/identity';
-import {map} from 'gmp/utils/array';
 
 class Tag extends Model {
 
@@ -34,13 +33,10 @@ class Tag extends Model {
   parseProperties(elem) {
     const ret = super.parseProperties(elem);
     if (isDefined(elem.resources)) {
-      ret.resources = map(ret.resources.resource,
-        res => new Model(res, elem.resources.type));
       ret.resource_type = elem.resources.type;
       ret.resource_count = elem.resources.count.total;
     }
     else {
-      ret.resources = [];
       ret.resource_count = 0;
     }
     return ret;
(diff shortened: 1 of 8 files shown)


gsa/src/web/pages/tags/resourcelist.js
```diff
@@ -0,0 +1,124 @@
...
+
+  constructor(...args) {
+    super(...args);
+
+    this.state = {res: []};
+  }
+
+  componentDidMount() {
+    const {gmp, entity} = this.props;
+    const {id, resource_type} = entity || {};
```
bjoernricks: I am not sure about this. I suppose if no entity is passed then no resources should be loaded. So either guard it with 
```javascript
if (isDefined(entity))
```
or even require the entity in the proptypes.
For both cases `|| {}` should be not used.

gsa/src/web/pages/tags/resourcelist.js
```diff
@@ -0,0 +1,124 @@
...
+    const filter = 'tag_id="' + id + '" rows=' + MAX_RESOURCES;
+    gmp[pluralizeType(normalizeType(resource_type))].get({filter})
+    .then(resources => this.setState({res: resources.data}));
```
bjoernricks: Would be nice to indent the `.then` line

gsa/src/gmp/models/tag.js
```diff
@@ -34,13 +33,10 @@ class Tag extends Model {
   parseProperties(elem) {
     const ret = super.parseProperties(elem);
     if (isDefined(elem.resources)) {
-      ret.resources = map(ret.resources.resource,
-        res => new Model(res, elem.resources.type));
       ret.resource_type = elem.resources.type;
       ret.resource_count = elem.resources.count.total;
```
bjoernricks: Would be nice to change resource_type and resource_count to camelCase and change resourceType to return a GSA type (by using `normalizeType` here) in a next step.

PR #865, opened 2018-08-09
Session handling

bjoernricks:
Renew user session only if:

* the page (in react router terms location) has changed
* a POST HTTP request has been send to gsad

Missing renews:

* a user opens a dialog (~~should the session be renewed in this case?~~ Yes it should)
    * ~~Task dialogs~~
* ~~the pagination is used at the list pages~~
* ~~filter is changed at the list pages~~
* ~~Dashboard changes~~
* Entity is changed on details pages

Missing features:

* OBSOLETE ~~session timeout isn't updated in the redux store after POST requests~~ Web code should always renew the session via api explicitly


swaterkamp:
Opening a dialog is a user action indicating that she is active. This should imho renew the session just like using the pagination and a filter change should. I don't know right now, though, how much of an effort implementing this would be and if it's ultimately worth it...

bjoernricks:
@swaterkamp thanks for you opinion. It's not much more effort. I'll add another action creator function that renews the session and updates the session timestamp in the store. This action creator function has to be called in all open dialog methods afterwards.

PR #871, opened 2018-08-15
Fix TagDialog

swaterkamp:
Fixing and changing include:
- Load resources according to new API (resources are no longer listed in the Tag-object)
- Change behavior of TagDialog concerning Resource selection:
-- as a get... command yields max. 1000 elements, the resource list of tags with more than 1000 resources was cut when editing/saving. Also: manual selection of resources for tags of a certain size, does not really make sense.
-- The new concept: For tags with just a few resources (<200 atm) don't change behavior. For more than a few resources, do not show the resource selection elements and thereby don't override the original list when saving. If users want to change resources of a huge bulk tag, it is easier to delete the old tag and create a new one with adjusted resources.

### DIFF
--- a/gsa/src/web/pages/tags/component.js
+++ b/gsa/src/web/pages/tags/component.js
@@ -33,16 +33,16 @@ import {getEntityType, typeName} from 'gmp/utils/entitytype';
 
 import {YES_VALUE} from 'gmp/parser';
 
-import PropTypes from '../../utils/proptypes.js';
-import compose from '../../utils/compose';
-import withGmp from '../../utils/withGmp.js';
-import withCapabilities from '../../utils/withCapabilities.js';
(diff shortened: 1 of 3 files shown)


gsa/src/web/pages/tags/component.js
```diff
@@ -131,33 +134,42 @@ class TagComponent extends React.Component {
...
+              dialogVisible: true,
+              id,
+              name,
+              resourceCount,
+              resource_ids: resources.map(res => res.id),
+              resource_type: isDefined(resourceType) ?
+                resourceType :
+                first(resource_types, [])[0],
+              resource_types,
+              resources_action:
```
bjoernricks: The variables should be changed to camelCase in another commit or PR.

PR #881, opened 2018-08-23
Overwrite default filters when loading entities in user settings

swaterkamp:
Also split scanconfigs and scanners into openvas and osp options.

### DIFF
--- a/gsa/src/web/pages/usersettings/usersettingspage.js
+++ b/gsa/src/web/pages/usersettings/usersettingspage.js
@@ -28,7 +28,16 @@ import {connect} from 'react-redux';
 
 import _ from 'gmp/locale';
 
-import {filterEmptyScanConfig} from 'gmp/models/scanconfig';
+import {ALL_FILTER} from 'gmp/models/filter';
+import {
+  filterEmptyScanConfig,
+  openVasScanConfigsFilter,
+  ospScanConfigsFilter,
+} from 'gmp/models/scanconfig';
+import {
+  openVasScannersFilter,
+  ospScannersFilter,
+} from 'gmp/models/scanner';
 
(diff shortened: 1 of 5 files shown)


gsa/src/web/pages/usersettings/usersettingspage.js
```diff
@@ -319,6 +328,15 @@ class UserSettings extends React.Component {
     if (isLoading) {
       return <Loading/>;
     };
+    const openVasScanConfigs = scanconfigs.filter(
+        conf => conf.scan_config_type === OPENVAS_SCAN_CONFIG_TYPE);
+    const ospScanConfigs = scanconfigs.filter(
+      conf => conf.scna_config_type === OSP_SCAN_CONFIG_TYPE);
```
bjoernricks: A little typo scna_config_type instead of scan_config_type

bjoernricks: We should add some filter functions to models/scanconfig.js for this
```js
const ospScanConfigsFilter = config => config.scan_config_type === OSP_SCAN_CONFIG_TYPE;
```
This will hide the access to the scan_config_type property which we are going to rename in future.

gsa/src/web/pages/usersettings/usersettingspage.js
```diff
@@ -319,6 +328,15 @@ class UserSettings extends React.Component {
...
+    const openVasScanners = scanners.filter(
+      scanner => scanner.scanner_type === OPENVAS_SCANNER_TYPE);
+    const ospScanners = scanners.filter(
+      scanner => scanner.scanner_type === OSP_SCANNER_TYPE);
```
bjoernricks: As above with the scan configs. We should add some filter functions to models/scanner.js

PR #901, opened 2018-08-31
Make credential dropdown for target sort visible

swaterkamp:


gsa/src/web/pages/targets/table.js
```diff
@@ -45,6 +48,12 @@ import TableRow from 'web/components/table/row';
 import TargetDetails from './details';
 import TargetRow from './row';
 
+const StyledSelect = styled(Select)`
+  color: ${Theme.black};
+  font-weight: normal;
```
bjoernricks: Shouldn't be color and font-weight put to the Select component directly to make these css values the default? imho the font-weight and color should not be inherited from the parent in the select.

swaterkamp: Inheriting makes sense in case of changing layout colors etc., so for example if we choose to have darkgray as standard font color instead of black. It's just this special case where the inherited color is white, which produces an issue. Therefore, I put it here instead of the Select, but, of course, I can change that if you prefer it that way.

PR #906, opened 2018-09-03
Allow to filter results for solution type via the dialog

swaterkamp:


gsa/src/web/components/powerfilter/solutiontypegroup.js
```diff
@@ -0,0 +1,143 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+import 'core-js/fn/string/includes';
```
bjoernricks: Should be removed. Seems to be a copy&paste leftover.

PR #935, opened 2018-09-12
Add Yarn offline mode option.

tuxmaster5000:
This will be needed, to build it without internet connection.
In this case all needed npm packages can be preloaded into the yarn cache.

### DIFF
--- a/gsa/CMakeLists.txt
+++ b/gsa/CMakeLists.txt
@@ -42,6 +42,10 @@ if (NOT YARN_FOUND)
 else (NOT YARN_FOUND)
   message (STATUS "Found Yarn: ${YARN_EXECUTABLE}")
   set (INSTALLER ${YARN_EXECUTABLE})
+  if (YARN_OFFLINE)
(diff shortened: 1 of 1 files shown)


tuxmaster5000:
Modify INSTALLER will fails with:
/bin/sh: /usr/bin/yarn --offline: No such file or directory
Because on COMMAND of add_custom_command.
Arguments must be given as an separate variable.
See:  https://cmake.org/cmake/help/latest/command/add_custom_command.html

bjoernricks:
Ah thanks for the hint! Wasn't aware of that problem yet. In that case your PR is fine and I would only suggest to use `set (YARN_OFFLINE` instead of `set (INSTALLER_ARGS`. `-DYARN_OFFLINE=1` can be kept.

tuxmaster5000:
Ah I understood. This make sense.

bjoernricks:
Looks good. Only the conflict has to be resolved.

tuxmaster5000:
Now it looks good. :)

PR #936, opened 2018-09-12
Let it work with cmake and cmake3.

tuxmaster5000:
Use the variable for cmake instant of hard coded cmake call.


bjoernricks:
Thanks!

bjoernricks:
@tuxmaster5000 would like like to be named at the https://github.com/greenbone/gsa/blob/master/CHANGES.md file?

tuxmaster5000:
Hi Björn, yes you can do it.

bjoernricks:
@tuxmaster5000 as tuxmaster5000 or another name? We are using real names for the changes normally.

PR #960, opened 2018-09-20
Don't use Boolean for wrap prop of Layouts

swaterkamp:
This concerned Divider and one instance in MultiSelect.


bjoernricks:
Replaced with #961

PR #969, opened 2018-09-21
Point support link to the new community portal

cfi-gb:
Other modules:

greenbone/openvas-scanner/pull/175
greenbone/gvm/pull/192
greenbone/gvm-libs/pull/123
greenbone/openvas-smb/pull/20


bjoernricks:
Thanks! I would love to make the community portal even more prominent in the issue template. Currently 90% of all created issues should have been put into the portal.

PR #971, opened 2018-09-24
Remove dashboard display menues

bjoernricks:


gsa/src/web/components/dashboard/display/datadisplay.js
```diff
@@ -83,6 +80,38 @@ const FilterString = styled.div`
...
+  z-index: ${Theme.Layers.higher};
+  transition: opacity 500ms;
+`;
+
+const DisplayBox = styled.div`
+  display: flex;
+  flex-grow: 1;
+  position: relative;
+
+  & .display-icon-bar {
```
swaterkamp: In styled-components you can directly reference components without the detour with classnames.
So what you can do here, is something like that:
'''
${IconBar}: {
  visibility: hidden;
  opacity: 0;
}

PR #995, opened 2018-10-04
GUI miscellaneous #2

swaterkamp:
* Fix ScanConfigDialog tables
* Fix translating FilterTypes
* Fix CVEs table
* Change titles in Result details
* ... some clean-up

### DIFF
--- a/gsa/src/gmp/models/tag.js
+++ b/gsa/src/gmp/models/tag.js
@@ -23,6 +23,7 @@
  */
 import {normalizeType} from 'gmp/utils/entitytype';
 import {isDefined} from 'gmp/utils/identity';
+import {isEmpty} from 'gmp/utils/string';
 
 import {parseInt} from 'gmp/parser';
 
@@ -42,6 +43,8 @@ class Tag extends Model {
     else {
       ret.resourceCount = 0;
     }
+    ret.value = isEmpty(elem.value) ? undefined : elem.value;
+
     return ret;
   }
 }
--- a/gsa/src/web/pages/results/detailspage.js
+++ b/gsa/src/web/pages/results/detailspage.js
@@ -238,15 +238,18 @@ const Details = ({
           title={_('Tags')}
         >
           <Divider>
-            {userTags.map(tag => (
-              <DetailsLink
-                key={tag.id}
-                id={tag.id}
-                type="tag"
-              >
-                {tag.name + '=' + tag.value}
-              </DetailsLink>
-            ))}
+            {userTags.map(tag => {
+              const valueString = tag.value === '' ? '' : '=' + tag.value;
+              return (
+                <DetailsLink
+                  key={tag.id}
+                  id={tag.id}
+                  type="tag"
+                >
+                  {tag.name + valueString}
+                </DetailsLink>
+              );
+            })}
           </Divider>
         </DetailsBlock>
       }
--- a/gsa/src/web/pages/results/detailspage.js
+++ b/gsa/src/web/pages/results/detailspage.js
@@ -239,7 +239,7 @@ const Details = ({
         >
           <Divider>
             {userTags.map(tag => {
-              const valueString = tag.value === '' ? '' : '=' + tag.value;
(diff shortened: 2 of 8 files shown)


gsa/src/web/pages/scanconfigs/editdialog.js
```diff
@@ -327,31 +328,33 @@ class NvtFamily extends React.Component {
           {name}
         </TableData>
         <TableData>
-          <Layout align="end">
+          <Layout align="start">
```
bjoernricks: Couldn't the align be set on the TableData here? For me the extra Layout seems to be unnecessary.

gsa/src/web/pages/scanconfigs/editdialog.js
```diff
@@ -84,10 +85,10 @@ class NvtPreferenceDisplay extends React.Component {
         <TableData style={{overflowWrap: 'break-word'}}>
           {preference.name}
         </TableData>
-        <TableData>
+        <TableData style={{overflowWrap: 'break-word'}}>
```
bjoernricks: Can we extract this into an own component now we even use this twice?

gsa/src/web/pages/scanconfigs/editdialog.js
```diff
@@ -464,9 +467,13 @@ class NvtFamilies extends React.Component {
...
                 {_('Total: {{count}}', {count: config.families.count})}
               </TableData>
-              <TableData align="end">
+              <TableData align="start">
                 {_('{{known}} of {{max}}', config.nvts)}
               </TableData>
+              {/* add empty cells to spread row to end of table */}
+              <TableData/>
+              <TableData/>
+              <TableData/>
```
bjoernricks: Seems strange to me. Shouldn't we remove the columns at all if we don't have content for them?

gsa/src/web/pages/results/detailspage.js
```diff
@@ -238,15 +238,18 @@ const Details = ({
...
-              <DetailsLink
-                key={tag.id}
-                id={tag.id}
-                type="tag"
-              >
-                {tag.name + '=' + tag.value}
-              </DetailsLink>
-            ))}
+            {userTags.map(tag => {
+              const valueString = tag.value === '' ? '' : '=' + tag.value;
```
bjoernricks: Could you use  `isEmpty(tag.value)` from gmp/utils/string instead?

bjoernricks: Maybe it would even better to parse the tag value as undefined if it is an empty string...

gsa/src/gmp/models/tag.js
```diff
@@ -42,6 +42,9 @@ class Tag extends Model {
     else {
       ret.resourceCount = 0;
     }
+    if (elem.value === '') {
+      ret.value = undefined;
+    }
```
bjoernricks: should be
```javascript
ret.value = isEmpty(elem.value) ? undefined : elem.value;
```

gsa/src/web/pages/scanconfigs/editdialog.js
```diff
@@ -79,13 +79,13 @@ class NvtPreferenceDisplay extends React.Component {
...
-        <TableData style={{overflowWrap: 'break-word'}}>
+        <TableData overflowWrap="break-word">
           {preference.nvt.name}
         </TableData>
-        <TableData style={{overflowWrap: 'break-word'}}>
+        <TableData overflowWrap="break-word">
           {preference.name}
         </TableData>
-        <TableData style={{overflowWrap: 'break-word'}}>
+        <TableData overflowWrap="break-word">
```
bjoernricks: Oh sorry you did misunderstood my comment. Should have been
```jsx
const StyledTableData = styled(TableData)`
  overflow-wrap: break-word;
`;

...

<StyledTableData>
  {preference.nvt.name}
</StyledTableData>
<StyledTableData>
  {preference.name}
</StyledTableData>
<TableData>
...
```
Hint: If this doesn't work currently TableData has to forwards the className prop.

PR #1006, opened 2018-10-09
Require yarn 1.0

bjoernricks:
Versions beyond 1.0 will fail because the command shortcut `yarn install` isn't supported. It would have been possible to support versions lower then 1.0 by adjusting the package install command but yarn 1.0 is the first stable version and already more then one year old. Therefore supporting older versions doesn't make sense to me.


cfi-gb:
PR to include this info in the INSTALL.md as well: https://github.com/greenbone/gsa/pull/1007

PR #1016, opened 2018-10-12
Tests

swaterkamp:
Add some Tests tests

### DIFF
--- a/gsa/src/gmp/__tests__/model.js
+++ b/gsa/src/gmp/__tests__/model.js
@@ -1,0 +1,207 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+ *
+ * Copyright:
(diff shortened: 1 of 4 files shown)


gsa/src/gmp/__tests__/model.js
```diff
@@ -0,0 +1,143 @@
...
+          {name: 'everything'},
+          {name: 'may_foo'},
+        ],
+      },
+    };
+    const model = new Model(elem);
+    expect(model.userCapabilities).toBeDefined();
+    expect(model.user_capabilities).toBeUndefined();
+    expect(model.userCapabilities._capabilities)
+      .toEqual(new Set(['everything', 'may_foo']));
```
bjoernricks: This tests internal behavior of Capabilities. Therefore it should be replaced with something like
```js
 expect(model.userCapabilities.length).toEqual(2);
```

If you use real permissions like `'get_foo'` you could also test for
```js
 expect(model.userCapabilities.mayAccess('foo')).toEqual(true);
```

PR #1017, opened 2018-10-12
Dashboard improvements

bjoernricks:
Make the dashboard settings store code completely transparent. That means the store doesn't know about the actual settings data _layout_

Move all specific settings code to the dashboard components modules.

At the end this will allow to load, save and restore all kind of data from the dashboard settings via the redux store. Each layer will expect it's own data layout without interfering the other layers.


bjoernricks:
By layers I mean 
```
backend <-> gmp api <-> settings store <-> dashboard <-> main dashboard
```
where the main dashboard holds multiple dashboards in its settings.

PR #1018, opened 2018-10-12
XML escape asset_hosts filter for create_target

timopollmeier:
With this filters containing quote marks and other XML reserved
characters can be used to create a new host.


timopollmeier:
Okay, I'll look for other cases in gsad where XML escaping is missing and create a new PR if I find any.

PR #1021, opened 2018-10-12
Improve entity tags and permissions table layout

swaterkamp:
* Center Actions column
* Remove toggleDetailsIcon from permissions tab

### DIFF
--- a/gsa/src/web/entity/tags.js
+++ b/gsa/src/web/entity/tags.js
@@ -154,7 +154,7 @@ class EntityTagsTable extends React.Component {
                 <TableHead>
                   {_('Comment')}
                 </TableHead>
-                <TableHead width="5em">
+                <TableHead width="8%" align="center">
                   {_('Actions')}
                 </TableHead>
               </TableRow>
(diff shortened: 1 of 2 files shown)


gsa/src/web/entity/tags.js
```diff
@@ -179,26 +179,28 @@ class EntityTagsTable extends React.Component {
...
-                            onClick={() => onTagRemoveClick(tag.id, entity)}
-                          />
-                          <EditIcon
-                            value={tag}
-                            title={_('Edit Tag')}
-                            onClick={this.handleEditTag}
-                          />
-                        </IconDivider>
+                      <TableData align="center">
+                        <span>
```
bjoernricks: Is the span really necessary? At least it shouldn't.

bjoernricks: Did you try setting the align at the IconDivider?

swaterkamp: Yes, I did, but it didn't work. What I did not test, though, is to set the IconDivider to grow. A growing divider plus the align="center" in its tag solves the issue.

PR #1024, opened 2018-10-16
More Model Tests #2

swaterkamp:
* Alert
* Credential
* Permission

### DIFF
--- a/gsa/src/gmp/models/__tests__/alert.js
+++ b/gsa/src/gmp/models/__tests__/alert.js
@@ -1,0 +1,105 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+ *
+ * Copyright:
+ * Copyright (C) 2018 Greenbone Networks GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


gsa/src/gmp/models/__tests__/alert.js
```diff
@@ -0,0 +1,105 @@
...
+
+import Model from 'gmp/model';
+import Alert from 'gmp/models/alert';
+import {testModel} from 'gmp/models/testing';
+
+testModel(Alert, 'alert');
+
+describe('Alert Model tests', () => {
+
+  test('should parse condition, event, and method', () => {
```
bjoernricks: I would have tested them separately but that's fine.

gsa/src/gmp/models/__tests__/alert.js
```diff
@@ -0,0 +1,105 @@
...
+      },
+      type: 'foo',
+    });
+  });
+
+  test('should return given filter as instance of filter model', () => {
+    const elem = {filter: 'rows=1337'};
+    const alert = new Alert(elem);
+
+    expect(alert.filter).toEqual(new Model('rows=1337', 'filter'));
```
bjoernricks: Uh maybe it would be better to use a real Filter model here in future. Not sure if that would create a cyclic dependency...

gsa/src/gmp/models/__tests__/permission.js
```diff
@@ -0,0 +1,81 @@
...
+  test('should parse resource as model of their type', () => {
+    const elem = {
+      resource: {
+        _id: '123',
+        type: 'alert',
+      },
+    };
+    const permission = new Permission(elem);
+
+    expect(permission.resource).toEqual(new Model(elem.resource, elem.resource.type));
```
bjoernricks: Would be better to test if type and id match then checking for an equal model.

PR #1030, opened 2018-10-22
GUI miscellaneous #6

swaterkamp:


gsa/src/web/pages/alerts/component.js
```diff
@@ -304,8 +309,8 @@ class AlertComponent extends React.Component {
           condition_data_at_least_filter_id: condition_data_filter_id,
           condition_data_at_least_count: parseInt(
             value(condition.data.count, 1)),
-          condition_data_severity: value(condition.data.severity,
-            DEFAULT_SEVERITY),
+          condition_data_severity: parseFloat(value(condition.data.severity,
```
bjoernricks: would be better to user parseSeverity instead of parseFloat

PR #1031, opened 2018-10-23
More model test #3

swaterkamp:
Tests for: 
* CertBundAdv
* CVE

### DIFF
--- a/gsa/src/gmp/models/__tests__/certbund.js
+++ b/gsa/src/gmp/models/__tests__/certbund.js
@@ -1,0 +1,278 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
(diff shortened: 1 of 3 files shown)


gsa/src/gmp/models/__tests__/certbund.js
```diff
@@ -0,0 +1,278 @@
...
+  test('should return string if single CategoryTree is given', () => {
+    const elem = {
+      raw_data: {
+        Advisory: {
+          CategoryTree: 'foo',
+        },
+      },
+    };
+    const certBundAdv = new CertBundAdv(elem);
+    expect(certBundAdv.categories).toEqual('foo');
```
bjoernricks: I guess you have spotted bug in CertBundAdv model. This should be
```js
['foo']
```
Could you fix the model in an additional commit/pr?

PR #1033, opened 2018-10-23
Remove maxlength of hosts for notes and overrides

timopollmeier:
This removes the 80 character limit of the text field for entering the
hosts of notes and overrides manually.


bjoernricks:
@timopollmeier is there still any restriction in gvmd/the database for these length?

timopollmeier:
@bjoernricks: There are restrictions to the number of hosts, but not the length of the string.

PR #1040, opened 2018-11-01
More model tests #7

swaterkamp:
Includes a polyfill fix for settings model and tests for:
* Scanner
* ScanConfig
* Credential
* Schedule
* SecInfo
* Settings
* Tag
* Target
* User
* Vulnerability

### DIFF
--- a/gsa/src/gmp/models/__tests__/settings.js
+++ b/gsa/src/gmp/models/__tests__/settings.js
@@ -1,0 +1,46 @@
+/* Greenbone Security Assistant
+ *
+ * Authors:
+ * Steffen Waterkamp <steffen.waterkamp@greenbone.net>
+ *
+ * Copyright:
+ * Copyright (C) 2018 Greenbone Networks GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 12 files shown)


gsa/src/gmp/models/__tests__/settings.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import Settings from 'gmp/models/settings';
+
+describe('Settings model tests', () => {
+
+  test('should have private settings', () => {
+    const settings = new Settings({});
+
+    expect(settings._settings).toBeDefined();
```
bjoernricks: Please don't test internal API. Internal API is marked with the underscore. The tests should only ensure the public api is working correctly and the internal api may change every time.

bjoernricks: This concerns all tests in this module.

swaterkamp: Doing this, I could only test set() and get() within a single test, but it should work. With the last push, I also added more commits, so just that you know :)

bjoernricks: > I could only test set() and get() within a single test

That's completely fine.

PR #1059, opened 2018-11-20
Credential additions

bjoernricks:
Allow to create smime and pgp credentials. It's now possible to create an email alert using these credentials yet.

@timopollmeier could you test smime and modify pgp? I am not having a smime cert and saving an existing pgp credential without changing anything in the dialog currently fails at manager side with
`double free or corruption (fasttop)`


timopollmeier:
Creating the credentials works fine but there are some issues with modifying the credential on the gvmd side I'll have to fix.

PR #1060, opened 2018-11-20
Fix QoD value for qodColorScale

swaterkamp:
We (GSF-dev and GVM-dev) should stick to the 11 QoD-values defined in the manual (https://docs.greenbone.net/GSM-Manual/gos-4/en/glossary.html#quality-of-detection-qod ).

This PR fixes the 76% to 75% value and GSF-dev will update the VTs that introduced 76% and 60% values.


bjoernricks:
Instead of using odinal scale we could also use a [linear scale](https://github.com/d3/d3-scale#continuous_domain)

PR #1069, opened 2018-11-23
Use SvgIcon for EditIcon and Trend

swaterkamp:
With this PR the ScanConfig-dialog should be usable again. All other icons should be changed in the future.


gsa/src/web/components/icon/trendmoreicon.js
```diff
@@ -0,0 +1,40 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import React from 'react';
+
+import SvgIcon from './svgicon';
+
+import {ReactComponent as Icon} from './svg/trend_more.svg';
+
+const TrendNoChangeIcon = props => (
```
bjoernricks:
```suggestion
const TrendMoreIcon = props => (
```

gsa/src/web/components/icon/trendmoreicon.js
```diff
@@ -0,0 +1,40 @@
...
+  <SvgIcon
+    {...props}
+  >
+    <Icon/>
+  </SvgIcon>
+);
+
+export default TrendNoChangeIcon;
```
bjoernricks:
```suggestion
export default TrendMoreIcon;
```

Issue #916, opened 2018-09-05
New encryption field for email alerts


timopollmeier:
There should be an option in the "New Alert" and "Edit Alert" dialogs to specify a GPG public key or S/MIME certificate for encrypting the emails sent by alerts, which I'd imagine to look something like this:

(  ) No encryption
(  ) Encrypt for public key or certificate [Upload File]

The preference would be named "email_encryption_public_key" and with the "No encryption" option it should send an empty string. For the second option should either send the selected file. If no file is selected, it should fail on the creation dialog and not send anything to keep the existing value on the edit dialog.

I'll update this issue once I've added the preference to gvmd.

timopollmeier:
Because the alert method data has to be resent every time an alert is modified, it makes more sense
to store the public key / certificate in a credential.

timopollmeier:
The PRs for the new gvmd/GMP functionality have been merged, so it should be possible to implement the GSA interface for the email encryption now.

PR #1070, opened 2018-11-23
Alert encrypt email

bjoernricks:
Allow to encrypt emails send by alerts via setting a pgp or smime credential.


bjoernricks:
Closes #916

PR #1072, opened 2018-11-27
Use ImportIcon on ReportFormat listpage

swaterkamp:


gsa/src/web/components/icon/importicon.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const ImportIcon = ({active = true, ...props}) => {
+  return (
+    <SvgIcon
+      {...props}
+    >
+      <Icon/>
+    </SvgIcon>
+  );
+};
```
bjoernricks: Can be simplified
```js
export const ImportIcon = props => (
  <SvgIcon
    {...props}
  >
    <Icon/>
  </SvgIcon>
);

bjoernricks: Could still be simplified but that's ok.

swaterkamp: Arg! That 'return' is just invisible in my brain... I'm going to make another commit for this in a future Misc-PR.

PR #1073, opened 2018-11-27
Add Scan Report Content Composer

swaterkamp:
This PR introduces the content composer for triggering alerts for and downloading of reports.


gsa/src/gmp/commands/users.js
```diff
@@ -267,6 +270,38 @@ class UserCommand extends EntityCommand {
...
+      const {setting = {}} = data.get_settings.get_settings_response;
+      const {value} = setting;
+      let defaults;
+
+      try {
+        defaults = JSON.parse(value);
+      }
+      catch (e) {
+        log.warn('Could not parse report composer defaults');
+        return;
```
bjoernricks: Maybe return sane defaults for the reportcomposerdefaults here?

gsa/src/web/components/dialog/composercontent.js
```diff
@@ -0,0 +1,121 @@
...
+`;
+
+const ComposerContent = ({
+  applyOverrides,
+  filterString,
+  includeNotes,
+  includeOverrides,
+  onValueChange,
+}) => {
+  return (
```
bjoernricks: Syntax could be shortened
```javascript
}) => (
```

gsa/src/web/components/dialog/composercontent.js
```diff
@@ -0,0 +1,121 @@
...
+      </FormGroup>
+    </Layout>
+  );
+};
+
+ComposerContent.propTypes = {
+  applyOverrides: PropTypes.numberOrNumberString,
+  filterString: PropTypes.string,
+  includeNotes: PropTypes.number,
+  includeOverrides: PropTypes.number,
```
bjoernricks: I suppose the values should be required. Isn't it an error if undefined values are passed?

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -51,27 +57,32 @@ class AlertActions extends React.Component {
...
     this.onAlertCreated = this.onAlertCreated.bind(this);
+    this.handleOpenTriggerAlertDialog = this.handleOpenTriggerAlertDialog.bind(this); /* eslint-disable-line max-len */
+    this.handleCloseTriggerAlertDialog = this.handleCloseTriggerAlertDialog.bind(this); /* eslint-disable-line max-len */
+    this.handleValueChange = this.handleValueChange.bind(this);
   }
 
   componentDidMount() {
     const {gmp} = this.props;
+    this.props.loadReportComposerDefaults();
     gmp.alerts.getAll().then(response => {
```
bjoernricks: Because the alerts are only necessary when opening the composer now, maybe be should load them only on demand when opening the composer dialog. Also it would be better to load them from the store.

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -81,49 +92,95 @@ class AlertActions extends React.Component {
...
+    const {
+      filter,
       gmp,
       report,
-      filter,
       showErrorMessage,
       showSuccessMessage,
     } = this.props;
 
+    filter.set('notes', includeNotes);
```
bjoernricks: Because set mutates the filter it would be better to create a copy before updating it.
```suggestion
    const newfilter = filter.copy()
    newfilter.set('notes', includeNotes);
```

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -81,49 +92,95 @@ class AlertActions extends React.Component {
...
+
     this.handleInteraction();
 
+    if (storeAsDefault) {
+      const defaults = {
+        applyOverrides,
+        includeNotes,
+        includeOverrides,
+      };
+      this.props.saveReportComposerDefaults(defaults);
```
bjoernricks: IMHO not assigning the new object to a variable looks better
```javascript
this.props.saveReportComposerDefaults({
  applyOverrides,
  includeNotes,
  includeOverrides,
});
```

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -81,49 +92,95 @@ class AlertActions extends React.Component {
...
-      alerts: [alert, ...alerts],
-      alert_id: alert.id,
+      showTriggerAlertDialog: false,
     });
   }
 
+  onAlertCreated(response) {
+    const {alerts} = this.state;
+    const {gmp} = this.props;
+    gmp.alert.get({id: response.data.id}).then(resp => {
```
bjoernricks: Would also be better to load the alert from the store too but that's a bit more difficult.

gsa/src/web/pages/reports/detailspage.js
```diff
@@ -554,6 +590,20 @@ class ReportDetails extends React.Component {
...
+            applyOverrides={applyOverrides}
+            filter={filter}
+            includeNotes={includeNotes}
+            includeOverrides={includeOverrides}
+            reportFormatId={reportFormatId}
+            reportFormats={reportFormats}
+            storeAsDefault={storeAsDefault}
+            onClose={this.handleCloseDownloadReportDialog}
+            onSave={this.handleReportDownload}
+            onValueChange={this.handleValueChange}
```
bjoernricks: I don't think the dialog should update the values in the state of the detailspage directly. This will cause unnecessary rerendering of the report details page. Instead use uncontrolled values in the dialog only update the state at the details page on save.

bjoernricks: It seems you are already doing this because you aren't using onValueChange in DownloadReportDialog. Please remove this line and the handleValueChange method then.

gsa/src/gmp/commands/users.js
swaterkamp: That would silently swallow the error disregarding that something is wrong in the database. In case the settings can't be loaded, sane defaults are set in the props for DownloadReportDialog and TriggerAlertDialog. If any of the defaults is undefined, they'll get set there.

bjoernricks: I think your code will crash currently in this case because you are expecting a response object in your action creator (access to response.data) and will get only undefined. Instead of returning sane defaults via return response.setData(...) you could also throw an exception and therefore cause the promise catch path to create an error action.

gsa/src/gmp/commands/users.js
```diff
@@ -45,6 +45,8 @@ import {parseInt} from '../parser';
 import EntitiesCommand from './entities';
 import EntityCommand from './entity';
 
+import {COMPOSER_CONTENT_DEFAULTS} from 'web/components/dialog/composercontent';
```
bjoernricks: Sorry the gmp part must not depend on the web part. You could define the defaults here and import them in the web part or you only return an empty object and set the defaults. The only aim is to ensure not to crash if an error did occur.

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -152,16 +150,15 @@ class AlertActions extends React.Component {
...
   onAlertCreated(response) {
-    const {alerts} = this.state;
-    const {gmp} = this.props;
-    gmp.alert.get({id: response.data.id}).then(resp => {
-      const alert = resp.data;
-      this.setState({
-        alerts: [alert, ...alerts],
-        alertId: alert.id,
+    this.props.loadAlerts().then(() => {
+      this.props.loadAlert(response.data.id).then(resp => {
```
bjoernricks: Not sure if both can't be loaded in parallel instead.

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -152,16 +150,15 @@ class AlertActions extends React.Component {
   }
 
   onAlertCreated(response) {
```
bjoernricks: Do you need the full information of response? I suppose only passing the alertId would be fine here.

gsa/src/web/pages/reports/alertactions.js
```diff
@@ -152,16 +150,15 @@ class AlertActions extends React.Component {
...
-    gmp.alert.get({id: response.data.id}).then(resp => {
-      const alert = resp.data;
-      this.setState({
-        alerts: [alert, ...alerts],
-        alertId: alert.id,
+    this.props.loadAlerts().then(() => {
+      this.props.loadAlert(response.data.id).then(resp => {
+        const alert = resp.data;
+        this.setState({
+          alertId: alert.id,
```
bjoernricks: Why not use the passed alert id here (currently from response object of onAlertCreated)?

bjoernricks:
Looks good :+1:

PR #1076, opened 2018-12-03
Allow rename main dashboards

bjoernricks:


gsa/src/web/pages/start/page.js
```diff
@@ -141,6 +143,10 @@ class StartPage extends React.Component {
     this.handleResetDashboards = this.handleResetDashboards.bind(this);
     this.handleSetDefaultSettings = this.handleSetDefaultSettings.bind(this);
 
+    this.handleCloseEditDashboardDialog =
+      this.handleCloseEditDashboardDialog.bind(this);
+    this.handleSaveEditDashboard = this.handleSaveEditDashboard.bind(this);
+
```
swaterkamp: This whole file could use a proper renaming of methods. Some use handle, some don't (e.g. handleConfirmRemoveDialogClose() and closeNewDashboardDialog() ) and also word order changes across methods (e.g. handleConfirmRemoveDialogClose() and handleOpenNewDashboardDialog() ). This isn't necessarily part of this PR and just about coding style but as the file gets touched anyway, it might be a good point to unify the naming.

bjoernricks: You are right completely. The naming in the module is messed up and I'll fix it in an additional commit.

PR #1089, opened 2018-12-12
Set min_qod=0 for "removed" filter in PowerFilter

swaterkamp:
Removing a filter should result in all results being shown. min_qod>0
reduces the number of results and should have been changed when
introducing the "remove filter" functionality already.


swaterkamp:
I was made aware that changing the min_qod might cause some trouble if there are too many results resulting from this, therefore I'm closing this PR.

PR #1108, opened 2018-12-21
Move named filter creation from powerfilter to filterdialog

swaterkamp:


swaterkamp:
The flow of props is really messy. I'd like to postpone changing it, though, because it's a bigger story that might break a lot right before the feature freeze for GOS 5.

gsa/src/web/pages/reports/filterdialog.js
```diff
@@ -122,12 +131,24 @@ const ReportFilterDialogComponent = ({
         onSortByChange={onSortByChange}
       />
 
+      {capabilities.mayCreate('filter') &&
+        <CreateNamedFilterGroup
+          filter={filter}
+          filterName={filterName}
+          filterNameValid={filterNameValid}
```
bjoernricks: I am really unsure about the filter name valid code because

* The displayed message is about missing name and the prop is about a valid name
* The filter group could check for the filter name by itself
* Saving could be disable if the input doesn't contain a name (as it was before with the powerfilter)
* Saving without a name could issue an error which is displayed in a more general error handling part like with the SaveDialog
* filterNameValid is set to true as default which isn't really the case in the initial rendering where the name is missing. So maybe using filterNameInvalid would be better

gsa/src/web/components/powerfilter/createnamedfiltergroup.js
```diff
@@ -0,0 +1,90 @@
...
+`;
+
+const CreateNamedFilterGroup = ({
+  filter,
+  filterName,
+  filterNameValid = true,
+  saveNamedFilter,
+  onValueChange,
+}) => {
+  return (
```
bjoernricks: Function component could be simplified by removing the return statement.

gsa/src/web/components/powerfilter/createnamedfiltergroup.js
```diff
@@ -0,0 +1,90 @@
...
+      </Divider>
+    </StyledLayout>
+  );
+};
+
+CreateNamedFilterGroup.propTypes = {
+  filter: PropTypes.filter.isRequired,
+  filterName: PropTypes.string,
+  filterNameValid: PropTypes.bool,
+  saveNamedFilter: PropTypes.bool,
```
bjoernricks: saveNamedFilter should be required or set to a default.

gsa/src/web/components/powerfilter/createnamedfiltergroup.js
```diff
@@ -0,0 +1,90 @@
...
+            {_('Please insert a name for the new filter')}
+          </Div>
+        }
+      </Divider>
+    </StyledLayout>
+  );
+};
+
+CreateNamedFilterGroup.propTypes = {
+  filter: PropTypes.filter.isRequired,
```
bjoernricks: Is the filter available on the initial rendering?

PR #1111, opened 2018-12-23
Update files with new standard header.

janowagner:
This covers all file types except ".js" which remains to be done.


cmake/FindNpm.cmake
```diff
@@ -22,6 +16,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
 
+# CMake find module for yarn package manager
```
wiegandm:
```suggestion
# CMake find module for node package manager
```

PR #1117, opened 2019-01-12
Add Content Composer to New/Edit Alerts

swaterkamp:
Needs #1115 and https://github.com/greenbone/gvmd/pull/319


gsa/src/web/pages/alerts/component.js
```diff
@@ -891,6 +1005,20 @@ class AlertComponent extends React.Component {
                 onSave={this.handleCreateCredential}
               />
             }
+            {contentComposerDialogVisible &&
+              <ContentComposerDialog
+                includeNotes={parseInt(includeNotes)}
+                includeOverrides={parseInt(includeOverrides)}
```
bjoernricks: both values are boolean and contained in the state. So why not parsing them (parseYesNo I guess) when changing them in the state.

gsa/src/web/pages/alerts/component.js
```diff
@@ -213,6 +220,57 @@ class AlertComponent extends React.Component {
...
+      includeOverrides,
+      filterId,
+      filterString,
+    } = this.state;
+    this.setState({
+      filterId,
+      filter_id: filterId,
+      filter_string: filterString,
+      method_data_composer_include_notes: includeNotes,
+      method_data_composer_include_overrides: includeOverrides,
```
bjoernricks: These four lines seem to be strange to me. Why putting something in the state from the state?

bjoernricks: Same for **closeContentComposerDialog** method

gsa/src/web/pages/alerts/component.js
```diff
@@ -341,6 +399,9 @@ class AlertComponent extends React.Component {
           comment: alert.comment,
           filters,
           filter_id: isDefined(alert.filter) ? alert.filter.id : NO_VALUE,
+          filterId: isDefined(alert.filter) ? alert.filter.id : NO_VALUE,
```
bjoernricks: The duplication seems to be weird to me.

gsa/src/web/pages/alerts/component.js
swaterkamp: The values are not always the same. This duplication was needed to fulfill user expectations about closing the composer without saving and so on.

gsa/src/web/pages/alerts/component.js
swaterkamp: The handler to change those values also handles non-boolean changes (filterId, filterString). Code changed to parse as YesNo instead of Int

swaterkamp: btw: parseYesNo does not return boolean but numbers.

bjoernricks: That's not completely correct. parseYesNo returns the boolean values for the protocol which are 1 and 0 as integers. We may decide to change them to true and false in GSA and in that case still using parseInt here would break it.

PR #1120, opened 2019-01-15
Migrate license headers to new template and update year of last modification

davidak:
In 2 places the script has also updated the copyright line. (footer and about)

Please review if that's in your interest.

### DIFF
--- a/gsa/src/gmp/__tests__/cancel.js
+++ b/gsa/src/gmp/__tests__/cancel.js
@@ -1,4 +1,4 @@
-/* Copyright (C) 2017-2018 Greenbone Networks GmbH
+/* Copyright (C) 2017-2019 Greenbone Networks GmbH
  *
  * SPDX-License-Identifier: GPL-2.0-or-later
  *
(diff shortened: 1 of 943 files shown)


davidak:
> Shouldn't the new headers include 2019?

This PR only migrates the license headers to the new template. Updating the date of last modification is another task i work on.

swaterkamp:
I just thought it could be done in one PR, as you are touching more than 940 files anyway. But go ahead :)

davidak:
@swaterkamp if it's easier for you to review i can add it to this PR.

swaterkamp:
That way all headers have to be "reviewed" only once instead of twice (once for the new template, and once whether the year got updated).

davidak:
OK, i will do so. The year is updated after the last change from git. Not every file was updated in 2019.

bjoernricks:
We tend to update all headers to 2019 because the copyright for the new year is still valid. It doesn't end in the last year when being touched.

davidak:
I also updated the year. Please note the new PR description!

swaterkamp:
The changes in About and Footer are definitely in our interest. Thanks!

PR #1130, opened 2019-01-21
Add CHANGES for gsa 7.0.4

timopollmeier:
This adds a list of changes made since GSA 7.0.3 to the CHANGES file.


cfi-gb:
While you're "at it" there are a few additional typos in older entries which could be fixed in one rush:

CHANGES:81: manger  ==> manager
CHANGES:684: explicitely  ==> explicitly
CHANGES:894: explicitely  ==> explicitly
CHANGES:1115: overrrides  ==> overrides
CHANGES:1182: eigth  ==> eighth, eight

PR #1152, opened 2019-01-30
Icons as component

bjoernricks:
Import our icons as react components whenever possible.


gsa/src/web/components/icon/downloadkeyicon.js
```diff
@@ -0,0 +1,36 @@
...
+
+import React from 'react';
+
+import SvgIcon from './svgicon';
+
+import {ReactComponent as Icon} from './svg/dl_key.svg';
+
+const DownloadKeyIcon = props => (
+  <SvgIcon
+    {...props}
```
swaterkamp: The other icons have titles. Is this on purpose that the DownloadKeyIcon doesn't? Is it passed via props?

PR #1162, opened 2019-02-05
Update gsad.c

cheese1:
Buffer not null-terminated.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsad/src/gsad.c
+++ b/gsad/src/gsad.c
@@ -2586,7 +2586,7 @@ start_unix_http_daemon (const char *unix_socket_path,
       return NULL;
     }
   addr.sun_family = AF_UNIX;
-  strncpy (addr.sun_path, unix_socket_path, sizeof (addr.sun_path));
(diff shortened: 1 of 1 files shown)


gsad/src/gsad.c
```diff
@@ -2586,7 +2586,7 @@ start_unix_http_daemon (const char *unix_socket_path,
       return NULL;
     }
   addr.sun_family = AF_UNIX;
-  strncpy (addr.sun_path, unix_socket_path, sizeof (addr.sun_path));
+  strncpy (addr.sun_path, unix_socket_path, sizeof (addr.sun_path) -1);
```
mattmundell:
```suggestion
  strncpy (addr.sun_path, unix_socket_path, sizeof (addr.sun_path) - 1);
```

timopollmeier:
I think the idea of this is to ensure the final character is the NULL byte even if the given socket path is as long as the buffer size or longer.

PR #1174, opened 2019-02-08
Doc https

wiegandm:
Use HTTPS instead of HTTP as protocol for accessing the documentation
available from `docs.greenbone.net`.

**Checklist**:

- [ ] Tests N/A


gsa/src/gmp/gmpsettings.js
```diff
@@ -19,9 +19,9 @@
 import {isDefined} from './utils/identity';
 
 export const DEFAULT_RELOAD_INTERVAL = 15 * 1000; // fifteen seconds
-export const DEFAULT_MANUAL_URL = 'http://docs.greenbone.net/GSM-Manual/gos-4/';
+export const DEFAULT_MANUAL_URL = 'https://docs.greenbone.net/GSM-Manual/gos-4/';
 export const DEFAULT_PROTOCOLDOC_URL =
-  'http://docs.greenbone.net/API/GMP/gmp-7.0.html';
+  'https://docs.greenbone.net/API/GMP/gmp-7.0.html';
```
bjoernricks: Should be updated to gmp 8 too. gsad will communicate with version 8.

bjoernricks: Oh master is will even use version 9...

PR #1176, opened 2019-02-11
Report rendering improvements

bjoernricks:
Improve rendering of big reports

**Checklist**:

- [n/a] Tests
- [ ] [CHANGES](https://github.com/greenbone/gsa/blob/master/CHANGES.md) Entry


bjoernricks:
IMHO that's it from the gsa(d) side. Rendering time is down to ~10 seconds now.

PR #1177, opened 2019-02-12
Report details update indicators

bjoernricks:
Show transparent tables at the result details page tabs if a user has causes a new loading procedure by changing the current filter. Currently there is no indicator that we are actually updating the data in the background if the filter has changed and we have a lot of data. In that case the user might think that nothing has changed and updating the filter doesn't work.


bjoernricks:
Wrong branch :cry:

PR #1186, opened 2019-02-13
Get reports without result tags, to save time

mattmundell:
This is to speed up the Report Details page, by requesting less info from Manager.

**Checklist**:

- [ ] Test N/A
- [ ] [CHANGES](https://github.com/greenbone/gsa/blob/master/CHANGES.md) Entry N/A


mattmundell:
Requires https://github.com/greenbone/gvmd/pull/393 to have any effect.

PR #1189, opened 2019-02-13
Remove considering unused reports in get_tasks.

janowagner:
Neither first_report nor second_last_report is used.
Those elements will not be present anymore in the GMP
response of get_tasks.

**Checklist**:

- [ ] Tests N/A
- [ ] [CHANGES](https://github.com/greenbone/gsa/blob/master/CHANGES.md) Entry N/A


bjoernricks:
@janowagner did you add the right commit to the pr? I guess you wanted to change gsad_gmp.c. This PR only contains changes created by auto-formatting via prettier.

janowagner:
The change is in the diff, see the reduced list for  "const reports =".

All other changes were automatically applied.

PR #1207, opened 2019-03-08
Menu permissions

swaterkamp:
This fixes wrongly displayed menues. We need to make sure that the first child of a Menu is a MenuEntry and not a MenuSection.

**Checklist**:

- [ N/A ] Tests
- [ N/A ] [CHANGES](https://github.com/greenbone/gsa/blob/master/CHANGES.md) Entry

### DIFF
--- a/gsa/src/web/components/menu/menu.js
+++ b/gsa/src/web/components/menu/menu.js
@@ -28,6 +28,8 @@ import Theme from 'web/utils/theme.js';
 
 import Link from 'web/components/link/link.js';
 
+import MenuSection from 'web/components/menu/menusection';
+
 const StyledMenu = styled.li`
   flex-grow: 1;
   flex-shrink: 1;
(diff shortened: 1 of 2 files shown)


gsa/src/web/components/menu/menu.js
```diff
@@ -114,13 +116,24 @@ const MenuList = styled.ul`
   }
 `;
 
+const checkChild = child => {
+  if (child.type === MenuSection && isArray(child.props.children)) {
```
bjoernricks: I am not 100 percent sure if this will work. React has a special API to access children because they aren't arrays https://reactjs.org/docs/react-api.html#reactchildren

gsa/src/web/components/menu/menu.js
```diff
@@ -114,13 +116,24 @@ const MenuList = styled.ul`
...
+    return child.props.children.find(chil => chil !== false);
```
bjoernricks: I am really sure this won't work always. Please use the React.Children API (https://reactjs.org/docs/react-api.html#reactchildren) and I guess null and undefined should be checked too. Maybe use
```javascript
chil => !!chil
```

gsa/src/web/components/menu/menu.js
```diff
@@ -114,13 +116,24 @@ const MenuList = styled.ul`
...
 const Menu = ({children, title, to, ...props}) => {
   let link;
   children = React.Children.toArray(children).filter(hasValue);
+
   if (isDefined(to)) {
     link = <Link to={to}>{title}</Link>;
   } else if (isDefined(children) && children.length > 0) {
-    const [child] = children;
+    let [child] = children;
+    child = checkChild(child);
```
bjoernricks: IMHO the function name checkChild is a bit misplaced. I do expect either to return a boolean or to raise an exception for a check named function. I guess
```javascript
getFirstMenuEntry(children)
```
would fit?

PR #1224, opened 2019-03-22
Add notification if user is to be created without a role

swaterkamp:
**Checklist**:

- [ X ] Tests
- [ X ] [CHANGES](https://github.com/greenbone/gsa/blob/master/CHANGES.md) Entry

### DIFF
--- a/gsa/src/web/components/dialog/confirmationdialog.js
+++ b/gsa/src/web/components/dialog/confirmationdialog.js
@@ -75,7 +75,7 @@ ConfirmationDialogContent.propTypes = {
   onResumeClick: PropTypes.func.isRequired,
 };
 
-const LinkConfirmationDialog = ({
+const ConfirmationDialog = ({
   width = DEFAULT_DIALOG_WIDTH,
   text,
   title,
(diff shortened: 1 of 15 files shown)


bjoernricks:
Looks good! Could you add a changelog entry and at least a test for the ConfirmationDialog?

gsa/src/web/components/dialog/confirmationdialog.js
```diff
@@ -47,18 +47,21 @@ class ConfirmationDialogContent extends React.Component {
   }
 
   render() {
-    const {moveprops, text, title} = this.props;
+    const {moveprops, text, title, rightButtonTitle} = this.props;
 
     return (
       <DialogContent>
         <DialogTitle
+          data-testid="confirmationdialog-title"
```
bjoernricks: Should be removed because it isn't used anymore.
```suggestion
```

gsa/src/web/components/dialog/title.js
```diff
@@ -46,12 +46,17 @@ const DialogTitleBar = styled(Layout)`
...
     <DialogTitleBar
+      data-testid="dialog-title-bar"
       align={['space-between', 'center']}
       onMouseDown={onMouseDown}
     >
       <span>{title}</span>
       {showClose && (
-        <DialogCloseButton title={_('Close')} onClick={onCloseClick} />
+        <DialogCloseButton
+          data-testid="dialog-close-button"
```
bjoernricks: Must be renamed because we already have a dialog-close-button at the twobuttonfooter.js file.
Maybe
```suggestion
          data-testid="dialog-title-close-button"
```

PR #1228, opened 2019-03-27
Avoid undefined behavior in filters with first and rows keywords

bjoernricks:
Filter keywords first with a value less then 1 are not allowed. Therefore all values less then 1 are converted to 1 automatically.

Also for rows and first must always use the equal = relation. Other relations like >, < are not defined.

**Checklist**:

- [x] Tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,8 @@ $ cd gsa && git log
  * Don't create a container task from the task dialog accidentially #1220
  * Use "Do not automatically delete reports" as default again in task dialog
    #1220
+ * Convert first filter keyword values less then one to one #1228
+ * Always use equal relation for first and rows filter keywords #1228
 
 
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -39,6 +39,8 @@ $ cd gsa && git log
  * Don't create a container task from the task dialog accidentially #1220
  * Use "Do not automatically delete reports" as default again in task dialog
    #1220
+ * Convert first filter keyword values less then one to one #1228
+ * Always use equal relation for first and rows filter keywords #1228
```
swaterkamp: Could you add the newer entries at the top, please? It also just occurred to me that the sorting is messed up and I didn't change that in my last PR, sorry. If you find the time, could you also sort the entries? If not, I will do it later on.

PR #1232, opened 2019-03-28
Remove character constraints

bjoernricks:
Remove character/text input limits for nearly all fields in GSA.

**Checklist**:

- [N/A ] Tests


bjoernricks:
@mattmundell I've removed nearly all character length constraints from gsa(d).

PR #1246, opened 2019-04-02
Guest user

bjoernricks:
Support for guest user login. With this change a config.js file with the following content

```javascript
config = {
    guestUsername: 'foo',
    guestPassword: 'bar',
}
```
can be provided to activate a `Login as Guest` button at the login page. To actually use this user
a user with the role Guest and the specified username/password combination must be created.

**Important**: It is not checked if the user is a restricted guest user. It would also be possible to auto login via a super admin user!

Additionally by adding a `disableLoginForm: true` to the config object in the `config.js` file it is possible to deactivate the *normal* login form at the login page to only provide the `Login as Guest` button.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/login/loginform.js
+++ b/gsa/src/web/pages/login/loginform.js
@@ -113,7 +113,7 @@ class LoginForm extends React.Component {
       <React.Fragment>
         <Wrapper>
           {showProtocolInsecure && (
-            <Panel>
+            <Panel data-testid="protocol-insecure">
               <Error>{_('Warning: Connection unencrypted')}</Error>
               <p>
                 {_(
@@ -156,7 +156,11 @@ class LoginForm extends React.Component {
                   />
                 </FormGroup>
                 <FormGroup size="4" offset="4">
-                  <Button title={_('Login')} onClick={this.handleSubmit} />
+                  <Button
+                    data-testid="login-button"
+                    title={_('Login')}
+                    onClick={this.handleSubmit}
+                  />
                 </FormGroup>
               </Layout>
             )}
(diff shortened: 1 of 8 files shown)


gsa/src/web/pages/login/loginform.js
```diff
@@ -93,62 +101,85 @@ class LoginForm extends React.Component {
   }
 
   render() {
-    const {error} = this.props;
+    const {
+      error,
+      showGuestLogin = false,
+      showLogin = false,
```
swaterkamp: If this is false by default, shouldn't NOT setting it to true explicitly on the LoginPage not show the LoginForm? I.e. when rendering ```<LoginForm/>``` we would need the prop ```showLogin={true}```

swaterkamp: Nevermind. It's done in the next commit...

gsa/src/web/utils/testing.js
```diff
@@ -52,6 +52,14 @@ afterEach(cleanup);
 const queryAllByName = (container, name) =>
   queryAllByAttribute('name', container, name);
 
+const queryByName = (container, name) => {
+  const els = queryAllByName(container, name);
```
swaterkamp: What does els stand for? Without the proper background knowledge this is not a descriptive variable name for me

bjoernricks: It uses the same name pattern as dom-testing-library:

queryAll... => return all found elements as array or empty array if no element is found
query... => return first found element or null if no element is found
get... => return first found elment or raise exception if no element is found
...ByTestId => search by data-test-id attribute
...ByName => search by name attribute

gsa/src/web/pages/login/loginform.js
bjoernricks: It's updated to showLogin=true in a later commit ;-)

gsa/src/web/utils/testing.js
swaterkamp: Thanks for the comprehensive answer. My question was merely about the variable naming though ;)

bjoernricks: `els` is an abbreviation for elements. I've used the same variables as in dom-testing-library
https://github.com/kentcdodds/dom-testing-library/blob/master/src/queries.js#L225

PR #1297, opened 2019-04-22
Scanner prefs

kroosec:
- use_mac_addr was removed from scanner https://github.com/greenbone/openvas-scanner/pull/181
- There are two 'ping_hosts' checks, remove one.

**Checklist**:

- [] Tests N/A
- [] [CHANGELOG](https://github.com/greenbone/gsa/blob/master/CHANGELOG.md) Entry N/A


bjoernricks:
@kroosec this code change is only valid for master and not GVM 10 right?

kroosec:
> @kroosec this code change is only valid for master and not GVM 10 right?

Valid for GVM 10 too, as they were removed 8 months ago. Added a PR for GSA-8 https://github.com/greenbone/gsa/pull/1298

bjoernricks:
Ok in that case I really only need one PR for the gsa-8.0 branch.

PR #1307, opened 2019-04-24
Send filt_id in get_report

timopollmeier:
This will add the filt_id to the GMP command sent to gvmd, falling back
to FILT_ID_USER_SETTING (-2) if it's missing.

**Checklist**:
- [ ] Tests


timopollmeier:
@bjoernricks: Feel free to just close this PR then. I was under the impression that you and @swaterkamp were expecting me to check if this was a gvmd issue first. Instead I found the missing filt_id in gsad get_report().
About FILT_ID_USER_SETTING: It's the common fallback in gsad_gmp.c:get_many() but in the end you should use whichever value fits the way you use get_report best.

PR #1385, opened 2019-05-10
Folding with styled components

bjoernricks:
Convert folding to use styled-component completely.

**Checklist**:

- [n/a] Tests
- [n/a] [CHANGELOG](https://github.com/greenbone/gsa/blob/master/CHANGELOG.md) Entry


bjoernricks:
Withdrawn because it is a bigger change which should not go into gsa-8.0.

PR #1412, opened 2019-05-16
Fix some problems with German translation in dialogs

sarahd93:
**Checklist**:

- [N/A] Tests


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Cleanup get_report function in gsad [#1263](https://github.com/greenbone/gsa/pull/1263)
 
 ### Fixed
+- Fix problems with German translation in Add Dashboard dialog and Results Filter dialog [#1412](https://github.com/greenbone/gsa/pull/1412)
```
swaterkamp: Could you change this to the SolutionTypeGroup instead of the two dialogs? Because that is the place where you actually fixed something, independent of where the group is used.

PR #1423, opened 2019-05-24
Fix result undefined error on result details

sarahd93:
**Checklist**:

- [N/A] Tests

### DIFF
--- a/gsa/src/web/pages/results/details.js
+++ b/gsa/src/web/pages/results/details.js
@@ -71,7 +71,6 @@ const ResultDetails = ({className, links = true, entity}) => {
     ? result.detection.result.details
     : undefined;
 
-  const delta = isDefined(result.delta) ? result.delta : undefined;
   const result2 = isDefined(delta) ? result.delta.result : undefined;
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/results/details.js
```diff
@@ -71,7 +71,8 @@ const ResultDetails = ({className, links = true, entity}) => {
     ? result.detection.result.details
     : undefined;
 
-  const result2 = result.delta.result;
+  const delta = isDefined(result.delta) ? result.delta : undefined;
```
swaterkamp:
```suggestion
```

gsa/src/web/pages/results/details.js
```diff
@@ -71,7 +71,8 @@ const ResultDetails = ({className, links = true, entity}) => {
...
+  const result2 = isDefined(delta) ? result.delta.result : undefined;
```
swaterkamp:
```suggestion
  const result2 = isDefined(result.delta) ? result.delta.result : undefined;
```

PR #1424, opened 2019-05-24
Consolidate new refs from NVTs

swaterkamp:
**Checklist**:

- [ X ] Tests

### DIFF
--- a/gsa/src/web/components/link/certlink.js
+++ b/gsa/src/web/components/link/certlink.js
@@ -25,7 +25,12 @@ import PropTypes from 'web/utils/proptypes';
 import DetailsLink from './detailslink';
 
 const CertLink = ({id, textOnly = false, type}) => {
-  if (type !== 'CERT-Bund' && type !== 'DFN-CERT') {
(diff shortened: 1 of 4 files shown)


janowagner:
I tried the patch and it worked for me: I can see the references in all places I visited.

gsa/src/web/components/link/certlink.js
```diff
@@ -25,7 +25,12 @@ import PropTypes from 'web/utils/proptypes';
 import DetailsLink from './detailslink';
 
 const CertLink = ({id, textOnly = false, type}) => {
-  if (type !== 'CERT-Bund' && type !== 'DFN-CERT') {
+  if (
+    type !== 'cert-bund' &&
+    type !== 'CERT-Bund' &&
+    type !== 'dfn-cert' &&
+    type !== 'DFN-CERT'
```
bjoernricks: Why are both writing (lower and upper case) possible. We should really avoid such things. If we really need this, it would also be better to user type.toLowerCase() instead

PR #1426, opened 2019-05-27
Fix loading timezone and locale at login

swaterkamp:
timezone and i18n are not found in the response.data but response.meta, 
so login variables have to be stitched together from both those fields.

This solves several issues including non-workign user settings and the 
performance page.

**Checklist**:

- [ X ] Tests

### DIFF
--- a/gsa/src/gmp/models/login.js
+++ b/gsa/src/gmp/models/login.js
@@ -24,17 +24,17 @@ import {isDefined} from 'gmp/utils/identity';
 
 class Login {
   constructor(elem) {
-    this.clientAddress = elem.client_address;
-    this.guest = elem.guest;
-    this.locale = elem.i18n;
-    this.role = elem.role;
-    this.severity = elem.severity;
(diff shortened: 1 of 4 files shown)


gsa/src/gmp/models/login.js
```diff
@@ -24,17 +24,17 @@ import {isDefined} from 'gmp/utils/identity';
...
-    this.locale = elem.i18n;
-    this.role = elem.role;
-    this.severity = elem.severity;
-    this.timezone = elem.timezone;
-    this.token = elem.token;
-    this.vendorVersion = elem.vendor_version;
-    this.version = elem.version;
-
-    const unixSeconds = parseInt(elem.session);
+    const {data = {}, meta = {}} = elem;
```
bjoernricks: Hi the Login model should be reworked to get all necessary values as arguments to the constructor again and should not have to know anything about meta and data. The command must pass the arguments from meta and data to the model instead.

PR #1444, opened 2019-06-18
all filter strings to lower case

saberlynx:
Amended FilterTerm to change all filter keywords to lowercase.
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/__tests__/filter.js
+++ b/gsa/src/gmp/models/__tests__/filter.js
@@ -20,7 +20,6 @@ import {isArray} from '../../utils/identity';
 
 import Filter, {UNKNOWN_FILTER_ID} from '../filter';
 import FilterTerm from '../filter/filterterm';
-import {tsExternalModuleReference} from '@babel/types';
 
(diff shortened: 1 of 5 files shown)


bjoernricks:
Could you do me a favor and add two additional tests?

* for Filter.fromString (see https://github.com/greenbone/gsa/blob/master/gsa/src/gmp/models/__tests__/filter.js#L24)
* for parsing filter from keywords (see https://github.com/greenbone/gsa/blob/master/gsa/src/gmp/models/__tests__/filter.js#L109)

gsa/src/gmp/models/__tests__/filter.js
```diff
@@ -20,6 +20,7 @@ import {isArray} from '../../utils/identity';
 
 import Filter, {UNKNOWN_FILTER_ID} from '../filter';
 import FilterTerm from '../filter/filterterm';
+import {tsExternalModuleReference} from '@babel/types';
```
bjoernricks: It seems vscode added some additional import accidentially. Could you remove this one?

```suggestion
```

PR #1451, opened 2019-06-24
updated user auth_method to camelCase

saberlynx:
updated variables to camelCase. Everything seems to work from manual testing. I ignored the commands/users code because changing it breaks the functionality to change password. Also ignored parts of dialog.
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/gmp/models/__tests__/user.js
+++ b/gsa/src/gmp/models/__tests__/user.js
@@ -118,10 +118,10 @@ describe('User model tests', () => {
     const user2 = new User(elem2);
     const user3 = new User({});
 
-    expect(user1.auth_method).toEqual(AUTH_METHOD_LDAP);
-    expect(user2.auth_method).toEqual(AUTH_METHOD_RADIUS);
+    expect(user1.authMethod).toEqual(AUTH_METHOD_LDAP);
+    expect(user2.authMethod).toEqual(AUTH_METHOD_RADIUS);
     expect(user1.sources).toBeUndefined();
-    expect(user3.auth_method).toEqual(AUTH_METHOD_PASSWORD);
+    expect(user3.authMethod).toEqual(AUTH_METHOD_PASSWORD);
   });
 });
 
--- a/gsa/src/gmp/models/user.js
+++ b/gsa/src/gmp/models/user.js
@@ -77,13 +77,13 @@ class User extends Model {
     if (isDefined(elem.sources)) {
       const {source} = elem.sources;
       if (source === 'ldap_connect') {
-        ret.auth_method = AUTH_METHOD_LDAP;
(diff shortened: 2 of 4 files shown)


gsa/src/web/pages/users/row.js
```diff
@@ -109,7 +109,8 @@ const Row = ({
     </DetailsLink>
   ));
 
-  const auth_method = convert_auth_method(entity.auth_method);
+  const authMethod = convert_auth_method(entity.authMethod);
+  delete entity.authMethod;
```
bjoernricks:
```suggestion
```

gsa/src/gmp/models/user.js
```diff
@@ -77,13 +77,13 @@ class User extends Model {
...
+        ret.authMethod = AUTH_METHOD_LDAP;
       } else if (source === 'radius_connect') {
-        ret.auth_method = AUTH_METHOD_RADIUS;
+        ret.authMethod = AUTH_METHOD_RADIUS;
       }
       delete ret.sources;
     } else {
-      ret.auth_method = AUTH_METHOD_PASSWORD;
+      ret.authMethod = AUTH_METHOD_PASSWORD;
     }
```
bjoernricks:
```suggestion
    delete ret.auth_method;
```

gsa/src/gmp/models/__tests__/user.js
```diff
@@ -118,10 +118,10 @@ describe('User model tests', () => {
     const user2 = new User(elem2);
     const user3 = new User({});
 
-    expect(user1.auth_method).toEqual(AUTH_METHOD_LDAP);
-    expect(user2.auth_method).toEqual(AUTH_METHOD_RADIUS);
+    expect(user1.authMethod).toEqual(AUTH_METHOD_LDAP);
+    expect(user2.authMethod).toEqual(AUTH_METHOD_RADIUS);
     expect(user1.sources).toBeUndefined();
-    expect(user3.auth_method).toEqual(AUTH_METHOD_PASSWORD);
+    expect(user3.authMethod).toEqual(AUTH_METHOD_PASSWORD);
```
bjoernricks: We could even add a check for having converted the auth_method property to the tests :-)

```suggestion
    expect(user3.auth_method).toBeUndefined()
```

saberlynx: Looking at model data from the tests, there doesn't appear to be an original auth_method. we determined the authMethod based on a property called "sources". Perhaps something like this? `expect(user3.sources).toBeUndefined();`

bjoernricks: Ah ok. It isn't parsed from an auth_method property. So we don't need another test for this :-)

PR #1452, opened 2019-06-25
changed scanner vars to camelCase

saberlynx:
Updated scanner vars (ca_pub and param_type) to camelCase. Ignored component, command and dialog (except to change '.' to '_'). Also updated tests.

**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/scanners/dialog.js
+++ b/gsa/src/web/pages/scanners/dialog.js
@@ -289,7 +289,7 @@ class ScannerDialog extends React.Component {
                       </Divider>
                     </Layout>
                     {is_edit && isDefined(state.ca_pub) && (
-                      <CertStatus info={state.ca_pub.info} />
+                      <CertStatus info={state.ca_pub_info} />
                     )}
                   </FormGroup>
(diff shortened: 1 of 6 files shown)


saberlynx:
> IMHO https://github.com/greenbone/gsa/blob/master/gsa/src/web/pages/scanners/component.js#L83 must be updated too. Currently the ca_pub prop passed to the scanner dialog is always undefined.

Edited. Hope it's correct.

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -289,7 +289,7 @@ class ScannerDialog extends React.Component {
                       </Divider>
                     </Layout>
                     {is_edit && isDefined(state.ca_pub) && (
-                      <CertStatus info={state.ca_pub.info} />
+                      <CertStatus info={state.ca_pub_info} />
```
bjoernricks: I don't think this will work because the dialog state doesn't contain a ca_pub_info property. But nevertheless for me it seems that was already broken before...

PR #1458, opened 2019-06-26
changed certbund vars to camelCase

saberlynx:
changed variables additional_information, remote_attack, reference_source, reference_url and revision_history
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/gmp/models/certbund.js
+++ b/gsa/src/gmp/models/certbund.js
@@ -29,7 +29,7 @@ class CertBundAdv extends Info {
   parseProperties(elem) {
     const ret = super.parseProperties(elem, 'cert_bund_adv');
 
-    ret.severity = parseSeverity(ret.max_cvss);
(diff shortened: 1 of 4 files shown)


gsa/src/gmp/models/certbund.js
```diff
@@ -29,26 +29,26 @@ class CertBundAdv extends Info {
...
 
     ret.categories = [];
     ret.description = [];
     ret.cves = [];
-    ret.additional_information = [];
+    ret.additionalInformation = [];
 
-    if (isDefined(ret.raw_data) && isDefined(ret.raw_data.Advisory)) {
+    if (isDefined(elem.raw_data) && isDefined(elem.raw_data.Advisory)) {
       const {raw_data} = ret;
```
bjoernricks: Better to use elem instead of ret here too
```suggestion
       const {raw_data} = elem;
```

PR #1471, opened 2019-07-10
Update gsa-zh_TW.po

jasoncheng7115:
The number of translations from 10% to 22%

**Checklist**:

- [] Tests

### DIFF
--- a/gsa/po/gsa-zh_TW.po
+++ b/gsa/po/gsa-zh_TW.po
@@ -448,7 +448,7 @@ msgstr "主機 (名稱)"
 #: gsa/src/web/pages/certbund/filterdialog.js:34
 #: gsa/src/web/pages/certbund/table.js:70
 msgid "Created"
-msgstr "已建立"
+msgstr "建立日期"
 
 #: gsa/src/web/pages/results/table.js:63
 msgid "Delta"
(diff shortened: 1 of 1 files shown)


gsa/po/gsa-zh_TW.po
```diff
@@ -448,7 +448,7 @@ msgstr "主機 (名稱)"
 #: gsa/src/web/pages/certbund/filterdialog.js:34
 #: gsa/src/web/pages/certbund/table.js:70
 msgid "Created"
-msgstr ""
+msgstr "已建立"
```
saberlynx: Thanks again for all the translations! In the table and in the filter dialog, my understanding is that we use "Created" as a short hand for "Time/Date Created". Perhaps something like "建立日期"? I speak Cantonese, so this might not be right in TW Chinese, so please correct me as you see fit. :)

gsa/po/gsa-zh_TW.po
```diff
@@ -681,12 +681,12 @@ msgstr ""
...
 #: gsa/src/web/pages/cves/filterdialog.js:34 gsa/src/web/pages/cves/table.js:70
 msgid "Complexity"
-msgstr ""
+msgstr "複雜性"
 
 #: gsa/src/web/pages/cves/filterdialog.js:54
 #: gsa/src/web/pages/cves/table.js:110
 msgid "Published"
-msgstr ""
+msgstr "已發佈"
```
saberlynx: Same as "Created"; we use "published" as a shorthand for "time published" and I'm understanding this as "already published" or "having been published". Please correct me if I'm wrong.

gsa/po/gsa-zh_TW.po
jasoncheng7115: Hoooo, you're right, "建立日期" is actually better than "已建立".
I'm going to change that. Thank you!

gsa/po/gsa-zh_TW.po
jasoncheng7115: I'm going to change that. Thank you!

PR #1476, opened 2019-07-15
Update gsa-zh_TW.po

jasoncheng7115:
The number of translations from 22% to 32%

**Checklist**:

- [] Tests


gsa/po/gsa-zh_TW.po
```diff
@@ -1469,27 +1469,27 @@ msgstr ""
...
+msgstr "下載 TLS 憑證"
 
 #: gsa/src/web/pages/reports/tlscertificatestable.js:165
 msgid "No TLS Certificates available"
 msgstr ""
 
 #: gsa/src/web/pages/reports/row.js:53
 msgid "Scan is active"
-msgstr ""
+msgstr "掃描啟動"
```
saberlynx: I'm understanding "scan start" from this and I think we mean "scan is running" (this tooltip displays when the user tries to delete a report based on a scan that is still running). How about something like 掃描正在運行 or 掃描運行中?

jasoncheng7115: Wow, your Chinese translation is better than mine...Let me admire!
I will modify it, thanks for your comment!

jasoncheng7115: I changed it to "掃描進行中".

jasoncheng7115:
> Thanks for your hard work!!

I'll keep working! :D

PR #1477, opened 2019-07-16
added relationselector and severityvaluesgroup component

saberlynx:
- added a component "RelationSelector" to select between '=', '<' and '>'. All it does is select relation.
- added a component "SeverityValuesGroup" replacing CvssBaseGroup (it is now more general) that has "RelationSelector" as child, enabling user to do eg. "cvss_base>3"

**Checklist**:

- [x] Tests


gsa/src/web/components/powerfilter/relationselector.js
```diff
@@ -0,0 +1,44 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+/* eslint-disable react/prop-types */
+import React from 'react';
+import Select from 'web/components/form/select';
+import PropTypes from 'web/utils/proptypes.js';
+
+const RelationSelector = ({relation, onChange}) => (
+  <div>
```
bjoernricks: I suppose the extra div shouldn't be necessary. Just return the Select component.

gsa/src/web/components/powerfilter/relationselector.js
```diff
@@ -0,0 +1,44 @@
...
+    <Select
+      value={relation}
+      onChange={onChange}
+      items={[
+        {label: 'is equal to', value: '='},
+        {label: 'is greater than', value: '>'},
```
bjoernricks:
```suggestion
        {label: _('is greater than'), value: '>'},
```

gsa/src/web/components/powerfilter/relationselector.js
```diff
@@ -0,0 +1,44 @@
...
+        {label: 'is less than', value: '<'},
```
bjoernricks:
```suggestion
        {label: _('is less than'), value: '<'},
```

gsa/src/web/components/powerfilter/severityvaluesgroup.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  render() {
+    const severity = parseSeverity(this.props.filter.get(this.props.name));
+    const newRelation = this.state.relation;
+    const keyword = this.props.name;
+    return (
+      <div>
+        <FormGroup title={this.props.title}>
+          <RelationSelector
```
bjoernricks: Would be nice to add an additional <Divider> around RelationSelector and NumberField here to add some margin between them.
<Divider><ReleationSelector .../><NumberField .../></Divider>

gsa/src/web/components/powerfilter/relationselector.js
saberlynx: in the future should all labels/titles have this format?

PR #1478, opened 2019-07-16
Fix displaying ical events without timezone info

bjoernricks:
Consider datetimes in ical data to be in the timezone passed in the
additional timezone field. Before the schedule datetimes onyl would have
been displayed in this timezone.

**Checklist**:

- [x] Tests


bjoernricks:
Replaced with https://github.com/greenbone/gsa/pull/1479

Github didn't let me change the target branch :disappointed:

PR #1485, opened 2019-07-17
added grow option to multiselect

saberlynx:
**Checklist**:

And updated snapshots.
- [] Tests

### DIFF
--- a/gsa/src/web/components/form/multiselect.js
+++ b/gsa/src/web/components/form/multiselect.js
@@ -176,6 +176,7 @@ class MultiSelect extends React.Component {
       items,
       menuPosition = 'adjust',
       width = DEFAULT_WIDTH,
+      grow = '0',
     } = this.props;
 
     const {search, selectedItems} = this.state;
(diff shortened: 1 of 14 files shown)


gsa/src/web/components/form/multiselect.js
```diff
@@ -176,6 +176,7 @@ class MultiSelect extends React.Component {
       items,
       menuPosition = 'adjust',
       width = DEFAULT_WIDTH,
+      grow = '0',
```
bjoernricks: Not sure if grow should get a default value? @swaterkamp what do you think?

saberlynx: the default value of 0 doesn't do anything as far as i can tell, so if we want the old appearance, setting it to 0 will give us that. I think only positive integers make a difference.

https://stackoverflow.com/questions/28473807/how-does-flex-grow0-get-interpreted

swaterkamp: We could set a default of 0 in the SelectContainer, where it actually will be applied in the end. I find NOT growing to be the right option as default and just explicitly setting it to 1 if we REALLY want it.

PR #1486, opened 2019-07-17
Add config files to run gsad as a unix daemon

bjoernricks:
Add config files for systemd, start arguments and logrotate.

Replaces #760

**Checklist**:

- [n/a] Tests


bjoernricks:
>   daily
>    rotate 7

I am going to remove them.

> copytruncate may cause lost logs, why we don't use create and kill HUP?

Because I didn't wanted to restart the gsad daemon and therefore not answering to requests. Currently SIGHUP is ignored https://github.com/greenbone/gsa/blob/gsa-8.0/gsad/src/gsad.c#L2489

> User=root
> Group=root

gsad should not run as root.

> After=openvas-scanner.service
> After=gvmd.service
> Wants=gvmd.service

> Don't we need gvmd & openvas-scanner before gsad ?

Yes but currently we don't have service files for them.

This PR is intended as a start to get some service files into the releases. I am fine with adding further PRs to improve them.

bjoernricks:
> copytruncate may cause lost logs, why we don't use create and kill HUP?

Just for curiosity how can log be lost in that case?

hsntgm:
> copytruncate
Using copytruncate with size, maxsize options is more secure.

While the truncate operation is very fast, the copy can be quite long (depending of how big is your logfile). Moreover, some log entry can be lost during the time between the copy operation and the truncate.

Also thank you for the info:

> Currently SIGHUP is ignored https://github.com/greenbone/gsa/blob/gsa-8.0/gsad/src/gsad.c#L2489

gvmd and open-scanner daemon are in the same situation about SIGHUP ? Then i Hhave to fix all my service files and logrotate files.

bjoernricks:
> gvmd and open-scanner daemon are in the same situation about SIGHUP ?

At least gvmd seems to ignore it too https://github.com/greenbone/gvmd/blob/master/src/gvmd.c#L2605

PR #1494, opened 2019-07-19
Added bullet point separator to horizontal lists

saberlynx:
**Checklist**:

- [] Tests


saberlynx:
> I really would like to introduce an own component for displaying the bullet. With an own component it is easier to change the behavior in future.

I added it. I still have to write a test for the horizontalsep component. I have some options in there because I already wrote them :P

PR #1508, opened 2019-07-24
Don't logout user at login page

bjoernricks:
Redirect to main page if the user is already logged in when visiting the
login page instead of clearing his token.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/login/loginpage.js
+++ b/gsa/src/web/pages/login/loginpage.js
@@ -53,6 +53,8 @@ import {
   setIsLoggedIn,
 } from 'web/store/usersettings/actions';
 
+import {isLoggedIn} from 'web/store/usersettings/selectors';
+
 import LoginForm from './loginform';
 
 const log = logger.getLogger('web.login');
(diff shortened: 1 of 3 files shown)


gsa/src/web/pages/login/loginpage.js
```diff
@@ -235,11 +241,15 @@ const mapDispatchToProps = (dispatch, {gmp}) => ({
   setIsLoggedIn: value => dispatch(setIsLoggedIn(value)),
 });
 
+const mapStateToProp = (rootState, ownProps) => ({
```
swaterkamp: For consistency, could you rename this to mapStateToProps?
```suggestion
const mapStateToProps = (rootState, ownProps) => ({
```

gsa/src/web/pages/login/loginpage.js
```diff
@@ -235,11 +241,15 @@ const mapDispatchToProps = (dispatch, {gmp}) => ({
...
+  isLoggedIn: isLoggedIn(rootState),
+});
+
 export default compose(
   withRouter,
   withGmp,
   connect(
-    null,
+    mapStateToProp,
```
swaterkamp: Same here :)
```suggestion
    mapStateToProps,
```

PR #1509, opened 2019-07-24
Use HSTS and the secure cookie by default.

tuxmaster5000:
**Checklist**:
- [-] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use HTTPS for documentation links
 - Cleanup and improve handling of http parameters and arguments for gmp
   requests in gsad [#1355](https://github.com/greenbone/gsa/pull/1355)
+- Use hsts by default
+- Use secure cookies
 
 ### Fixed
 - Fixed outer click issues for multi select and select boxes
--- a/gsad/config/gsad.service.in
+++ b/gsad/config/gsad.service.in
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 Type=forking
 EnvironmentFile=-${DEFAULT_CONFIG_DIR}/gsad
-ExecStart=${SBINDIR}/gsad --listen $GSA_ADDRESS --port $GSA_PORT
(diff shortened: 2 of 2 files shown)


cfi-gb:
I'm strongly voting against using HSTS by default.

Rationale:

Any certificate problem makes the GSA unavailable via HTTP. An administrator should know the implications when enabling HSTS before enabling it, that's why it should be only explicitly enabled.

tuxmaster5000:
But sending the password for gsa via http is complete insecure. So the best way will be use HSTS or complete disable http.

bjoernricks:
> But sending the password for gsa via http is complete insecure. So the best way will be use HSTS or complete disable http.

You are completely right. On productive servers gsad should always be run with https. Also we are warning the user for an unencrypted connection if gsa is running via http. But currently I am not sure if we should make both options the default in our service file directly. Often gsad is running behind a https proxy or only for development.

Feel free to use #1514 and set  `--http-sts --secure-cookie`  as default settings on your system.

cfi-gb:
As an additional note (please correct me if i'm wrong):

gsad is not offering the login prompt via HTTP by default (this would need to be enforced via ``--http-only``) and always redirects from HTTP to HTTPs (behavior can be disabled completely via ``--no-redirect``).

The ``--secure-cookie`` still makes sense as long as gsad is only setting this for cookies created by a login via a secured connection.

bjoernricks:
> As an additional note (please correct me if i'm wrong):
> 
> gsad is not offering the login prompt via HTTP by default (this would need to be enforced via `--http-only`) and always redirects from HTTP to HTTPs (behavior can be disabled completely via `--no-redirect`). Sure that there are other attack vectors like a MITM but for this the admin can add measures like the `--http-sts` if required.
> 
> The `--secure-cookie` still makes sense as long as gsad is only setting this for cookies created by a login via a secured connection.

Indeed by default gsad redirects from the http port to the https port if `--http-only` is not set.

If gsad is running without `--http-only` `Set-Cookie: secure` is used by default and therefore the `--secure-cookie` argument is unnecessary.

Additionally to this observations I don't think hsts should be activated by default. Therefore I am closing this PR for now. Feel free to reopen if you think I am missing something.

PR #1520, opened 2019-07-26
Change authentification to authentication.

cfi-gb:
While ``authentification`` is correct according to e.g. https://english.stackexchange.com/questions/7844/is-authentification-a-real-word or https://en.wiktionary.org/wiki/authentification rename the word to ``authentication`` to keep / use the same wording like in the documentation:

https://docs.greenbone.net/GSM-Manual/gos-4/en/search.html?q=authentication

Note that even some spellcheckers like the build-in of Firefox is also directly reporting a spelling error for ``authentification``.

**Checklist**:

- [] Tests


cfi-gb:
Replacement in https://github.com/greenbone/gsa/pull/1522 as i probably broke something while trying to rebase.

PR #1528, opened 2019-07-30
Fix host details OS link

swaterkamp:
Instead of linking to a broken cpe page, link to the corresponding OS

**Checklist**:

- [ N/A ] Tests

### DIFF
--- a/gsa/src/web/pages/hosts/detailspage.js
+++ b/gsa/src/web/pages/hosts/detailspage.js
@@ -16,6 +16,9 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+
+import 'core-js/library/fn/array/find';
+
 import React from 'react';
 
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/hosts/detailspage.js
```diff
@@ -164,13 +164,24 @@ const Hop = styled.div`
...
     : undefined;
+
+  const bestOsMatchingIdentifier = identifiers.find(
+    identifier => identifier.name === 'OS' && identifier.value === os_cpe,
+  );
+
+  const os = isDefined(bestOsMatchingIdentifier)
+    ? bestOsMatchingIdentifier.os
+    : {};
+  const bestMatchingOsId = os.id;
```
bjoernricks: Should be changed to
```js
  const bestMatchingOsId = isDefined(bestOsMatchingIdentifier)
    ? bestOsMatchingIdentifier.os.id
    :  undefined;
```

gsa/src/web/pages/hosts/detailspage.js
```diff
@@ -164,13 +164,24 @@ const Hop = styled.div`
...
+  const {details = {}, identifiers = [], routes = [], severity} = entity;
+
   const os_cpe = isDefined(details.best_os_cpe)
     ? details.best_os_cpe.value
     : undefined;
   const os_txt = isDefined(details.best_os_txt)
     ? details.best_os_txt.value
     : undefined;
+
+  const bestOsMatchingIdentifier = identifiers.find(
```
bjoernricks: Maybe that's something for the being in the model already?

bjoernricks: Btw. please check if a core-js import is necessary for IE 11 find support.

swaterkamp: I thought about putting it in the model, but as long as we need this info only in one place, I think it makes more sense to load it here rather than for all hosts everywhere in GSA.

PR #1529, opened 2019-07-30
refactored severityvaluesgroup with react hooks

saberlynx:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/components/powerfilter/severityvaluesgroup.js
+++ b/gsa/src/web/components/powerfilter/severityvaluesgroup.js
@@ -24,7 +24,7 @@ import FormGroup from 'web/components/form/formgroup.js';
 import {parseSeverity} from 'gmp/parser.js';
 import RelationSelector from 'web/components/powerfilter/relationselector';
 import NumberField from 'web/components/form/numberfield';
-import Divider from '../layout/divider';
+import Divider from 'web/components/layout/divider';
 
 const SeverityValuesGroup = ({filter, name, title, onChange}) => {
   const [rel, setRel] = useState('=');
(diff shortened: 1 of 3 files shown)


gsa/src/web/components/powerfilter/severityvaluesgroup.js
```diff
@@ -16,64 +16,41 @@
...
+
+const SeverityValuesGroup = ({filter, name, title, onChange}) => {
+  const [rel, setRel] = useState('=');
+  const severity = parseSeverity(filter.get(name));
+  const keyword = name;
+
+  return (
+    <FormGroup title={title}>
+      <Divider>
+        <RelationSelector relation={rel} onChange={newRel => setRel(newRel)} />
```
bjoernricks: I am seeing a potential bug here. Changing the relation isn't propagated to the calling component anymore. You should call the onChange handler here too. Something like
```jsx
<RelationSelector relation={rel} onChange={newRel => {setRel(newRel); onChange(severity, name, newRel);} />
```

gsa/src/web/components/powerfilter/severityvaluesgroup.js
```diff
@@ -16,64 +16,41 @@
...
-          />
-        </Divider>
-      </FormGroup>
-    );
-  }
-}
+import Divider from 'web/components/layout/divider';
+
+const SeverityValuesGroup = ({filter, name, title, onChange}) => {
+  const [rel, setRel] = useState('=');
```
bjoernricks: I am thinking about another issue which has been in the component before already. Could it be possible that the relation is not "=" when initially being rendered. I guess we have to get the relation from the filter.

```js
const term = filter.getTerm(name);
severity = isDefined(term) ? parseSeverity(term.value) : undefined;
relation = isDefined(term) ? term.relation : '=';
...
```

saberlynx: Thanks for this catch. I put the relation bit into the useState argument.

gsa/src/web/components/powerfilter/severityvaluesgroup.js
saberlynx: thanks. done.

PR #1537, opened 2019-08-01
Don't show last Comma in hosts lists at target details

swaterkamp:
Also fix small typo

**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/alerts/details.js
+++ b/gsa/src/web/pages/alerts/details.js
@@ -147,7 +147,7 @@ const AlertDetails = ({capabilities, entity, links = true}) => {
 
           {tasks.length > 0 && (
             <TableRow>
-              <TableData>{_('Task using this Alert')}</TableData>
+              <TableData>{_('Tasks using this Alert')}</TableData>
               <TableData>
                 <HorizontalSep wrap>
                   {tasks.map(task => (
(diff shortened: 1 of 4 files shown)


gsa/src/web/pages/alerts/details.js
```diff
@@ -147,10 +147,10 @@ const AlertDetails = ({capabilities, entity, links = true}) => {
 
           {tasks.length > 0 && (
             <TableRow>
-              <TableData>{_('Task using this Alert')}</TableData>
+              <TableData>{_('Tasks using this Alert')}</TableData>
               <TableData>
                 <HorizontalSep wrap>
-                  {tasks.map(task => (
+                  {tasks.map((task, index) => (
```
bjoernricks: Why did you add the unused index here?

swaterkamp: It was an accident... I forgot to remove it.

swaterkamp:
> Could you check if something like
> 
> ```js-jsx
> <HorizontalSep seperator="," wrap>....map(host => <span>{host}</span>)</HorizontalSep>
> ```
> 
> could fit here too?

I tried the HorizontalSep and it's not optimal for this use case. It's more useful for separating items with a centered item and not a left-bound one. Also the whole string got shifted to the left as well, leading to a broken left-alignment of information.

swaterkamp:
> just use
> 
> ```diff
> [...]
> ```

I didn't have in mind to actually change the HorizontalSep. Done now, looks good.

PR #1545, opened 2019-08-05
(Multi)Select clickable area

swaterkamp:
**Checklist**:

- [ X ] Tests

### DIFF
--- a/gsa/src/web/components/layout/withLayout.js
+++ b/gsa/src/web/components/layout/withLayout.js
@@ -69,7 +69,7 @@ const withLayout = (defaults = {}) => Component => {
     basis: PropTypes.string,
     flex: PropTypes.oneOf([true, 'column', 'row']),
     grow: PropTypes.oneOfType([
-      PropTypes.oneOf([true]),
+      PropTypes.oneOf([true, false]),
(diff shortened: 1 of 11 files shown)


gsa/src/web/components/layout/withLayout.js
```diff
@@ -69,7 +69,7 @@ const withLayout = (defaults = {}) => Component => {
     basis: PropTypes.string,
     flex: PropTypes.oneOf([true, 'column', 'row']),
     grow: PropTypes.oneOfType([
-      PropTypes.oneOf([true]),
+      PropTypes.oneOf([true, false]),
```
bjoernricks: I don't think this is correct. Allowed are `<Layout grow/> <Layout grow={true}/> <Layout grow="1"/> <Layout grow="0"/> <Layout grow={2}/> <Layout grow={0}/>` but false shouldn't be used at the moment. I suppose it will work but syntactically  a css flex-grow value of false doesn't exist. false is not handled within withLayout besides true which is converted to 1. So either withLayout should be adjusted to convert false into 0 or the expressions like `grow={!itemsAreSelected}` should be converted to `grow={itemsAreSelected == true ? 1 : 0}

PR #1563, opened 2019-08-12
Fix "once" option for schedules in TaskDialog

swaterkamp:
**Checklist**:

- [] Tests


swaterkamp:
Seems like it was such a small, easy thing I couldn't find it. It seems to work for me and I don't know if I should be happy or sad ;)

PR #1564, opened 2019-08-13
Search for "yarnpkg" in FindYarn.cmake

cfi-gb:
On Debian >= Buster yarn is now shipped but the binary is called "yarnpkg" instead of "yarn":

https://packages.debian.org/buster/all/yarnpkg/filelist

**Checklist**:

- [] Tests


bjoernricks:
Backport of #1463

PR #1571, opened 2019-08-14
TLS Certificate Assets

swaterkamp:
This PR concerns some initial implementations in order to introduce TLS Certificate assets to GSA.
The final implementation of the new feature will be done via several PRs in order to keep the review easier.

**Checklist**:

- [] Tests


gsa/src/gmp/commands/tlscertificates.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+class TlsCertificatesCommand extends EntitiesCommand {
+  constructor(http) {
+    super(http, 'tls_certificate', TlsCertificate);
+  }
+
+  get(params = {}, options) {
+    return super.get({...params, details: 1}, options);
```
bjoernricks: Is this really necessary? Do we need to send details=1?

gsa/src/gmp/models/tlscertificate.js
```diff
@@ -0,0 +1,76 @@
...
+            sourceHostIps.push(source.location.host.ip);
+          }
+          if (isDefined(source.location.port)) {
+            sourcePorts.push(source.location.port);
+          }
+        }
+      });
+    }
+
+    ret.sourceReportIds = unique(sourceReportIds);
```
bjoernricks: It would be better to use an objects with port, ip and origin_id as keys and converting the objects to arrays afterwards. Even easier would be to use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

gsa/src/gmp/utils/array.js
```diff
@@ -119,4 +119,17 @@ export const arraysEqual = (arr1, arr2) => {
...
+/**
+ * Get only unique values from an array
+ *
+ * @param {Array} array      Array to unifiy values
+ *
+ * @returns {Array} array with only unique values
+ */
+export const unique = array => {
+  return array.filter(function(value, index, self) {
+    return self.indexOf(value) === index;
```
bjoernricks: I don't think this algorithm has a complexity of O(n^2) which isn't very good.

gsa/src/web/pages/tlscertificates/details.js
```diff
@@ -0,0 +1,110 @@
...
+  selector as tlsCertificateSelector,
+} from 'web/store/entities/tlscertificates';
+
+import compose from 'web/utils/compose';
+import withGmp from 'web/utils/withGmp';
+import PropTypes from 'web/utils/proptypes';
+import {renderYesNo} from 'web/utils/render';
+
+const TlsCertificateDetails = ({entity, links = true}) => {
+  console.log('enityt details', entity);
```
bjoernricks:
```suggestion
```

gsa/src/web/pages/tlscertificates/details.js
```diff
@@ -0,0 +1,110 @@
...
+  return (
+    <React.Fragment>
```
bjoernricks: Fragment is not necessary here.

gsa/src/web/pages/tlscertificates/table.js
```diff
@@ -0,0 +1,116 @@
...
+  currentSortDir: PropTypes.string,
+  links: PropTypes.bool,
+  sort: PropTypes.bool,
+  onSortChange: PropTypes.func,
+};
+
+const Footer = createEntitiesFooter({
+  span: 9,
+  delete: true,
+  download: 'os.xml',
```
bjoernricks:
```suggestion
  download: 'tls-certificates.xml',
```

gsa/src/gmp/commands/tlscertificates.js
swaterkamp: I included it for the first implementation in order to test things and because I didn't know what to put into the details. As of now, this is not really necessary, but to be perfectly honest, I don't want to spent a lot of time changing these things for the PR and then immediately change them back to continue developing. For the final implementation this might look different ;)

bjoernricks: Just remove the get method completely and use the one from the parent class :)

swaterkamp: Done :)

gsa/src/web/pages/tlscertificates/details.js
swaterkamp: This should have been gone already, it's not in my local changes anymore and I rebased everything. This is weird...

gsa/src/web/pages/tlscertificates/details.js
```diff
@@ -0,0 +1,101 @@
...
+          {isDefined(entity.comment) && (
+            <TableRow>
+              <TableData>{_('Comment')}</TableData>
+              <TableData>{entity.comment}</TableData>
+            </TableRow>
+          )}
+          {isDefined(entity.trust) && (
+            <TableRow>
+              <TableData>{_('Trusted')}</TableData>
+              <TableData>{renderYesNo(entity.trust)}</TableData>
```
bjoernricks: We should really parse trust as boolean in the model.

gsa/src/web/pages/tlscertificates/details.js
```diff
@@ -0,0 +1,101 @@
...
+            </TableRow>
+          )}
+          {isDefined(entity.valid) && (
+            <TableRow>
+              <TableData>{_('Valid')}</TableData>
+              <TableData>{renderYesNo(entity.valid)}</TableData>
```
bjoernricks: Same as trust should be parsed as boolean in the model.

gsa/src/gmp/models/tlscertificate.js
bjoernricks: I meant to use the Set here directly. Put the ports, ips and origin_ids into a set directly and convert it into an array if your are done.
e.g.
```
const sourceReportIds = new Set();
....
sourceReportIds.add(source.origin.origin_id)
...
ret.sourceReportIds = [...sourceReportIds];

File: gsa/src/web/pages/tlscertificates/details.js (hunk 7 above)
Comment (bjoernricks) [reply gsa/src/web/pages/tlscertificates/details.js]: Not sure why you marked this as resolved. Still not a boolean in the model as far as i can see.

File: gsa/src/web/pages/tlscertificates/details.js (hunk 8 above)
Comment (bjoernricks) [reply gsa/src/web/pages/tlscertificates/details.js]: Same as above.

File: gsa/src/web/pages/tlscertificates/details.js [hunk 9]
```diff
@@ -0,0 +1,101 @@
...
+
+const mapStateToProps = (rootState, {entity = {}}) => {
+  const tlsCertificateSel = tlsCertificateSelector(rootState);
+  return {
+    tlsCertificate: tlsCertificateSel.getEntity(entity.id),
+  };
+};
+
+const mapDispatchToProps = (dispatch, {gmp}) => ({
+  loadTlsCertificate: id => dispatch(loadTlsCertificate(gmp)({id, details: 1})),
```
bjoernricks:
```suggestion
  loadTlsCertificate: id => dispatch(loadTlsCertificate(gmp)({id})),
```
I really don't think details is necessary here. And also i am not sure if should be `loadTlsCertificate(gmp)(id)` instead.

PR #1573, opened 2019-08-16
Reuse scanconfig edit dialogs for policies and add tests for san config edit dialogs

sarahd93:
**Checklist**:

- [x] Tests


gsa/src/web/pages/policies/component.js
```diff
@@ -751,12 +753,13 @@ class PolicyComponent extends React.Component {
...
           <EditPolicyFamilyDialog
-            policy={policy}
-            policyName={policyName}
+            config={policy}
+            config_name={policyName}
             family_name={family_name}
             id={id}
             nvts={nvts}
             selected={selected}
+            text={_('Policy')}
```
swaterkamp: Could you rename this to something more meaningful from which it is possible to figure out what the text is used for?

gsa/src/web/pages/scanconfigs/editconfigfamilydialog.js
```diff
@@ -197,6 +197,7 @@ class EditDialogComponent extends React.Component {
       config_name,
       family_name,
       id,
+      text,
```
swaterkamp: Please require the text param to be set as you do in the next commit.

gsa/src/web/pages/policies/component.js
```diff
@@ -768,12 +766,13 @@ class PolicyComponent extends React.Component {
...
           <EditNvtDetailsDialog
-            policy={policy}
-            policyName={policyName}
+            config={policy}
+            config_name={policyName}
             family_name={family_name}
             manual_timeout={manual_timeout}
             nvt={nvt}
             preference_values={preference_values}
+            text={_('Policy')}
```
swaterkamp: Same as with the the text param in the previous commit. This time you are requiring the text already, though.

PR #1574, opened 2019-08-19
trigger alert button disabled in report details page when no permission

saberlynx:
**Checklist**:

- [] Tests


gsa/src/web/pages/reports/alertactions.js
```diff
@@ -205,6 +210,7 @@ class AlertActions extends React.Component {
 
 AlertActions.propTypes = {
   alerts: PropTypes.array,
+  capabilities: PropTypes.array,
```
swaterkamp: We have a Capabilities proptype. Please use that one. Also, I guess this should be required, as l.172 would crash if caps are missing.
```suggestion
  capabilities: PropTypes.capabilities.isRequired,
```

PR #1578, opened 2019-08-24
Extend TLS Certificate Assets

swaterkamp:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/tlscertificates/details.js
+++ b/gsa/src/web/pages/tlscertificates/details.js
@@ -24,6 +24,10 @@ import _ from 'gmp/locale';
 
 import {isDefined} from 'gmp/utils/identity';
 
+import DateTime from 'web/components/date/datetime';
+
+import Layout from 'web/components/layout/layout';
+
 import InfoTable from 'web/components/table/infotable';
 import TableData from 'web/components/table/data';
 import TableBody from 'web/components/table/body';
@@ -43,7 +47,7 @@ import {renderYesNo} from 'web/utils/render';
 
 const TlsCertificateDetails = ({entity, links = true}) => {
   return (
-    <React.Fragment>
+    <Layout grow="1" flex="column">
       <InfoTable>
         <colgroup>
           <Col width="10%" />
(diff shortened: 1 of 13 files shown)


gsa/src/web/pages/tlscertificates/details.js
```diff
@@ -68,15 +72,31 @@ const TlsCertificateDetails = ({entity, links = true}) => {
               <TableData>{renderYesNo(entity.valid)}</TableData>
             </TableRow>
           )}
+          {isDefined(entity.activationTime) && (
+            <TableRow>
+              <TableData>{_('Activates')}</TableData>
+              <TableData>
+                {dateTimeWithTimeZone(entity.activationTime)}
```
bjoernricks: Should use the DateTime component instead of the dateTimeWithTimeZone function

gsa/src/web/pages/tlscertificates/details.js
```diff
@@ -68,15 +72,31 @@ const TlsCertificateDetails = ({entity, links = true}) => {
...
+              </TableData>
+            </TableRow>
+          )}
+          {isDefined(entity.expirationTime) && (
+            <TableRow>
+              <TableData>{_('Expires')}</TableData>
+              <TableData>
+                {dateTimeWithTimeZone(entity.expirationTime)}
```
bjoernricks: Same here. It should use the DateTime component instead of the dateTimeWithTimeZone function

gsa/src/web/pages/tlscertificates/detailspage.js
```diff
@@ -0,0 +1,295 @@
...
+};
+
+ToolBarIcons.propTypes = {
+  entity: PropTypes.model.isRequired,
+  onTlsCertificateDeleteClick: PropTypes.func.isRequired,
+  onTlsCertificateDownloadClick: PropTypes.func.isRequired,
+};
+
+const Details = ({entity, ...props}) => {
+  const showSourceBlock = [entity.sourceReportIds].reduce(
```
bjoernricks: I don't get this line :-) Any reason why not using `const showSourceBlock = entity.sourceReportIds.length > 0`?

swaterkamp: Yes, it was late at night ;) In the beginning I just did for the showXY what we do in the menubar. Turned out to be much simpler but I didn't remove the leftover...

PR #1588, opened 2019-08-27
Disable editing settings if permission is denied

swaterkamp:
**Checklist**:

- [] Tests


swaterkamp:
We should. I'll update the Changelog with the next PR that receives an entry

PR #1595, opened 2019-08-29
Add download functionality to TLS Certificates

swaterkamp:
This PR adds a download functionality to the already existing export-method. I kept the latter to be consistent with the other assets, which can be exported.

**Checklist**:

- [] Tests


swaterkamp:
The disabled checkbox is from some discussion but I can't find it again. I just read it, thought it was easy to implement, and did it.

The xml export is just in order to be consistent across other resources and especially assets. AFAIK we offer the xml export regardless of whether they can be imported again or not.

PR #1614, opened 2019-09-05
Add tests for tables and listpages for policies and audits

sarahd93:
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/audits/__tests__/listpage.js
+++ b/gsa/src/web/pages/audits/__tests__/listpage.js
@@ -139,10 +139,7 @@ describe('AuditPage ToolBarIcons test', () => {
     });
 
     const {element} = render(
-      <ToolBarIcons
-        entity={audit}
-        onAuditCreateClick={handleAuditCreateClick}
-      />,
+      <ToolBarIcons onAuditCreateClick={handleAuditCreateClick} />,
     );
     expect(element).toMatchSnapshot();
   });
@@ -156,10 +153,7 @@ describe('AuditPage ToolBarIcons test', () => {
     });
 
     const {getAllByTestId} = render(
-      <ToolBarIcons
-        entity={audit}
-        onAuditCreateClick={handleAuditCreateClick}
-      />,
+      <ToolBarIcons onAuditCreateClick={handleAuditCreateClick} />,
     );
 
     const icons = getAllByTestId('svg-icon');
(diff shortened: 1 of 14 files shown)


gsa/src/web/pages/audits/__tests__/table.js
```diff
@@ -0,0 +1,278 @@
...
+  first: 1,
+  all: 1,
+  filtered: 1,
+  length: 1,
+  rows: 2,
+});
+
+const filter = Filter.fromString('rows=2');
+
+describe('Table tests', () => {
```
swaterkamp: Could you name your tests a little more specific so that we can easily see which table is tested in comparison to the policy table for example. I'm thinking about simply naming this one "Audits table test". When yarn runs all the tests it's faster to see what went wrong

gsa/src/web/pages/policies/__tests__/table.js
```diff
@@ -0,0 +1,201 @@
...
+  first: 1,
+  all: 1,
+  filtered: 1,
+  length: 1,
+  rows: 2,
+});
+
+const filter = Filter.fromString('rows=2');
+
+describe('Table tests', () => {
```
swaterkamp: Same as above

gsa/src/web/pages/audits/__tests__/listpage.js
```diff
@@ -0,0 +1,187 @@
...
+    );
+
+    const icons = getAllByTestId('svg-icon');
+
+    fireEvent.click(icons[0]);
+    expect(handleAuditCreateClick).toHaveBeenCalled();
+    expect(icons[0]).toHaveAttribute('title', 'New Audit');
+  });
+
+  test('should not call click handlers without permission', () => {
```
swaterkamp: This is not really tested here. You are testing that an icon is not rendered when the permissions/caps are wrong. depending on the icon's behavior either the test or the title should be adjusted (in this case it's the title, I guess).

gsa/src/web/pages/policies/__tests__/listpage.js
```diff
@@ -0,0 +1,194 @@
...
+    fireEvent.click(icons[0]);
+    expect(handlePolicyCreateClick).toHaveBeenCalled();
+    expect(icons[0]).toHaveAttribute('title', 'New Policy');
+
+    fireEvent.click(icons[1]);
+    expect(handlePolicyImportClick).toHaveBeenCalled();
+    expect(icons[1]).toHaveAttribute('title', 'Import Policy');
+  });
+
+  test('should not call click handlers without permission', () => {
```
swaterkamp: Same as with audits

PR #1615, opened 2019-09-06
Added change methods to CheckboxComponent

saberlynx:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/components/form/checkbox.js
+++ b/gsa/src/web/components/form/checkbox.js
@@ -101,6 +101,6 @@ CheckboxComponent.propTypes = {
   onChange: PropTypes.func,
 };
 
-export default compose(withLayout())(CheckboxComponent);
+export default withLayout(CheckboxComponent);
 
 // vim: set ts=2 sw=2 tw=80:
--- a/gsa/src/web/components/form/checkbox.js
+++ b/gsa/src/web/components/form/checkbox.js
@@ -29,13 +29,15 @@ import withLayout from 'web/components/layout/withLayout';
 
 const convertChecked = (value, props) => {
   const {checkedValue, unCheckedValue} = props;
-
+  let val;
   if (value && isDefined(checkedValue)) {
-    value = checkedValue;
+    val = checkedValue;
   } else if (!value && isDefined(unCheckedValue)) {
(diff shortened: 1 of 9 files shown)


gsa/src/web/components/form/checkbox.js
```diff
@@ -49,19 +49,47 @@ const CheckboxComponent = ({
...
-        <StyledTitle data-testid="checkbox-title" disabled={disabled}>
-          {title}
-        </StyledTitle>
-      )}
-      {children}
-    </Divider>
-  </StyledElement>
-);
+}) => {
+  const [value, setValue] = useState(undefined);
```
bjoernricks: I guess it should be initialized with the value of a prop named checked (i am not sure if it must be named checked or value). E.g. `<Checkbox checked={true} ... />`

gsa/src/web/components/form/checkbox.js
```diff
@@ -73,12 +101,6 @@ CheckboxComponent.propTypes = {
...
 };
 
-export default compose(
-  withLayout(),
-  withChangeHandler({
-    convert_func: convert_checked,
-    value_func: event => event.target.checked,
-  }),
-)(CheckboxComponent);
+export default compose(withLayout())(CheckboxComponent);
```
bjoernricks: You can drop the compose function if we only have one HOC left.

```suggestion
export default withLayout(CheckboxComponent);
```

gsa/src/web/components/form/checkbox.js
```diff
@@ -49,19 +50,49 @@ const CheckboxComponent = ({
...
-        <StyledTitle data-testid="checkbox-title" disabled={disabled}>
-          {title}
-        </StyledTitle>
-      )}
-      {children}
-    </Divider>
-  </StyledElement>
-);
+}) => {
+  const [value, setValue] = useState(undefined);
```
bjoernricks: Either we need to initialize value with the value from the prop or we drop the internal state completely. Dropping the internal state would make the component controlled only but that would be fine for me.

saberlynx: > Dropping the internal state would make the component controlled only but that would be fine for me.

Sounds great to me!

PR #1617, opened 2019-09-09
Report list loading without details

bjoernricks:
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/commands/reports.js
+++ b/gsa/src/gmp/commands/reports.js
@@ -34,7 +34,7 @@ import EntityCommand from './entity';
 
 const log = logger.getLogger('gmp.commands.reports');
 
-class ReportsCommand extends EntitiesCommand {
+export class ReportsCommand extends EntitiesCommand {
   constructor(http) {
     super(http, 'report', Report);
   }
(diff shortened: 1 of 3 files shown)


gsa/src/gmp/commands/reports.js
```diff
@@ -59,6 +59,16 @@ class ReportsCommand extends EntitiesCommand {
       filter,
     });
   }
+
+  get(params, options) {
+    return super.get(
+      {
+        details: 0, // ensure to request no details by default
+        ...params,
+      },
```
swaterkamp: Changing the order of these two lines would actually ensure details: 0. In this order, details might be overwritten if they are set in params. I'm not 100% sure, wat we want here...

bjoernricks: the api may decide to request reports with details like `gmp.reports.get({details: 1, filter=...});`

swaterkamp: In that case it's the wrong order in gmp.results.get() ?

bjoernricks: yes you are right. gmp.results.get doesn't allow for not getting the details. IMHO it should be changed.

swaterkamp: Ok, would you like to put it here as well, or do we need another PR for that?

bjoernricks: Another PR because it isn't related

PR #1623, opened 2019-09-10
Added custom page title to all pages

saberlynx:
Added a pagetitle functional component, and put it in every single page. For ease of review, I will note that all I changed on each page is add a `<React.Fragment>` wrapper and then add `<PageTitle />` right below it. Then the `</React.Fragment>`
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/agents/detailspage.js
+++ b/gsa/src/web/pages/agents/detailspage.js
@@ -175,7 +175,7 @@ const Page = ({
       >
         {({activeTab = 0, onActivateTab}) => (
           <React.Fragment>
-            <PageTitle title={_(`Agent: ${entity.name}`)} />
+            <PageTitle title={_('Agent: {{name}}', {name: entity.name})} />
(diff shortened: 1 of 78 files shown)


gsa/src/web/pages/agents/detailspage.js
```diff
@@ -173,49 +174,52 @@ const Page = ({
...
-                <EntitiesTab entities={entity.userTags}>
-                  {_('User Tags')}
-                </EntitiesTab>
-                <EntitiesTab entities={permissions}>
-                  {_('Permissions')}
-                </EntitiesTab>
-              </TabList>
-            </TabLayout>
+          <React.Fragment>
+            <PageTitle title={_(`Agent: ${entity.name}`)} />
```
bjoernricks: To be translatable this needs to be
```suggestion
            <PageTitle title={_('Agent: {{name}}', {name: entity.name})} />
```

PR #1624, opened 2019-09-11
Scan config dialog improvements

bjoernricks:
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/scanconfig.js
+++ b/gsa/src/gmp/models/scanconfig.js
@@ -108,7 +108,7 @@ class ScanConfig extends Model {
       if (isDefined(ret.known_nvt_count)) {
         // number of known nvts by the scanner from last sync. should always be
         // equal or less then nvt_count because only the db may contain nvts not
-        // known nvts by the scanner e.g. a imported scan config contains
+        // known nvts by the scanner e.g. an imported scan config contains
         // private nvts
         ret.nvts.known = parseCount(ret.known_nvt_count);
         delete ret.known_nvt_count;
--- a/gsa/src/gmp/models/scanconfig.js
+++ b/gsa/src/gmp/models/scanconfig.js
@@ -98,6 +98,7 @@ class ScanConfig extends Model {
 
     if (isDefined(ret.nvt_count)) {
       ret.nvts = {
+        // number of selected nvts
         count: parseCount(ret.nvt_count.__text),
         trend: parseTrend(ret.nvt_count.growing),
       };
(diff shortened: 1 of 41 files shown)


gsa/src/gmp/models/scanconfig.js
```diff
@@ -98,19 +98,25 @@ class ScanConfig extends Model {
...
         trend: parseTrend(ret.nvt_count.growing),
       };
 
       delete ret.nvt_count;
 
       if (isDefined(ret.known_nvt_count)) {
-        ret.nvts.known = parse_count(ret.known_nvt_count);
+        // number of known nvts by the scanner from last sync. should always be
+        // equal or less then nvt_count because only the db may contain nvts not
+        // known nvts by the scanner e.g. a imported scan config contains
```
swaterkamp:
```suggestion
        // known nvts by the scanner e.g. an imported scan config contains
```

gsa/src/web/pages/scanconfigs/nvtpreferences.js
```diff
@@ -65,15 +65,17 @@ class NvtPreferenceDisplay extends React.Component {
   }
 }
 
-NvtPreferenceDisplay.propTypes = {
-  preference: PropTypes.shape({
-    nvt: PropTypes.shape({
-      oid: PropTypes.string.isRequired,
-      name: PropTypes.string.isRequired,
-    }).isRequired,
+export const NvtPreferencePropType = PropTypes.shape({
```
swaterkamp: I think this should be defined in utils/proptypes.js. It's rather special and not used anywhere else, but it would make looking up proptype information much easier if all the proptypes that we have are defined together (in a place where  we already have specific proptypes like error, capabilities, icon etc).

gsa/src/web/pages/scanconfigs/scannerpreferences.js
```diff
@@ -143,15 +143,15 @@ const ScannerPreferences = ({
   </Section>
 );
 
+export const ScannerPreferencePropType = PropTypes.shape({
```
swaterkamp: Same as with NvtPreferencePropType

gsa/src/web/pages/scanconfigs/nvtpreferences.js
bjoernricks: I am not sure about that. They aren't very generic and only used at these two places.

swaterkamp: I thought about that as well. But if that was completely true, there would be no need to export them (even if it was "only" for testing or so, this would make it more than one place, where it's used). The proptype iconSize, e.g. is also not very generic, despite being used in several places.

swaterkamp:
All in all this looks good to me. Just the failing tests need to be fixed. I don't know, if our CI can't handle .finally() or what else is wrong with it. As .then() is a working function, the failing code is not due to .get() being no Promise or so...

gsa/src/web/pages/scanconfigs/nvtpreferences.js
bjoernricks: Not generic did mean it is has a very special use case and will not be used outside of these both components.

bjoernricks: I've thought about your comment about the proptypes again and I don't think we should add them to the proptypes module because they are really specific to the used components in this PR. The preferences do contain more data then displayed and used in these components here. So using them at possible other places would require to extend them. But in that case they would also require more data then needed here and the test cased have to extended to avoid proptype warnings.

PR #1625, opened 2019-09-11
Added change method to Field component and refactored DatePicker into function

saberlynx:
**Checklist**:

- [] Tests


bjoernricks:
To fix the build error https://github.com/greenbone/gsa/blob/master/gsa/CMakeLists.txt#L278 need to be removed.

PR #1631, opened 2019-09-17
Gsa sets default filter

swaterkamp:
**Checklist**:

- [] Tests


gsa/src/web/store/usersettings/defaults/actions.js
```diff
@@ -58,4 +62,36 @@ export const loadUserSettingDefaults = gmp => () => (dispatch, getState) => {
...
+export const loadUserSettingDefault = gmp => id => (dispatch, getState) => {
+  const rootState = getState();
+  const selector = getUserSettingsDefaults(rootState);
+
+  if (selector.isLoading()) {
+    // we are already loading data
+    return Promise.resolve();
+  }
+
+  dispatch(loadingActions.request());
```
bjoernricks: Personally I wouldn't reuse the SETTINGS**S** actions for loading a single setting. But because updating the actions and reducers for isLoading and error creates a lot of headaches i would accept the current implementation.

But reducing a single settings as settings should override all other settings if i read the code correctly. Therefore this needs to be checked and should get tests!

gsa/src/web/pages/usersettings/usersettingspage.js
```diff
@@ -368,6 +368,48 @@ class UserSettings extends React.Component {
       autoCacheRebuild = {},
     } = this.props;
 
+    // TODO replace the following with nullish coalescent operator once available
+    agentsFilter = isNull(agentsFilter) ? {} : agentsFilter;
```
bjoernricks: You should drop the defaults above and use the following code instead. We should only set the expected value once.
```suggestion
    agentsFilter = hasValue(agentsFilter) ? {} : agentsFilter;
```

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,129 @@
...
+import {loadUserSettingsDefaultFilter} from 'web/store/usersettings/defaultfilters/actions';
+import {getUserSettingsDefaultFilter} from 'web/store/usersettings/defaultfilters/selectors';
+
+import PropTypes from 'web/utils/proptypes';
+import withGmp from 'web/utils/withGmp';
+
+const ROWS_PER_PAGE_SETTING_ID = '5f5a8712-8017-11e1-8556-406186ea4fc5';
+const DEFAULT_FALLBACK_FILTER = Filter.fromString('sort=name first=1');
+
+const FilterProvider = ({children, filter, gmp, gmpname, fallbackFilter}) => {
```
bjoernricks: I don't get the purpose of fallbackFilter. Either there is a filter set as fallback or not. Why do we need two variables?
Also the filter param (props.location.query.filter) case is missing here isn't it?

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,129 @@
...
+      dispatch(loadUserSettingDefault(gmp)(ROWS_PER_PAGE_SETTING_ID));
+    }
+  }, [retFilter, rowsPerPage, gmp, dispatch]);
+
+  const pageFilter = useSelector(state => {
+    const pFilter = getPage(state).filter;
+    return pFilter;
+  });
+
+  if (isDefined(filter)) {
```
bjoernricks: Sorry I also don't get your logic here. Maybe that's because I don't understand what filter should contain.

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,129 @@
...
+import {getUserSettingsDefaultFilter} from 'web/store/usersettings/defaultfilters/selectors';
+
+import PropTypes from 'web/utils/proptypes';
+import withGmp from 'web/utils/withGmp';
+
+const ROWS_PER_PAGE_SETTING_ID = '5f5a8712-8017-11e1-8556-406186ea4fc5';
+const DEFAULT_FALLBACK_FILTER = Filter.fromString('sort=name first=1');
+
+const FilterProvider = ({children, filter, gmp, gmpname, fallbackFilter}) => {
+  if (isString(filter)) {
```
bjoernricks: I don't like checking for a string here? Is this really necessary?

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,129 @@
...
+    filter = Filter.fromString(filter);
+  }
+  if (isString(fallbackFilter)) {
+    fallbackFilter = Filter.fromString(fallbackFilter);
```
bjoernricks: Same as above? Couldn't we expect a Filter model instance here?

swaterkamp: changed

gsa/src/web/pages/usersettings/usersettingspage.js
swaterkamp: This is not equivalent. hasValue() explicitly checks for the variable to be NOT null

bjoernricks: Sorry should be
```javascript
 agentsFilter = hasValue(agentsFilter) ? agentsFilter : {};
```
of course.

gsa/src/web/store/utils.js
```diff
@@ -27,8 +27,14 @@ import {isDefined} from 'gmp/utils/identity';
...
  */
-export const filterIdentifier = filter =>
-  isDefined(filter) ? `filter:${filter.toFilterString()}` : 'default';
+export const filterIdentifier = filter => {
+  if (filter === null) {
+    return `filter:null`;
+  } else if (isDefined(filter)) {
+    return `filter:${filter.toFilterString()}`;
+  }
+  return 'default';
```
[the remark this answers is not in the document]
swaterkamp: filter === null case removed and merged with isDefined()
filter === null can and will happen
filter === null was introduced as a means of telling that the filter was tried to be loaded but doesn't exist

As the backend default should not happen anymore with the new FilterProvider, I propose to use the following:
```
export const filterIdentifier = filter => {
  if (isDefined(filter) && !isNull(filter)) {
    return `filter:${filter.toFilterString()}`;
  }
};
```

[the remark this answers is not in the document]
swaterkamp: Or in a simplified way:
```
export const filterIdentifier = filter => isDefined(filter) && !isNull(filter) ? `filter:${filter.toFilterString()}` : undefined;
```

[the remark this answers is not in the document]
swaterkamp: This fails at least 116 tests, though...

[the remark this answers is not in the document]
swaterkamp: For now it is:
```
export const filterIdentifier = filter => hasValue(filter) ? `filter:${filter.toFilterString()}` : 'default';
```

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,121 @@
...
+      {showChildren ? children({filter: returnedFilter}) : <Loading />}
+    </React.Fragment>
+  );
+};
+
+FilterProvider.propTypes = {
+  fallbackFilter: PropTypes.filter,
+  gmp: PropTypes.gmp.isRequired,
+  gmpname: PropTypes.string,
+  locationQuery: PropTypes.object,
```
bjoernricks:
```suggestion
  locationQuery: PropTypes.shape({
      filter: PropTypes.string.isRequired,
  }),
```

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+    const {baseElement, getByTestId} = render(
+      <FilterProvider locationQuery={locationQuery} gmpname="task">
+        {renderFunc}
+      </FilterProvider>,
+    );
+
+    await waitForElement(() => getByTestId('awaiting-span'));
+
+    expect(renderFunc).toHaveBeenCalledWith({filter: resultingFilter});
+    expect(baseElement).toMatchSnapshot();
```
bjoernricks: I don't see any value in the snapshots. If you are looking at the snap file it always contains the same html elements for all tests...

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+import {loadingActions} from 'web/store/usersettings/defaults/actions';
+import {defaultFilterLoadingActions} from 'web/store/usersettings/defaultfilters/actions';
+import {pageFilter} from 'web/store/pages/actions';
+
+import {rendererWith, waitForElement} from 'web/utils/testing';
+
+import FilterProvider, {DEFAULT_FALLBACK_FILTER} from '../filterprovider';
+
+describe('FilterProvider component tests', () => {
+  test('should return locationQuery filter', async () => {
```
bjoernricks:
```suggestion
  test('should prefer locationQuery filter over default setting filter', async () => {
```

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+      </FilterProvider>,
+    );
+
+    await waitForElement(() => getByTestId('awaiting-span'));
+
+    expect(renderFunc).toHaveBeenCalledWith({filter: resultingFilter});
+    expect(baseElement).toMatchSnapshot();
+  });
+
+  test('should return pageFilter', async () => {
```
bjoernricks:
```suggestion
  test('should prefer pageFilter over default setting filter', async () => {
```

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+      <FilterProvider gmpname="task">{renderFunc}</FilterProvider>,
+    );
+
+    await waitForElement(() => getByTestId('awaiting-span'));
+
+    expect(renderFunc).toHaveBeenCalledWith({filter: resultingFilter});
+    expect(baseElement).toMatchSnapshot();
+  });
+
+  test('should return defaultSettingFilter', async () => {
```
bjoernricks:
```suggestion
  test('should use defaultSettingFilter', async () => {
```

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+    // if no pageFilter exists use the user default from the usersettings
+    const resultingFilter = Filter.fromString('foo=bar rows=42');
```
bjoernricks: Is it correct to merge the default rows setting into the default filter? How is it handled in gvmd currently?

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+      <FilterProvider gmpname="task">{renderFunc}</FilterProvider>,
+    );
+
+    await waitForElement(() => getByTestId('awaiting-span'));
+
+    expect(renderFunc).toHaveBeenCalledWith({filter: resultingFilter});
+    expect(baseElement).toMatchSnapshot();
+  });
+
+  test('should return fallbackFilter if given', async () => {
```
bjoernricks:
```suggestion
  test('should use fallbackFilter if given', async () => {
```

gsa/src/web/entities/__tests__/filterprovider.js
```diff
@@ -0,0 +1,257 @@
...
+      </FilterProvider>,
+    );
+
+    await waitForElement(() => getByTestId('awaiting-span'));
+
+    expect(renderFunc).toHaveBeenCalledWith({filter: resultingFilter});
+    expect(baseElement).toMatchSnapshot();
+  });
+
+  test('should return default fallbackFilter', async () => {
```
bjoernricks:
```suggestion
  test('should use default fallbackFilter as last resort', async () => {
```

gsa/src/web/entities/__tests__/filterprovider.js
bjoernricks: Ah I see it is only merged if rows is not set in the filter. Perfect :+1:

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+import getPage from 'web/store/pages/selectors';
+import {loadUserSettingDefault} from 'web/store/usersettings/defaults/actions';
+import {getUserSettingsDefaults} from 'web/store/usersettings/defaults/selectors';
+import {loadUserSettingsDefaultFilter} from 'web/store/usersettings/defaultfilters/actions';
+import {getUserSettingsDefaultFilter} from 'web/store/usersettings/defaultfilters/selectors';
+
+import PropTypes from 'web/utils/proptypes';
+import withGmp from 'web/utils/withGmp';
+
+const ROWS_PER_PAGE_SETTING_ID = '5f5a8712-8017-11e1-8556-406186ea4fc5';
```
bjoernricks: Shouldn't this ID be put into https://github.com/greenbone/gsa/blob/master/gsa/src/gmp/commands/users.js ?

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+export const DEFAULT_FALLBACK_FILTER = Filter.fromString('sort=name first=1');
```
bjoernricks: Maybe it belongs to the end of https://github.com/greenbone/gsa/blob/master/gsa/src/gmp/models/filter.js

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+
+  let returnedFilter;
+
+  let defaultSettingFilter = useSelector(state =>
+    getUserSettingsDefaultFilter(state, gmpname).getFilter(),
+  );
+
+  useEffect(() => {
+    if (!isDefined(defaultSettingFilter)) {
+      dispatch(loadUserSettingsDefaultFilter(gmp)(gmpname));
```
bjoernricks: I would love to replace this with something like `loadUserSettingsDefaultFilterIfNeeded` which checks the state itself if the setting has already been loaded. But this is for another PR.

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }, [defaultSettingFilter, dispatch, gmp, gmpname]);
+
+  const rowsPerPage = useSelector(state => {
```
bjoernricks: Can be simplified
```suggestion
  const rowsPerPage = useSelector(state => getUserSettingsDefaults(state).getValueByName('rowsperpage'));
```

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+    return rows;
+  });
+
+  useEffect(() => {
+    if (!isDefined(rowsPerPage)) {
+      dispatch(loadUserSettingDefault(gmp)(ROWS_PER_PAGE_SETTING_ID));
+    }
+  }, [returnedFilter, rowsPerPage, gmp, dispatch]);
+
+  const pageFilter = useSelector(state => {
```
bjoernricks: Can be simplified
```suggestion
  const pageFilter = useSelector(state => getPage(state).getFilter(gmpname));
```

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+  } else if (isDefined(defaultSettingFilter) && defaultSettingFilter !== null) {
+    returnedFilter = defaultSettingFilter;
+  } else if (isDefined(fallbackFilter)) {
+    returnedFilter = fallbackFilter;
+  } else {
+    returnedFilter = DEFAULT_FALLBACK_FILTER;
+  }
+
+  // TODO remove the following line once the resultsfilter is working correctly
+  if (gmpname === 'result') {
```
bjoernricks: I don't get why this is necessary at the moment. Could you clarify this?

gsa/src/web/entities/filterprovider.js
```diff
@@ -0,0 +1,130 @@
...
+  const showChildren =
+    isDefined(returnedFilter) &&
+    isDefined(defaultSettingFilter) &&
+    isDefined(rowsPerPage);
+
+  return (
+    <React.Fragment>
+      {showChildren ? (
+        <React.Fragment>
+          <span data-testid="awaiting-span" />
```
bjoernricks: Should be dropped and only be rendered in the tests.
```javascript
const renderFunc = jest.fn().mockReturnValue(<span data-testid="awaiting-span" />);
```
```suggestion
```

gsa/src/web/pages/policies/__tests__/listpage.js
```diff
@@ -62,6 +62,12 @@ const currentSettings = jest.fn().mockReturnValue(
   }),
 );
 
+const getSetting = jest.fn().mockReturnValue(
```
bjoernricks: Btw. this can be simplified by using jest.fn().mockResolvedValue https://jestjs.io/docs/en/mock-function-api#mockfnmockresolvedvaluevalue

gsa/src/web/entities/__tests__/filterprovider.js
swaterkamp: I'll leave one snapshot test, just in case.

gsa/src/web/entities/filterprovider.js
swaterkamp: If the results default filter is loaded, it uses '0' as filter id which results in an error. As you told me that you are working on the results filters right now and I should not change anything about them, I just made sure that GSA still renders results.

gsa/src/web/entities/filterprovider.js
swaterkamp: But it's an internal thing in FilterProvider. If I remove it there and put it into the test as a child or so, the test doesn't work anymore.

swaterkamp:
Known issue which needs a fix: If a locationQuery is used and the filter is changed by the user, it does not change the URL, so at the next rerender the query from the URL will be used again instead of the user's filter.

gsa/src/web/entities/filterprovider.js
bjoernricks: Oh i guess we misunderstood. I don't work on the result filters directly. I work on the report details page and therefore on the report results too. Why is it returning 0 at all and what's the meaning of 0? Is 0 == no filter?

PR #1640, opened 2019-09-23
send details=1 for get_report

sarahd93:
Adjust gsa to send details=1 for get_report and change gsad to forward details to gvmd.

**Checklist**:

- [] Tests


swaterkamp:
We should change the type to whatever is used in the other "Details" params then.

sarahd93:
I looked into the gsad code again and all the other times details params are used it is together with the gmp_arguments_t. I am not comfortable working with that, though, I feel like any further changes I could try to make now would involve too much guessing on my part, because I do not know enough about C. 
Should we just merge the PR like it is for now so at least the report details page will work again?

bjoernricks:
I am fine with merging it

PR #1646, opened 2019-09-26
Add OSP Sensor type to GSA

saberlynx:
**Checklist**:
Changes:
- Added OSP Sensor Scanner Type. It is now Type 5, and updated scannerTypeName function.
- Updated tests, to test for this new name and cloneability/writeability
- Add OSP Sensor Type to Scanner Dialog. credential_id = '' if a scanner is of this type.
- OSP Sensor Type is now in Task Dialog, with the same options as GMP Scanners
- Made OSP Sensors selectable on performance page.
- Modified validator of credential_id to allow empty string.

- [] Tests


bjoernricks:
It already works as expected :+1:

gsa/src/web/pages/performance/performancepage.js
```diff
@@ -162,7 +162,9 @@ const Selector = withClickHandler()(styled.span`
   }}
 `);
 
-const SLAVE_SCANNER_FILTER = Filter.fromString('type=' + GMP_SCANNER_TYPE);
+const SLAVE_SCANNER_FILTER = Filter.fromString(
```
bjoernricks: Would be nice if we could rename this variable in another PR. Something like SENSOR_SCANNERS_FILTER

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -182,8 +186,16 @@ class ScannerDialog extends React.Component {
...
-    const isGmpScannerType = type === GMP_SCANNER_TYPE;
+      scanner.credential.credential_type === CLIENT_CERTIFICATE_CREDENTIAL_TYPE;
+
+    const isOspSensorType = type === OSP_SENSOR_SCANNER_TYPE;
+    const isOspScannerType = type === OSP_SCANNER_TYPE;
+
+    let {credential_id} = this.props;
+
+    if (isOspSensorType) {
+      credential_id = '';
```
bjoernricks: @saberlynx could you check if undefined fits now with Timos patch https://github.com/greenbone/gvmd/pull/759 too? Personally I don't want to send special values like '0' or '' for "no value". Instead the API should just get not value.

saberlynx: Do you mean we should be able to do something like credential_id = undefined or credential_id = null? I was able to make it work with '0', but not '', null or undefined.

saberlynx: Nevermind. I misunderstood. I got rid of line 196-198 entirely and it works great.

saberlynx: Test results:
credential_id = null => Given credential_id was invalid
crednetial_id = '' => Given credential_id was invalid
credential_id = undefined => Given credential_id was invalid
credential_id = '0' => successfully saved without a credential

getting rid of lines 196-198 entirely => scanner will save with default 'cc' credential.

bjoernricks: Ah ok I guess now gsad's validation must be adjusted to allow an empty credential_id for this scanner type.

PR #1649, opened 2019-09-30
Extracted error.message from error object for dashboard

saberlynx:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/components/dashboard/dashboard.js
+++ b/gsa/src/web/components/dashboard/dashboard.js
@@ -261,7 +261,7 @@ export class Dashboard extends React.Component {
     if (isDefined(error) && !isLoading) {
       return (
         <RowPlaceHolder>
-          {_(`Could not load dashboard settings. Reason: ${error.message}`)}
+          {_(`Could not load dashboard settings. Reason: {{error}}`, {error: error.message})}
         </RowPlaceHolder>
       );
     } else if (!isDefined(rows) && isLoading) {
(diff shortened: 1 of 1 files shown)


gsa/src/web/components/dashboard/dashboard.js
```diff
@@ -255,12 +255,13 @@ export class Dashboard extends React.Component {
...
       ...props
     } = this.props;
+
     const rows = this.getRows();
 
     if (isDefined(error) && !isLoading) {
       return (
         <RowPlaceHolder>
-          {_('Could not load dashboard settings. Reason: {{error}}', {error})}
+          {_(`Could not load dashboard settings. Reason: ${error.message}`)}
```
bjoernricks: I don't think your change it translatable. Therefore it should be
```suggestion
          {_(`Could not load dashboard settings. Reason: {{error}}`, {error: error.message})}
```

gsa/src/web/components/dashboard/dashboard.js
```diff
@@ -255,12 +255,13 @@ export class Dashboard extends React.Component {
...
       ...props
     } = this.props;
+
     const rows = this.getRows();
 
     if (isDefined(error) && !isLoading) {
       return (
         <RowPlaceHolder>
-          {_('Could not load dashboard settings. Reason: {{error}}', {error})}
+          {_(`Could not load dashboard settings. Reason: {{error}}`, {error: error.message})}
```
bjoernricks:
```suggestion
          {_('Could not load dashboard settings. Reason: {{error}}', {error: error.message})}
```

PR #1653, opened 2019-10-01
Filterprovider improvements

bjoernricks:
Handle errors safely

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/filter.js
+++ b/gsa/src/gmp/models/filter.js
@@ -759,6 +759,8 @@ export const DEFAULT_FALLBACK_FILTER = Filter.fromString('sort=name first=1');
 
 export const RESET_FILTER = Filter.fromString('first=1');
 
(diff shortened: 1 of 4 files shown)


gsa/src/gmp/models/filter.js
```diff
@@ -759,6 +759,8 @@ export const DEFAULT_FALLBACK_FILTER = Filter.fromString('sort=name first=1');
 
 export const RESET_FILTER = Filter.fromString('first=1');
 
+export const DEFAULT_ROWS_PER_PAGE = 10;
```
swaterkamp: As per request of PS, this should be much higher. About 100 rows should fit user requirements better.
```suggestion
export const DEFAULT_ROWS_PER_PAGE = 100;
```

bjoernricks: As discussed the fallback will be 50 now.

PR #1671, opened 2019-10-08
Decide whether to default to full and fast scan config (task dialog)

saberlynx:
Summary of changes:
- moved handleScannerChange out of ScannerSelect
- Refactored ScannerSelect into function
- Added state variables configType and prevConfigType, and calculate/update them in handleScannerChange, and upon onScanConfigChange.

Expected behavior:
- Changing between scanners of the same config type should not change displayed scan config selection
- Edit dialog should show correct scan config
- Scan config should default to "Full and Fast" when 1. creating a new task 2. changing configType from something else TO openvas config type.

**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/tasks/dialog.js
+++ b/gsa/src/web/pages/tasks/dialog.js
@@ -17,7 +17,7 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import React from 'react';
+import React, {useState} from 'react';
 
 import _ from 'gmp/locale';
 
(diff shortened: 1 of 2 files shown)


swaterkamp:
I was thinking to use the task component's state to control the dialog content (scanner and scan config types), but I think I haven't been clear about that, sorry.

gsa/src/web/pages/tasks/dialog.js
```diff
@@ -225,6 +181,57 @@ const TaskDialog = ({
...
+      scanner_type === GMP_SCANNER_TYPE
+    ) {
+      setConfigType('openvas');
+    } else if (scanner_type === OSP_SCANNER_TYPE) {
+      setConfigType('osp');
+    } else {
+      setConfigType('other');
+    }
+
+    log.debug('on scanner change', value, scanner);
```
bjoernricks: Should be removed. I've added it because the whole scanner and config changes are very complicated as you also noticed while working on this PR ;-)
```suggestion
```

PR #1673, opened 2019-10-08
Fix report cves

bjoernricks:
**Checklist**:

- [x] Tests


gsa/src/web/pages/reports/cvestable.js
```diff
@@ -89,6 +103,11 @@ const Row = ({entity}) => {
           ))}
         </Divider>
       </TableData>
+      <TableData>
+        <DetailsLink type="nvt" id={id} title={nvtName}>
+          {shorten(nvtName, 80)}
```
swaterkamp: What happens if the windows gets smaller? Would it be necessary to add line-breaking here, although it might look ok in "full-screen"?

bjoernricks: Shouldn't it break automatically? IMHO it only doesn't break if we use css to forbid that.

PR #1682, opened 2019-10-10
Remount ErrorBoundary on path change

saberlynx:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/page.js
+++ b/gsa/src/web/pages/page.js
@@ -19,6 +19,8 @@
 
 import React from 'react';
 
+import {withRouter} from 'react-router-dom';
+
 import styled from 'styled-components';
 
 import _ from 'gmp/locale';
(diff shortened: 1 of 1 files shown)


gsa/src/web/pages/page.js
```diff
@@ -43,6 +44,7 @@ import CapabilitiesContext from 'web/components/provider/capabilitiesprovider';
 import Footer from 'web/components/structure/footer';
 import Header from 'web/components/structure/header';
 import Main from 'web/components/structure/main';
+import withRouter from 'react-router-dom/withRouter';
```
bjoernricks: Could you move this import to the top after the react import and change it to the following?
```suggestion
import {withRouter} from 'react-router-dom';
```

bjoernricks: Just the part about resorting the import is missing :-)
> Could you move this import to the top after the react import

PR #1685, opened 2019-10-10
Drop "All Secinfo" section

janowagner:
This removes the menu item SecInfo->All SecInfo and all related code elements.

**Checklist**:

- [ ] Tests N/A


bjoernricks:
Looks very good already :-) 

`gsa/src/web/components/icon/__tests__/allinfoicon.js`can be removed too

PR #1704, opened 2019-10-21
Fix TLS certificate download

swaterkamp:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/tlscertificates/component.js
+++ b/gsa/src/web/pages/tlscertificates/component.js
@@ -56,7 +56,7 @@ class TlsCertificateComponent extends React.Component {
   handleTlsCertificateDownload(cert) {
     const {detailsExportFileName, gmp, username, onDownloaded} = this.props;
 
-    gmp.tlscertificate.get({id: cert.id}).then(response => {
+    return gmp.tlscertificate.get({id: cert.id}).then(response => {
       const {data} = response;
 
       const {
--- a/gsa/src/web/pages/tlscertificates/component.js
+++ b/gsa/src/web/pages/tlscertificates/component.js
@@ -81,7 +81,7 @@ class TlsCertificateComponent extends React.Component {
         username,
       });
 
-      onDownloaded({
+      return onDownloaded({
         filename,
         mimetype: 'application/x-x509-ca-cert',
         data: create_pem_certificate(certificate),
--- a/gsa/src/web/pages/tlscertificates/component.js
+++ b/gsa/src/web/pages/tlscertificates/component.js
@@ -25,7 +25,6 @@ import {isDefined} from 'gmp/utils/identity';
 
 import EntityComponent from 'web/entity/component';
 
-import {loadEntity, selector} from 'web/store/entities/tlscertificates';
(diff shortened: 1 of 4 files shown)


gsa/src/web/pages/tlscertificates/component.js
```diff
@@ -142,6 +144,8 @@ const mapStateToProps = (rootState, {id}) => {
   const detailsExportFileName = userDefaultsSelector.getValueByName(
     'detailsexportfilename',
   );
+  // console.log('Selector', tlsCertificateSelector, id);
```
bjoernricks:
```suggestion
```

gsa/src/web/pages/tlscertificates/component.js
```diff
@@ -142,6 +144,8 @@ const mapStateToProps = (rootState, {id}) => {
...
+  // console.log('state2props', tlsCertificateSelector.getEntity(id));
```
bjoernricks:
```suggestion
```

gsa/src/web/pages/tlscertificates/component.js
```diff
@@ -57,34 +57,36 @@ class TlsCertificateComponent extends React.Component {
...
-      resourceName: name,
-      resourceType: entityType,
-      username,
-    });
-
-    onDownloaded({
-      filename,
-      mimetype: 'application/x-x509-ca-cert',
-      data: create_pem_certificate(certificate),
+    this.props.loadEntity(cert.id).then(response => {
```
bjoernricks: Does not make much sense to use the action creator function here to load the tls cert. It even would crash if a second request is already running.

gsa/src/web/pages/tlscertificates/component.js
```diff
@@ -55,36 +54,38 @@ class TlsCertificateComponent extends React.Component {
...
-      username,
-    });
-
-    onDownloaded({
-      filename,
-      mimetype: 'application/x-x509-ca-cert',
-      data: create_pem_certificate(certificate),
+    const {detailsExportFileName, gmp, username, onDownloaded} = this.props;
+
+    gmp.tlscertificate.get({id: cert.id}).then(response => {
```
bjoernricks:
```suggestion
    return gmp.tlscertificate.get({id: cert.id}).then(response => {
```

gsa/src/web/pages/tlscertificates/component.js
```diff
@@ -55,36 +54,38 @@ class TlsCertificateComponent extends React.Component {
...
+        extension: 'pem', // this gets overwritten to .cer in chrome
+        fileNameFormat: detailsExportFileName,
+        id,
+        modificationTime,
+        resourceName: name,
+        resourceType: entityType,
+        username,
+      });
+
+      onDownloaded({
```
bjoernricks:
```suggestion
      return onDownloaded({
```

gsa/src/web/pages/tlscertificates/component.js
swaterkamp: Indeed, I forgot about that!

PR #1719, opened 2019-10-24
if details not defined, then choose first OS from identifiers

saberlynx:
When the host details are defined, it should still work as before.
**Checklist**:

- [] Tests


gsa/src/web/pages/hosts/row.js
```diff
@@ -112,13 +112,26 @@ const Row = ({
...
+  );
+
+  if (isDefined(details) && isDefined(details.best_os_cpe)) {
+    os_cpe = details.best_os_cpe.value;
+  } else if (isDefined(identifiers) && isDefined(identifiers[os_index])) {
+    os_cpe = identifiers[os_index].value;
+  } else {
+    os_cpe = undefined;
+  }
+
```
bjoernricks: Could be simplified
```javascript
  if (isDefined(details) && isDefined(details.best_os_cpe)) {
    os_cpe = details.best_os_cpe.value;
  } else if (isDefined(identifiers)) {
    os_cpe = identifiers.find(
      identifier => identifier.name === 'OS',
    );
  } 
```
find returns undefined if no value matches.

Please ensure when using find or findIndex that the core-js import is used in the module to support IE 11. Either include

```javascript
import 'core-js/features/array/find-index';
```
or
```javascript
import 'core-js/features/array/find';
```

bjoernricks:
Oh and it would be nice to write a test for this solution

gsa/src/web/pages/hosts/row.js
saberlynx: Done! i put the array.find in a separate constant and checked if its defined, because we need to access the value of that object and it's not always loaded immediately. I wanted to avoid an error.

PR #1725, opened 2019-10-28
Multistep dialog feature, implemented on scanner dialog

saberlynx:
Summary of Changes: 
- Added a MultiStepFooter with previous/next buttons
- Changed SaveDialog to be able to use MultiStepFooter. multiStep is by default false. Added step change method.
- Changed ScannerDialog to multiStep = true, and put form fields in different tabs.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/components/dialog/__tests__/multistepfooter.js
+++ b/gsa/src/web/components/dialog/__tests__/multistepfooter.js
@@ -1,0 +1,158 @@
+/* Copyright (C) 2019 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+import React from 'react';
+
+import {setLocale} from 'gmp/locale/lang';
+
+import {render, fireEvent} from 'web/utils/testing';
+import Theme from 'web/utils/theme';
+
+import MultiStepFooter from 'web/components/dialog/multistepfooter';
+
+setLocale('en');
+
+describe('MultiStepFooter tests', () => {
+  test('should render', () => {
+    const {element} = render(<MultiStepFooter rightButtonTitle="Foo" />);
+
+    expect(element).toMatchSnapshot();
+
+    const button = element.querySelector('button[title="Foo"]');
+    const prevButton = element.querySelector('button[title="🠬"]');
+    const nextButton = element.querySelector('button[title="🠮"]');
+
+    expect(button).toHaveStyleRule('background', Theme.lightGreen);
+    expect(prevButton).toHaveStyleRule('background', Theme.white);
+    expect(nextButton).toHaveStyleRule('background', Theme.white);
+  });
+
+  test('should render loading and disable cancel button', () => {
+    const {element} = render(
+      <MultiStepFooter rightButtonTitle="Foo" loading={true} />,
+    );
+
+    expect(element).toMatchSnapshot();
+
+    const button = element.querySelector('button[title="Foo"]');
+    const buttonLeft = element.querySelector('button[title="Cancel"]');
+
+    expect(button).toHaveStyleRule(
+      'background',
+      `${Theme.lightGreen} url(/img/loading.gif) center center no-repeat`,
+    );
+    expect(buttonLeft).toHaveAttribute('disabled');
+  });
+
+  test('should render footer with default titles', () => {
+    const {element} = render(<MultiStepFooter rightButtonTitle="Foo" />);
+
+    const buttons = element.querySelectorAll('button');
+
+    expect(buttons).toHaveLength(4);
+
+    expect(buttons[0]).toHaveAttribute('title', 'Cancel');
+    expect(buttons[0]).toHaveTextContent('Cancel');
+    expect(buttons[1]).toHaveAttribute('title', '🠬');
+    expect(buttons[1]).toHaveTextContent('🠬');
+    expect(buttons[2]).toHaveAttribute('title', '🠮');
+    expect(buttons[2]).toHaveTextContent('🠮');
+    expect(buttons[3]).toHaveAttribute('title', 'Foo');
+    expect(buttons[3]).toHaveTextContent('Foo');
+  });
+
+  test('should render footer with custom titles', () => {
+    const {element} = render(
+      <MultiStepFooter
+        rightButtonTitle="Foo"
+        leftButtonTitle="Bar"
+        previousButtonTitle="Back"
+        nextButtonTitle="Forward"
+      />,
+    );
+
+    const buttons = element.querySelectorAll('button');
+
+    expect(buttons).toHaveLength(4);
+
+    expect(buttons[0]).toHaveAttribute('title', 'Bar');
+    expect(buttons[0]).toHaveTextContent('Bar');
+    expect(buttons[1]).toHaveAttribute('title', 'Back');
+    expect(buttons[1]).toHaveTextContent('Back');
+    expect(buttons[2]).toHaveAttribute('title', 'Forward');
+    expect(buttons[2]).toHaveTextContent('Forward');
+    expect(buttons[3]).toHaveAttribute('title', 'Foo');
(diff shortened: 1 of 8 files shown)


gsa/src/web/components/dialog/multistepfooter.js
```diff
@@ -0,0 +1,106 @@
...
+
+import Divider from 'web/components/layout/divider';
+
+const StyledLayout = styled(DialogFooterLayout)`
+  justify-content: space-between;
+`;
+
+const MultiStepFooter = ({
+  leftButtonTitle = _('Cancel'),
+  nextButtonTitle = _('🠮'),
```
swaterkamp: I don't think this needs to be translatable ;)

gsa/src/web/components/dialog/__tests__/multistepfooter.js
```diff
@@ -0,0 +1,180 @@
+/* Copyright (C) 2018 Greenbone Networks GmbH
```
swaterkamp:
```suggestion
/* Copyright (C) 2019 Greenbone Networks GmbH
```

gsa/src/web/components/dialog/__tests__/multistepfooter.js
```diff
@@ -0,0 +1,180 @@
...
+  });
+
+  test('should render loading and disable cancel button', () => {
+    const {element} = render(
+      <MultiStepFooter rightButtonTitle="Foo" loading={true} />,
+    );
+
+    expect(element).toMatchSnapshot();
+
+    const button = element.querySelector('button[title="Foo"]');
```
swaterkamp: When using English strings it might be necessary to set the locale to 'en' to make sure the test will pass on all machines independent of the language setting

gsa/src/web/components/dialog/multistepfooter.js
```diff
@@ -0,0 +1,106 @@
+/* Copyright (C) 2018-2019 Greenbone Networks GmbH
```
swaterkamp:
```suggestion
/* Copyright (C) 2019 Greenbone Networks GmbH
```

gsa/src/web/components/dialog/savedialog.js
```diff
@@ -146,14 +170,18 @@ const SaveDialog = ({
...
   title,
   values,
   width,
   onClose,
   onError,
   onErrorClose,
   onSave,
+  ...props
 }) => {
+  const {nextDisabled, onNext, onPrev, prevDisabled} = props;
```
swaterkamp: This is equivalent to just putting these props into the block, where you added the ...props. Please just extend the list above instead of making the detour with this line.

gsa/src/web/components/hooks/usepageturn.js
```diff
@@ -0,0 +1,61 @@
+/* Copyright (C) 2018-2019 Greenbone Networks GmbH
```
swaterkamp:
```suggestion
/* Copyright (C) 2019 Greenbone Networks GmbH
```

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -140,211 +165,233 @@ class ScannerDialog extends React.Component {
...
+    scanner.credential.credential_type === CLIENT_CERTIFICATE_CREDENTIAL_TYPE;
+
+  const isGreenboneSensorType = type === GREENBONE_SENSOR_SCANNER_TYPE;
+  const isOspScannerType = type === OSP_SCANNER_TYPE;
+
+  if (isGreenboneSensorType) {
+    credential_id = '';
+  }
+
+  const [activeTab, setActiveTab] = useState(0);
```
swaterkamp: It would be nice to have the hooks declared at the beginning of the function component to be able to quickly find how variables are defined. imho useState, usePageTurn and so on should be collectively used right after }) => { in line 155.

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -140,211 +165,233 @@ class ScannerDialog extends React.Component {
...
+  const totalPages = 2;
+
+  const {page, handlePageTurn, prevDisabled, nextDisabled} = usePageTurn({
+    activeTab,
+    lastPage: totalPages - 1,
+  });
+
+  useEffect(() => {
+    setActiveTab(page);
+  }, [page]);
```
swaterkamp: Is there a fundamental difference between setting the active tab here and setting the page in usePageTurn()? It seems that we are syncing between two states here, which might lead to trouble if that fails. It would be nicer to have just one place in the code that defines on which page we are.

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -140,211 +165,233 @@ class ScannerDialog extends React.Component {
...
+
+  return (
+    <SaveDialog
+      title={title}
+      onClose={onClose}
+      onSave={onSave}
+      multiStep={true} // the option to switch to the MultiStepFooter. Leaving it out defaults to false
+      onNext={() => handlePageTurn('Next')} // 'Next' turns the next page
```
swaterkamp: Maybe you could define a constant like PAGETURN_DIRECTION_NEXT = 'next' and PAGETUNR_DIRECTION_PREVIOUS='previous' or so that gets exported from the usePageTurn() hook. By this we can easily change that value in just one place and don't have to bother about whether the value is case sensitive or not. It just makes things less errorprone.

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -140,211 +165,233 @@ class ScannerDialog extends React.Component {
...
+    activeTab,
+    lastPage: totalPages - 1,
+  });
+
+  useEffect(() => {
+    setActiveTab(page);
+  }, [page]);
+
+  return (
+    <SaveDialog
```
swaterkamp: A small nitpick: Could you sort the props by alphabet and put the on... handlers after every other prop?

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -140,211 +165,233 @@ class ScannerDialog extends React.Component {
...
+      {({values: state, onValueChange}) => {
+        return (
+          <Layout flex="column">
+            <DialogTabLayout grow="1" align={['start', 'end']}>
+              <TabList
+                active={activeTab}
+                align={['start', 'stretch']}
+                onActivateTab={handleActivateTab}
+              >
+                <Tab style={{backgroundColor: 'red'}}>{_('Metadata')}</Tab>
```
swaterkamp: Please use styled-components to style components instead of the style attribute. Also, I don't get, why this needs to be red? Also also: if possible, please use colors from our theme.js instead of defining new ones.

swaterkamp: We should also discuss how to call this basic tab. It will probably be part of most multi-step dialogs that are to come, and it should have a proper name. I don't think that 'Metadata' is an adequate title for all expected use cases.

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -140,211 +165,233 @@ class ScannerDialog extends React.Component {
...
+            </DialogTabLayout>
+            <Tabs active={activeTab}>
+              <TabPanels>
+                <TabPanel>
+                  <FormGroup title={_('Name')}>
+                    <TextField
+                      name="name"
+                      grow="1"
+                      value={state.name}
+                      size="30"
```
swaterkamp: Could you check if setting the size here (and in the next TextField) is senseful? I don't think we are using this anymore, especially because grow="1" should render the Field in the whole available space.

gsa/src/web/pages/scanners/dialog.js
saberlynx: you're right. this is unnecessarily convoluted. I got rid of it upon edit.

gsa/src/web/pages/scanners/dialog.js
saberlynx: This is code I left in by mistake. I will get rid of it.

gsa/src/web/components/dialog/__tests__/multistepfooter.js
saberlynx: Done :)

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -132,6 +132,8 @@ CertStatus.propTypes = {
   info: PropTypes.object.isRequired,
 };
 
+const SCANNER_DIALOG_MAX_HEIGHT = '300px';
```
swaterkamp: If this deals with the initial height, please name it SCANNER_DIALOG_INITIAL_HEIGHT. From MAX_HEIGHT I'd expect that the dialog can not be enlarged.

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -120,16 +132,29 @@ CertStatus.propTypes = {
...
-  constructor(...args) {
-    super(...args);
-
-    this.handleTypeChange = this.handleTypeChange.bind(this);
-  }
-
-  handleTypeChange(value, name) {
-    const {credentials, credential_id, onScannerTypeChange} = this.props;
-
+const SCANNER_DIALOG_MAX_HEIGHT = '300px';
```
swaterkamp:
```suggestion
const SCANNER_DIALOG_INITIAL_HEIGHT = '300px';
```

PR #1761, opened 2019-11-06
[8.0] Reload on inactive tabs

bjoernricks:
Improve Reload component to consider hidden status of the current browser tab for calculating the reload interval.

**Checklist**:

- [x] Tests


swaterkamp:
I'm not a 100% sure about the differing intervals of 3 and 5 seconds respectively.

bjoernricks:
Currently the reload intervals are by default (via gmp settings):
* normal -> 15 seconds
* active -> 3 seconds
* inactive -> 60 seconds

let us discuss them before we merge the PR.

PR #1782, opened 2019-11-13
Form validation feature

saberlynx:
This PR contains the new form validation feature.

As a proof of concept, this has been implemented on three dialogs to hopefully cover all common cases:

- Create ticket dialog. The simplest case. Check if input is valid. If yes, save.
- Edit ticket dialog. Conditional validation: whether an input is valid depends on other variables that don't need to be validated

**Checklist**:

- [x] Tests


gsa/src/web/components/form/useForm.js
```diff
@@ -0,0 +1,109 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+import {useState, useEffect} from 'react';
+
+export const capitalizeFirstLetter = string => {
```
swaterkamp, line 21: We already have a capitalizeFirstLetter function in gmp/utils/string

gsa/src/web/components/form/useForm.js
```diff
@@ -0,0 +1,109 @@
...
+
+  const words = parsedString.split(' ');
+  let errorString = '';
+
+  words.forEach(word => (errorString += capitalizeFirstLetter(word) + ' '));
+  return errorString.trim(); // parse "name" to something human readable. Accepts underscores and camelcase. "credential_id" > "Credential Id", etc.
+};
+
+export const syncVariables = (values, formState, dependencies = {}) => {
+  Object.keys(formState).forEach(key => (values[key] = formState[key]));
```
swaterkamp, line 39: A polyfill for IE11 is missing. `import 'core-js/features/object/keys';`

gsa/src/web/components/form/useForm.js
```diff
@@ -0,0 +1,109 @@
...
+    formStatus,
+    formState,
+    handleDependencyChange,
+    handleSubmit,
+    handleValueChange,
+  };
+};
+
+// Some common validation rules
+export const testNonEmptyString = string => string.length > 0 && string !== ' ';
```
swaterkamp, line 106: This will invalidate `' '` but not `'  '` or `'     '`

saberlynx: Hm. maybe we should use a trim() function on the string and see if it's still length zero. And for that matter we should also invalidate stuff like ' hi'. Maybe.

gsa/src/web/pages/scanners/validationrules.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import {isDefined} from 'gmp/utils/identity';
+
+import {testNonEmptyString} from 'web/components/form/useForm';
+
+const validationRules = {
+  name: value => {
+    const validity = testNonEmptyString(value);
+    const error = validity ? '' : 'Missing name.';
```
swaterkamp, line 27: We need these to be translatable.

gsa/src/web/components/form/useForm.js
swaterkamp: I would just trim ' hi' to 'hi' without even asking the user or something

gsa/src/web/pages/scanners/dialog.js
```diff
@@ -185,7 +185,7 @@ const ScannerDialog = ({
 
   const {gmp} = props;
 
-  if (gmp.settings.enableGreenboneSensor) {
+  if (isDefined(gmp) && gmp.settings.enableGreenboneSensor) {
```
swaterkamp, line 188: Is this necessary? If this is needed for tests only, we should make sure that we provide gmp in the test.

gsa/src/web/pages/tickets/editdialog.js
```diff
@@ -32,6 +32,9 @@ import TextArea from 'web/components/form/textarea';
 
 import PropTypes from 'web/utils/proptypes';
 import {renderSelectItems} from 'web/utils/render';
+import useForm, {syncVariables} from 'web/components/form/useForm';
+import {editTicketRules as validationRules} from './validationrules';
+import ErrorBubble from 'web/components/form/errorbubble';
```
swaterkamp, line 38: Remember to sort the imports by our scheme

gsa/src/web/components/form/useForm.js
```diff
@@ -0,0 +1,109 @@
...
+    Object.keys(formState).forEach(key => {
+      !formStatus[key].validity && errorVars.push(parseAlias(key)); // Collects form fields with errors
+    });
+
+    const hasErrorInState = Object.keys(formState).some(key => {
+      return formStatus[key].validity === false;
+    }, []); // checks if any field is invalid
+
+    if (hasErrorInState) {
+      setErrorMessage('Error in fields: ' + errorVars.join(', ').toString()); // Outputs an error string containing all errors, or does nothing at all except show error bubbles. This is good for multi step dialogs.
```
swaterkamp, line 86: We need to make sure all strings are translatable.

gsa/src/web/components/form/useForm.js
```diff
@@ -0,0 +1,109 @@
...
+  words.forEach(word => (errorString += capitalizeFirstLetter(word) + ' '));
+  return errorString.trim(); // parse "name" to something human readable. Accepts underscores and camelcase. "credential_id" > "Credential Id", etc.
+};
+
+export const syncVariables = (values, formState, dependencies = {}) => {
+  Object.keys(formState).forEach(key => (values[key] = formState[key]));
+  Object.keys(dependencies).forEach(key => (values[key] = dependencies[key]));
+}; // sync all form states with the values known to SaveDialog
+
+const useForm = (stateSchema, validationRules, callback, extras = {}) => {
```
swaterkamp, line 43: I think, we should name this more appropriately. We are not using a form, we are using form validation. => I'd call it useFormValidation.

saberlynx: This is a great idea. Implemented.

gsa/src/web/components/form/useFormValidation.js
```diff
@@ -110,7 +105,7 @@ const useFormValidation = (
 };
 
 // Some common validation rules
-export const testNonEmptyString = string => string.length > 0 && string !== ' ';
+export const testNonEmptyString = string => string.trim().length > 0;
```
bjoernricks, line 108: Shouldn't the test function be safe against undefined values? In that case it would be better to use something like
```
(string = '') => string.trim().length > 0;

File: gsa/src/web/components/form/useFormValidation.js [hunk 10]
```diff
@@ -110,7 +105,7 @@ const useFormValidation = (
...
 export const testValidPassword = password => password.length > 5;
```
bjoernricks, line 109: Same here
```
(password = '') => password.length > 5;

File: gsa/src/web/components/form/useForm.js (hunk 7 above)
Comment (saberlynx) [reply gsa/src/web/components/form/useForm.js]: I'm not quite sure how to make dynamically generated error messages translatable. Maybe we need to drop this feature and look for a workaround to achieve the same effect of showing error in fields not rendered.

File: gsa/src/web/components/form/useFormValidation.js (hunk 9 above)
Comment (saberlynx) [reply gsa/src/web/components/form/useFormValidation.js:108]: Good idea. Thanks!

File: gsa/src/web/components/form/textfield.js [hunk 11]
```diff
@@ -18,26 +18,31 @@
...
-  const checkMarkColor = shouldValidate ? Theme.warningRed : 'white';
+const TextField = ({hasError = false, errorContent, title, ...props}) => {
   return (
     <React.Fragment>
       <Field
         {...props}
-        title={errorContent}
-        style={{background: bgColor}}
+        hasError={hasError}
+        title={hasError ? errorContent : title}
```
saberlynx, line 32: Do we need both title and errorContent? What is the usecase you have in mind?

swaterkamp: If we have an input field were we would like to actually set a title (as tooltip for example) this would get swallowed here.

gsa/src/web/components/layout/withLayout.js
```diff
@@ -78,7 +78,9 @@ const withLayout = (defaults = {}) => Component => {
     wrap: PropTypes.oneOf([true, 'wrap', 'nowrap']),
   };
 
-  return LayoutComponent;
+  return React.forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
swaterkamp: yes

PR #1809, opened 2019-11-26
Fix icon titles

bjoernricks:
Finally fix the icon titles.

**Checklist**:

- [x] Tests


gsa/src/web/components/icon/testing.js
```diff
@@ -52,6 +50,14 @@ export const testIcon = Icon => {
     });
   });
 
+  test('should change appearance when if inactive', () => {
```
sarahd93:
```suggestion
  test('should change appearance when inactive', () => {
```

Issue #1807, opened 2019-11-25
Missing text fields


Niii:
Hi,

I dunno if I missing something but on gsa 8.0+beta2 (from pkg on centos7), 8.0.1 and 9.0.0 (from source on debian10) some text fields are missing like when I try to create a new override.
Tested on Chrome, Firefox and IE.
![image](https://user-images.githubusercontent.com/1022522/69571297-0fb7d780-0fc2-11ea-8402-0aa6f9d3eb53.png)

With gsa 7.x on Kali text fields are present.

Has it ever happened to someone?

bjoernricks:
Could you show me a screenshot of the missing fields?

Niii:
Something like this ?
![image](https://user-images.githubusercontent.com/1022522/69620481-be4d2e00-103d-11ea-9551-10399dfb470a.png)

bjoernricks:
Sorry but I can't reproduce your issue.

![image](https://user-images.githubusercontent.com/897575/69647443-c9b64e80-1069-11ea-8759-7aeff159e7c1.png)

Also it is strange that this happens only in one single dialog because this dialog isn't different from others. So either no input field should be visible or all. Could you try to change the (GTK) theme if they are getting visible again? We are using the browser defaults (and therefore the GUI library defaults) for the input fields.

Niii:
Tested on different OS and browser, with and without theme, same default.
It's like the input box is missing when I check source code.

![image](https://user-images.githubusercontent.com/1022522/69668342-fa10e380-108f-11ea-92b2-fc01e7c0e929.png)

bjoernricks:
This looks really strange. Does this happen in other dialogs too? The `<span>` tag is suspicious. At first sight it seems it is opened as the edit dialog in the fashion of the create dialog.

It should be like

![image](https://user-images.githubusercontent.com/897575/69703066-e21d7c00-10f0-11ea-96f7-b7c7fbe5e8e2.png)

bjoernricks:
ok I guess you are using a different procedure to open this specific dialog. From you screenshot it looks like you are opening the dialog from an existing result right?

In that case it should be something like this

![image](https://user-images.githubusercontent.com/897575/69703456-b949b680-10f1-11ea-878e-7b85ffbb842f.png)

Niii:
Maybe I use the wrong process yes, I'm new with Greenbone.

[...]

PR #1814, opened 2019-11-27
Override: Make host/port/task fields editable even when fixed, and display oid when no name is defined

saberlynx:
**Checklist**:

- [] Tests


bjoernricks:
Should fix #1807

PR #1833, opened 2019-12-05
Changed new override dialog defaults

saberlynx:
**Checklist**:

- [] Tests


gsa/src/web/pages/overrides/dialog.js
```diff
@@ -430,7 +430,7 @@ const OverrideDialog = ({
               <Radio
                 name="result_id"
                 title={_('Any')}
-                checked={state.result_id === RESULT_ANY}
+                checked={true}
```
bjoernricks: I don't think this is correct. Setting it to true always doesn't allow to change it to unchecked.

PR #1834, opened 2019-12-05
Limit the length of result description with a div

saberlynx:
**Checklist**:

- [] Tests


bjoernricks:
The changelog needs an update

PR #1839, opened 2019-12-06
Single redux delete action

saberlynx:
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/entities/container.js
+++ b/gsa/src/web/entities/container.js
@@ -54,6 +54,8 @@ import PropTypes from 'web/utils/proptypes';
 import {generateFilename} from 'web/utils/render';
 import SelectionType from 'web/utils/selectiontype';
 
+import {createDeleteEntity} from 'web/store/entities/utils/actions';
+
 import SortBy from 'web/components/sortby/sortby';
 
 import TagDialog from 'web/pages/tags/dialog';
@@ -92,6 +94,7 @@ class EntitiesContainer extends React.Component {
 
     this.handleChanged = this.handleChanged.bind(this);
     this.handleCreateTag = this.handleCreateTag.bind(this);
+    this.handleDeleted = this.handleDeleted.bind(this);
     this.handleDeselected = this.handleDeselected.bind(this);
     this.handleDeleteBulk = this.handleDeleteBulk.bind(this);
     this.handleDownloadBulk = this.handleDownloadBulk.bind(this);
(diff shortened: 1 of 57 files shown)


gsa/src/web/entities/container.js
```diff
@@ -171,6 +174,12 @@ class EntitiesContainer extends React.Component {
     this.props.reload(filter);
   }
 
+  handleDeleted(entity) {
```
bjoernricks:
```suggestion
  handleDelete(entity) {
```

gsa/src/web/entities/container.js
```diff
@@ -526,6 +535,7 @@ class EntitiesContainer extends React.Component {
           sortBy,
           sortDir,
           onChanged: this.handleChanged,
+          onDeleted: this.handleDeleted,
```
bjoernricks: Not sure if this will break something
```suggestion
          onDelete: this.handleDelete,
```

gsa/src/web/pages/reports/listpage.js
```diff
@@ -194,8 +194,8 @@ class Page extends React.Component {
   }
 
   handleReportDeleteClick(report) {
-    const {gmp, onChanged, onError} = this.props;
-    return gmp.report.delete(report).then(onChanged, onError);
+    const {onDeleted} = this.props;
+    return onDeleted(report);
```
bjoernricks: Should be named `onDelete` because we are going to delete something and aren't calling this function after something has been deleted.
```suggestion
    return onDelete(report);
```

gsa/src/web/pages/reports/listpage.js
```diff
@@ -268,6 +268,7 @@ Page.propTypes = {
   loadTasks: PropTypes.func.isRequired,
   tasks: PropTypes.arrayOf(PropTypes.model),
   onChanged: PropTypes.func.isRequired,
+  onDeleted: PropTypes.func.isRequired,
```
bjoernricks:
```suggestion
  onDelete: PropTypes.func.isRequired,
```

gsa/src/web/store/entities/utils/__tests__/actions.js
```diff
@@ -662,4 +663,33 @@ describe('entities actions tests', () => {
...
   });
 });
+
+describe('createDeleteEntity tests', () => {
+  test('should create a entity delete success action', () => {
+    const id = 'id1';
+    const gmp = {
+      foo: {
+        delete: jest.fn().mockReturnValue(
+          Promise.resolve({
```
bjoernricks: This can be simplified by using [jest.fn().mockResolvedValue({data: 'ba'});](https://jestjs.io/docs/en/mock-function-api#mockfnmockresolvedvaluevalue)

gsa/src/web/pages/reports/listpage.js
saberlynx: Good point. Will do.

PR #1847, opened 2019-12-09
Fix undefined policy in create audit dialog

sarahd93:
**Checklist**:

- [] Tests

### DIFF
--- a/gsa/src/web/pages/audits/dialog.js
+++ b/gsa/src/web/pages/audits/dialog.js
@@ -22,7 +22,6 @@ import React from 'react';
 import _ from 'gmp/locale';
 
 import {isDefined} from 'gmp/utils/identity';
-import {selectSaveId} from 'gmp/utils/id';
 
 import {NO_VALUE, YES_VALUE} from 'gmp/parser';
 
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/audits/dialog.js
```diff
@@ -132,8 +131,6 @@ const AuditDialog = ({
       values={controlledData}
     >
       {({values: state, onValueChange}) => {
-        const currentPolicyId = selectSaveId(policies, state.policyId);
```
bjoernricks: Ok this line was definitively wrong.

gsa/src/web/pages/audits/dialog.js
```diff
@@ -71,7 +70,7 @@ const AuditDialog = ({
   maxHosts = DEFAULT_MAX_HOSTS,
   name = _('Unnamed'),
   policies = [],
-  policyId,
+  policyId = UNSET_VALUE,
```
bjoernricks: I am not sure about this change. Shouldn't we ensure that a valid policy is used instead of allowing the unset value?

PR #1852, opened 2019-12-11
Threshold warning for content composer

swaterkamp:
**Checklist**:

- [] Tests


gsa/src/web/components/dialog/composercontent.js
```diff
@@ -40,6 +42,20 @@ export const COMPOSER_CONTENT_DEFAULTS = {
   includeOverrides: YES_VALUE,
 };
 
+export const ThresholdMessage = threshold => {
+  return (
```
bjoernricks: Could be simplified and put into an own js module. Also it isn't a component. When fixing the props params the proptypes need to be added too.
```js
const ThresholdMessage = ({threshold}) => (
  <ErrorContainer>
```

gsa/src/web/pages/reports/downloadreportdialog.js
```diff
@@ -54,7 +55,9 @@ const DownloadReportDialog = ({
   includeOverrides = COMPOSER_CONTENT_DEFAULTS.includeOverrides,
   reportFormatId,
   reportFormats,
+  showThresholdMessage,
```
bjoernricks: Otherwise you should required the prop.
```suggestion
  showThresholdMessage = false,
```

gsa/src/web/pages/reports/triggeralertdialog.js
```diff
@@ -56,7 +57,9 @@ const TriggerAlertDialog = ({
   filter = {},
   includeNotes = COMPOSER_CONTENT_DEFAULTS.includeNotes,
   includeOverrides = COMPOSER_CONTENT_DEFAULTS.includeOverrides,
+  showThresholdMessage,
```
bjoernricks: Same here. Add the default or require the prop.
```suggestion
  showThresholdMessage = false,
```

gsa/src/web/components/dialog/thresholdmessage.js
```diff
@@ -0,0 +1,46 @@
...
+        'the threshold of {{threshold}}. Therefore, this action can take ' +
+        'a really long time to finish. It might even exceed the session ' +
+        'timeout!',
+      threshold,
+    )}
+  </ErrorContainer>
+);
+
+ThresholdMessage.propTypes = {
+  threshold: PropTypes.number,
```
bjoernricks: IMHO the threshold is required because otherwise the dialog would contain `... the threshold of undefined. Therefore ...`
```suggestion
  threshold: PropTypes.number.isRequired,
```

PR #1863, opened 2019-12-13
Report results loading

bjoernricks:
Missing:
- ~~Report ID filter~~
- ~~Reloading~~
- ~~fix isUpdating indicator while loading full report~~
- fix changing filter while loading report -> was broken before and will be fixed in another PR
- ~~fix setting sort=foo when sort-reverse=bar is applied~~
- ~~Error handling~~

**Checklist**:
- [ ] Tests


gsad/src/gsad_gmp.c
```diff
@@ -1255,11 +1256,27 @@ get_many (gvm_connection_t *connection, const char *type,
             }
           else if (str_equal (type, "result"))
             {
+              gchar *report_id_filter;
+              and_report_id = params_value (params, "_report_id");
```
timopollmeier: Shouldn't this get the `_and_report_id` parameter instead of `_report_id`?
```suggestion
              and_report_id = params_value (params, "_and_report_id");
```

gsad/src/gsad_gmp.c
```diff
@@ -1255,11 +1256,27 @@ get_many (gvm_connection_t *connection, const char *type,
...
+
+              if (and_report_id == NULL || str_equal (and_report_id, ""))
+                {
+                  report_id_filter =
+                    g_strdup_printf ("_and_report_id=%s", and_report_id);
```
timopollmeier: I think this assignment is in the block for the wrong condition as it's writing the keyword if the parameter is missing or empty.
It looks like it should be in the `else` block instead while the one from there should be here.

gsad/src/gsad_gmp.c
```diff
@@ -1255,11 +1256,27 @@ get_many (gvm_connection_t *connection, const char *type,
             }
           else if (str_equal (type, "result"))
             {
+              gchar *report_id_filter;
+              and_report_id = params_value (params, "_report_id");
+
+              if (and_report_id == NULL || str_equal (and_report_id, ""))
```
timopollmeier: It may be a good idea to add a check to see if `_and_report_id` is invalid or just missing.

gsad/src/gsad_gmp.c
bjoernricks: Of course it should be `_and_report_id`. Fixed.

gsad/src/gsad_gmp.c
bjoernricks: You are completely right. I've mixed up both branches.

gsad/src/gsad_gmp.c
bjoernricks: I've added a check for the `_and_report_id` to the get_results function because I feared to leave a memory leak if this check gets added to get_many.

bjoernricks:
Thanks for the reviews :+1:

PR #1869, opened 2019-12-17
Severity levels

bjoernricks:
Remove severity levels from filter if all levels are unchecked.

**Checklist**:

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Lower memory usage when getting a report [#1858](https://github.com/greenbone/gvmd/pull/1858)
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Lower memory usage when getting a report [#1858](https://github.com/greenbone/gvmd/pull/1858)
 
 ### Fixed
+- Fixed removing levels filter keyword if all severity levels are unchecked [#1869](https://github.com/greenbone/gsa/pull/1869)
```
swaterkamp: Isn't this a new addition rather than a fix? I would put it under ### Added

bjoernricks: I don't think so. It is a fix of wrong behavior. Currently removing all levels results in filtering all columns for a levels value. The filter will be  `~levels`.

PR #1908, opened 2020-01-09
handle undefined trash in case of an error

sarahd93:
In case of an error on the trashcanpage the trash might be undefined which would cause the trashcanpage to break. This PR should fix that by catching this special case.

**Checklist**:

- [N/A] Tests

### DIFF
--- a/gsa/src/web/pages/extras/trashcanpage.js
+++ b/gsa/src/web/pages/extras/trashcanpage.js
@@ -352,15 +352,13 @@ class Trashcan extends React.Component {
     }
 
     const {scan: tasks, compliance: audits} = separateByUsageType(
-      isDefined(trash) ? trash.task_list : undefined,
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -342,20 +342,25 @@ class Trashcan extends React.Component {
...
 
     if (!isDefined(trash) && !isDefined(error)) {
       return <Loading />;
+    } else if (!isDefined(trash) && isDefined(error)) {
+      trash = {};
     }
 
     const {scan: tasks, compliance: audits} = separateByUsageType(
-      trash.task_list,
+      isDefined(trash) ? trash.task_list : undefined,
```
bjoernricks: This doesn't work anymore. trash is always defined.

PR #1935, opened 2020-01-20
Remove componentWillReceiveProps

sarahd93:
ComponentWillReceiveProps is deprecated. Update components that still use it to use getDerivedStateFromProps instead.

**Checklist**:

- [N/A] Tests

### DIFF
--- a/gsa/src/web/components/form/multiselect.js
+++ b/gsa/src/web/components/form/multiselect.js
@@ -83,6 +83,7 @@ class MultiSelect extends React.Component {
     const {value} = this.props;
 
     this.state = {
+      value,
       search: '',
       selectedItems: isArray(value) ? value : [],
     };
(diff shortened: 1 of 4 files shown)


gsa/src/web/components/form/multiselect.js
```diff
@@ -144,12 +145,12 @@ class MultiSelect extends React.Component {
     this.notifyChange(copy);
   }
 
-  componentWillReceiveProps(nextProps) {
-    const {value} = nextProps;
-
-    if (isDefined(value) && !arraysEqual(value, this.state.selectedItems)) {
-      this.setState({selectedItems: value});
+  static getDerivedStateFromProps(props, state) {
```
bjoernricks: I am really not sure if using MultiSelect uncontrolled still works now. Maybe we should just remove supporting the uncontrolled case and drop this completely?

sarahd93: This is basically the test case "Should change displayed values" in the multiselect tests. The test passes so it should still work. But I am not sure if this is actually something that happens in GSA. If it never happens it should be safe to remove it. Everything else seems to be working fine without getDerivedStateFromProps.

bjoernricks: I just had a quick look at the MultiSelect code and I would love to remove the internal state and just use the props.value as selected items. If we only provide the controlled mode (by removing the internal state selectedItems) we can just drop the cwrp/getDerivedStateFromProps method.

PR #1948, opened 2020-01-29
Add clean-up-translations script

sarahd93:
Add clean-up-translations script to remove empty strings from translations. This will ensure that the English text is used for strings that were not translated yet.

**Checklist**:

- [N/A] Tests


gsa/scripts/cleanuptranslations.js
```diff
@@ -0,0 +1,56 @@
...
+  } catch (err) {
+    console.log('An error occurred while trying to read from', path);
+    console.log(err);
+    continue;
+  }
+
+  // remove all strings that were not translated
+  const newContent = {};
+  Object.assign(newContent, content);
+  for (const key in content) {
```
bjoernricks: Please use for...of instead of for...in
```suggestion
  for (const key of Object.keys(content)) {
```

gsa/scripts/cleanuptranslations.js
```diff
@@ -0,0 +1,56 @@
...
+    content = JSON.parse(jsonString);
+  } catch (err) {
+    console.log('An error occurred while trying to read from', path);
+    console.log(err);
+    continue;
+  }
+
+  // remove all strings that were not translated
+  const newContent = {};
+  Object.assign(newContent, content);
```
bjoernricks: Personally I would prefer object destructuring here
```js
const newContent = {...content};
```

PR #1949, opened 2020-01-29
Fix task listpage observer tooltip

saberlynx:
**Checklist**:

- [x] Tests


gsa/src/web/pages/tasks/row.js
```diff
@@ -103,19 +103,25 @@ const Row = ({
...
-      if (isDefined(observers.group)) {
-        obs.push(_('Group {{name}}', {name: observers.group.name}));
-      }
+    if (isDefined(observers.user)) {
+      obs.user = _('Users ');
+      observers.user.forEach(user => (obs.user += user + ' '));
+      obs.user += '\n';
+    }
+    if (isDefined(observers.role)) {
+      obs.role = _('Roles ');
```
bjoernricks:
```suggestion
      obs.role = _('Roles {{roles}}', {roles: observers.role.join(' ')});
```

gsa/src/web/pages/tasks/row.js
```diff
@@ -103,19 +103,25 @@ const Row = ({
...
-      obs.push(_('User {{name}}', {name: observers}));
-    } else {
-      if (isDefined(observers.role)) {
-        obs.push(_('Role {{name}}', {name: observers.role.name}));
-      }
-      if (isDefined(observers.group)) {
-        obs.push(_('Group {{name}}', {name: observers.group.name}));
-      }
+    if (isDefined(observers.user)) {
+      obs.user = _('Users ');
```
bjoernricks:
```suggestion
      obs.user = _('Users {{users}}', {users: observers.users.join(' ')});
```

gsa/src/web/pages/tasks/row.js
```diff
@@ -103,19 +103,25 @@ const Row = ({
...
+      observers.user.forEach(user => (obs.user += user + ' '));
+      obs.user += '\n';
+    }
+    if (isDefined(observers.role)) {
+      obs.role = _('Roles ');
+      observers.role.forEach(role => (obs.role += role.name + ' '));
+      obs.role += '\n';
+    }
+    if (isDefined(observers.group)) {
+      obs.group = _('Groups ');
```
bjoernricks:
```suggestion
      obs.group = _('Groups {{groups}}', {groups: observers.groups.join(' ')});
```

gsa/src/web/pages/tasks/row.js
```diff
@@ -142,12 +148,17 @@ const Row = ({
...
             />
-            {isDefined(observers) && observers.length > 0 && (
+            {isDefined(observers) && Object.keys(observers).length > 0 && (
               <ProvideViewIcon
                 size="small"
-                title={_('Task made visible for: {{observers}}', {
-                  observers: obs.join(', '),
-                })}
+                title={_(
+                  'Task made visible for: \n {{users}} {{roles}} {{groups}}',
```
bjoernricks: This is not really translatable because users, roles and groups contain translations already. But currently I don't have a good idea how to fix this.

gsa/src/web/pages/tasks/row.js
bjoernricks: I guess we should also display users, roles and groups with comma separation because names may contain spaces already.
Should be e.g. 
```javascript
observers.groups.join(', ')
```

gsa/src/gmp/models/task.js
```diff
@@ -86,6 +87,14 @@ export function parse_yes(value) {
   return value === 'yes' ? YES_VALUE : NO_VALUE;
 }
 
+function parseIntoArray(obj) {
+  if (!isArray(obj)) {
+    const array = Array(obj);
```
bjoernricks:
```suggestion
    return [obj];
```
this function can also be shortened to
```javascript
const parseIntoArray = value => isArray(value) ? value : [value];
```

gsa/src/web/pages/tasks/row.js
saberlynx: Hm. I'll do some research into this.

gsa/src/gmp/models/task.js
saberlynx: This looks much cleaner. Thanks!

saberlynx:
> Could you also add or extend gmp/models/**tests**/task.js to contain a test for parsing the observers?

of course! :)

PR #1969, opened 2020-02-17
Fix deleteById() returning response

swaterkamp:
**Checklist**:

- [X] Tests


bjoernricks:
Nice catch!

PR #1970, opened 2020-02-17
Parse xml rejection message (unescape)

swaterkamp:
Error messages like "Failed to find task &apos;foo&apos;" are now handled correctly. This concerned xml gsad_response and action_result messages.

**Checklist**:

- [X] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Deleting a single entity now removes its ID from store [#1839](https://github.com/greenbone/gsa/pull/1839)
 
 ### Fixed
+- Fix parsing xml rejection messages [#1970](https://github.com/greenbone/gsa/pull/1970)
 - Fix returning bulk_delete response [#1969](https://github.com/greenbone/gsa/pull/1969)
 - Fix parsing DFN-Cert CVE entries [#1965](https://github.com/greenbone/gsa/pull/1965)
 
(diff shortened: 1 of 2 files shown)


bjoernricks:
I am really not sure why this should be necessary. If this is necessary for the rejection/error case it must be necessary for the success case too. So some other thing is wrong here.

bjoernricks:
Any by the way transform/xml.js is the **old** transform which shouldn't be used anymore. transform/fastxml.js should be the default now. IMHO transform/xml.js can be removed completely because it is unused. Any if it is still used this is a mistake and should be changed asap.

bjoernricks:
> Any by the way transform/xml.js is the old transform which shouldn't be used anymore. transform/fastxml.js should be the default now. IMHO transform/xml.js can be removed completely because it is unused. Any if it is still used this is a mistake and should be changed asap.

Ok this is wrong. transform.js contains the common code for the old (xml2js) and the current (fastxml) transforms. But I am still not sure why the new escaping is necessary.

swaterkamp:
It was necessary, because we didn't hand PARSER_OPTIONS to the parse() method for xml rejections. For the success case, we did that already and it worked fine.
updating PR...

bjoernricks:
> It was necessary, because we didn't hand PARSER_OPTIONS to the parse() method for xml rejections. For the success case, we did that already and it worked fine.

 :disappointed: stupid mistake ...

swaterkamp:
Indeed ;) I hardly found it, because it was such a small thing, but now everything should be fine.

bjoernricks:
Thanks so much for taking care!

PR #1975, opened 2020-02-19
Don't crash OsIcon when osCpe or osTxt are undefined

swaterkamp:
**Checklist**:

- [N/A] Tests


swaterkamp:
It now returns undefined, if it receives an empty string as name-param. The issue was that it crashed when name was undefined and .include() couldn't be called at all.

The real issue definitely lies deeper and has been discussed across several teams in the last days. With this small fix we just take care that whenever this kind of situation comes up again, GSA doesn't crash and at least allows the user to continue work.

bjoernricks:
Hmm after looking at the code and your changes again, I don't get how your PR fixes a problem. For me it doesn't really change the behavior.

bjoernricks:
The only case were it could crash is if osTxt is defined and not a string. In that case neither the old nor the new code works.

PR #1987, opened 2020-02-24
Fix auto_delete_value of 0 set by wizard

swaterkamp:
**Checklist**:

- [N/A] Tests

### DIFF
--- a/gsa/src/gmp/models/audit.js
+++ b/gsa/src/gmp/models/audit.js
@@ -189,8 +189,9 @@ class Audit extends Model {
                 : AUTO_DELETE_NO;
             break;
           case 'auto_delete_data':
+            const value = parseInt(pref.value);
             copy.auto_delete_data =
-              pref.value === 0
+              value === 0
                 ? AUTO_DELETE_KEEP_DEFAULT_VALUE
                 : parseInt(pref.value);
             break;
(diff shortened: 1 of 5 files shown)


swaterkamp:
We are setting that in the model, and I thought the wizard tasks were somehow circumventing that.
Having a second look now, however, it seems like we are checking against '0' instead of 0. That might be the actual problem. Revisiting right now.

swaterkamp:
Indeed that seemed to fix the problem. "Normal" tasks were working before the change to the model, that's what made me think about fixing it in the dialog instead.

bjoernricks:
The models should fix weird or unexpected behavior of our data. The models are intended to get a clean view of the data in our react layer.

gsa/src/gmp/models/audit.js
```diff
@@ -190,7 +190,7 @@ class Audit extends Model {
             break;
           case 'auto_delete_data':
             copy.auto_delete_data =
```
bjoernricks: Maybe this should be refactored a bit to be independent of the xml parser.
```javascript
const value = parseInt(pref.value);
copy.auto_delete_data = value === 0 ? AUTO_DELETE_KEEP_DEFAULT_VALUE : value;
```

gsa/src/gmp/models/task.js
```diff
@@ -237,7 +237,7 @@ class Task extends Model {
             break;
           case 'auto_delete_data':
             copy.auto_delete_data =
-              pref.value === '0'
+              pref.value === 0
```
bjoernricks: Same as above.

PR #1998, opened 2020-02-26
Translate About page to German

swaterkamp:
Also add some missing translations

**Checklist**:

- [N/A] Tests


gsa/public/locales/gsa-de.json
```diff
@@ -1,4 +1,6 @@
 {
+  "\nThis could have the following reasons:\n1. You might have followed an incorrect link and the CVE does not exist\n2. The CVE might not be included in the SCAP database yet. For new CVEs it can take a month or more until they become available.": "\nDies kann folgende Ursachen haben:\n1. Möglicherweise war der Link für diese CVE inkorrekt und sie existiert nicht.\n2. Die CVE ist vielleicht noch nicht in der SCAP-Datenbank enthalten. Für neue CVEs kann es einen Monat oder länger dauern bis sie verfügbar sind.",
+  "\nYou might have followed an incorrect link and the {{entity}} does not exist.": "Möglicherweise war der Link für diese Ressource inkorrekt und sie existiert nicht.",
```
bjoernricks: This translation doesn't really fit not does it contain the {{entity}}

gsa/public/locales/gsa-de.json
```diff
@@ -178,6 +181,7 @@
   "Base Score": "Basisscore",
   "Base URL": "Basis URL",
   "Base Vector": "Basisvektor",
+  "Base with a minimum set of NVTs": "Basis Konfiguration mit einer minimalen Menge an NVTs",
```
bjoernricks: Maybe better
"Basis Konfiguration mit einer minimalen Anzahl an NVTs" or
"Basis Konfiguration mit einem minimalen Satz von NVTs"

gsa/public/locales/gsa-de.json
```diff
@@ -187,8 +191,10 @@
   "CERT-Bund Advisories": "CERT-Bund-Advisories",
   "CERT-Bund Advisories Filter": "CERT-Bund Advisories-Filter",
   "CERT-Bund Advisories by CVSS (Total: {{count}})": "CERT-Bund Advisories nach CVSS (Gesamt: {{count}})",
+  "CERT-Bund Advisories by CVSS (Total: {{count}})nullplural": "",
```
bjoernricks: What's that?

gsa/public/locales/gsa-de.json
```diff
@@ -690,6 +704,7 @@
   "Fold": "Einklappen",
   "Fold all details": "Alle Details einklappen",
   "Follow Link": "Link folgen",
+  "Foo": "Foo",
```
bjoernricks: ???

gsa/src/web/pages/help/about.js
```diff
@@ -116,7 +116,7 @@ const About = ({gmp}) => (
           </DivP>
           <DivP>
             {_('The GMP documentation is available ')}
-            <ProtocolDocLink title="here" />.
+            <ProtocolDocLink title={_('here')} />.
```
bjoernricks: The whole think will not be translatable but I also don't have a better solution

swaterkamp: The German translation is "Die GMP-Dokumentation finden Sie hier." and it's working perfectly on the about page. We need to deal with it again, once we offer more languages, but until then it's fine, I think,

gsa/public/locales/gsa-de.json
swaterkamp: This was more of a joke. It's used in one of our tests and shows up for translation ;)

gsa/public/locales/gsa-de.json
swaterkamp: It's an artifact of the i18n-extract command. If there is the keyword "count" in any of the to-be-translated strings, it puts a nullplural in the end without a translation. Normally this will be removed, as long as we don't translate it, with yarn clean-up-translations. I'll clean it up once more to get rid of this.

gsa/public/locales/gsa-de.json
swaterkamp: As above, this translation was already in GSA, I just put it back in. I can change it, of course.

gsa/public/locales/gsa-de.json
swaterkamp: It doesn't contain the entity to prevent using "der/die/das" in the German translation, as we have no idea what kind of entity it will be.
This translation was already in GSA before it was lost with the merge that Sarah mentioned.

swaterkamp: Happy for any better translation :)

gsa/public/locales/gsa-de.json
bjoernricks: Ah ok. We should not extract strings from tests in future :-)

gsa/src/web/pages/help/about.js
bjoernricks: Yeah it will be fine for German and I really don't know a better or reliable solution currently. We'll leave it as it is.

gsa/public/locales/gsa-de.json
bjoernricks: ok. strange thing. I could image that our setup is wrong for this kind of plural/singular thing atm. Nevertheless we should not care about this in this pr.

gsa/public/locales/gsa-de.json
bjoernricks: You can decide what's best.

gsa/public/locales/gsa-de.json
bjoernricks: Should we ask @k-schlosser?

gsa/public/locales/gsa-de.json
swaterkamp: Actually, if we just clean up properly after an i18n-extract again, this will not show up at all. I don't think this will ever be a problem. The clean-up script seems not to have been executed correctly. If it had, you would never have seen this ;)

gsa/public/locales/gsa-de.json
swaterkamp: Done. We agreed on leaving the entity out because of the article issue and k-schlosser approximated the first part better to the English original.

PR #2029, opened 2020-03-12
Hyperion api settings

bjoernricks:


gsa/src/web/pages/login/loginpage.js
```diff
@@ -121,33 +119,44 @@ const isIE11 = () =>
     : false;
 
 const LoginPage = props => {
```
saberlynx: Maybe use `const LoginPage = () => {...` since there are no props?

bjoernricks: oh you are right

PR #2048, opened 2020-03-18
Fix audit listpage observer tooltip

sarahd93:
ProvideViewIcon in audit row was only shown for users but not for roles and groups. Applied the changes in task model to audits as well to fix this.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/task.js
+++ b/gsa/src/gmp/models/task.js
@@ -27,10 +27,11 @@ import {
   parseInt,
   parseProgressElement,
   parseYesNo,
+  parseYes,
+  parseIntoArray,
   parseText,
   parseDuration,
   NO_VALUE,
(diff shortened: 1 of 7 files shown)


gsa/src/gmp/models/task.js
```diff
@@ -87,7 +87,7 @@ export function parse_yes(value) {
   return value === 'yes' ? YES_VALUE : NO_VALUE;
 }
 
-const parseIntoArray = value => (isArray(value) ? value : [value]);
+export const parseIntoArray = value => (isArray(value) ? value : [value]);
```
swaterkamp: Could you extract this function in to gmp/parser, please? It should not be exported by one model to use in another. We'd have an unnecessary dependency between task and audit model.

PR #2049, opened 2020-03-18
Add GraphQL capabilities to task listpage icons and update tests

saberlynx:
**Checklist**:

- [x] Tests


gsa/src/web/entities/table.js
```diff
@@ -270,7 +270,11 @@ class EntitiesTable extends React.Component {
         <UpdatingStripedTable
           header={header}
           footer={footer}
-          isUpdating={isUpdating}
+          isUpdating={
+            isDefined(entities) && entities[0].entityType === 'task'
```
bjoernricks: isArray is from gmp/utils/identity and getEntityType from gmp/utils/entitytpe
```suggestion
            isArray(entities) && getEntityType(entities[0]) === 'task'
```

gsa/src/web/entity/icon/cloneicon.js
```diff
@@ -38,6 +39,17 @@ const EntityCloneIcon = ({
   onClick,
   ...props
 }) => {
+  let capabilities;
+
+  const query = useGetCaps();
```
bjoernricks: Shouldn't useGetCaps be extended to return the Capabilities class directly?

gsa/src/web/entity/icon/__tests__/cloneicon.js
```diff
@@ -39,7 +39,7 @@ describe('Entity CloneIcon component tests', () => {
     const {render} = rendererWith({capabilities: caps});
 
     const {element} = render(
-      <CloneIcon entity={entity} onClick={clickHandler} />,
+      withEmptyMock(<CloneIcon entity={entity} onClick={clickHandler} />),
```
bjoernricks: I guess withEmptyMock could be integrated into the render function directly because it wouldn't make a difference for components not using graphql yet.

gsa/src/web/entity/icon/editicon.js
```diff
@@ -25,18 +25,29 @@ import {getEntityType, typeName} from 'gmp/utils/entitytype';
...
+
+  const query = useGetCaps();
+  const {data} = query();
+
+  if (isDefined(data)) {
+    capabilities = new Capabilities(data.capabilities);
+  } else {
+    capabilities = props.capabilities;
+  }
+
```
bjoernricks: If you copy & paste something more then two times it really needs to be abstracted in some kind. Not sure if a function can be used here or it needs to be a react hook (which is actually also a function just the name must start with use).

gsa/src/web/entity/icon/__tests__/cloneicon.js
saberlynx: Hm. I don't actually know what will happen if we just wrap all components with it. Should be fine though. I'll try it.

gsa/src/web/pages/tasks/graphql.js
```diff
@@ -32,6 +34,17 @@ export const useGetCaps = () => {
   return toFruitfulQuery(useQuery)(GET_CAPS);
 };
 
+export const useCaps = caps => {
```
bjoernricks: Looks good. Just some additions :-)

Could you move it to web/utils/useCapabilities.js and rename the hook to useCapabilities? We have code completion in our editors therefore we should use meaningful and expressive names for function etc.

gsa/src/web/entities/table.js
saberlynx: Done!

gsa/src/web/entity/icon/__tests__/cloneicon.js
bjoernricks: I am going to look at the apollo testing docs and will come up with an idea

gsa/src/web/utils/useCapabilities.js
```diff
@@ -0,0 +1,46 @@
...
+
+import {toFruitfulQuery} from 'web/utils/graphql';
+
+export const GET_CAPS = gql`
+  query Capabilities {
+    capabilities
+  }
+`;
+
+export const useGetCaps = () => {
```
bjoernricks:
```suggestion
export const useGetCapabilities = () => {
```

gsa/src/web/utils/testing.js
```diff
@@ -132,7 +134,9 @@ export const rendererWith = (
         <TestingGmpPropvider gmp={gmp}>
           <TestingCapabilitiesProvider capabilities={capabilities}>
             <TestingStoreProvider store={store}>
-              <TestingRouter history={history}>{ui}</TestingRouter>
+              <MockedProvider mocks={[]} addTypename={false}>
```
bjoernricks: Looks very good to me. I think we can extend this in future too and add something like `mocks` to the rendererWith params and pass it to the MockedProvider :-)

gsa/src/web/pages/tasks/component.js
```diff
@@ -96,15 +95,25 @@ import ScheduleComponent from 'web/pages/schedules/component';
...
 import ContainerTaskDialog from './containerdialog';
 import {setTimezone} from 'web/store/usersettings/actions';
 import {useModifyTask, useCreateContainerTask, useCreateTask} from './graphql';
+import {useGetCaps} from 'web/utils/useCapabilities';
 
 const TaskComponent = props => {
   const modifyTask = useModifyTask();
   const createTask = useCreateTask();
   const createContainerTask = useCreateContainerTask();
+  const query = useGetCaps();
```
bjoernricks:
```suggestion
  const query = useGetCapabilities();
```

gsa/src/web/pages/tasks/component.js
```diff
@@ -96,15 +95,25 @@ import ScheduleComponent from 'web/pages/schedules/component';
 import AlertComponent from 'web/pages/alerts/component';
 import TargetComponent from 'web/pages/targets/component';
 
+import Capabilities from 'gmp/capabilities/capabilities';
 import TaskDialog from './dialog';
 import ContainerTaskDialog from './containerdialog';
 import {setTimezone} from 'web/store/usersettings/actions';
 import {useModifyTask, useCreateContainerTask, useCreateTask} from './graphql';
+import {useGetCaps} from 'web/utils/useCapabilities';
```
bjoernricks:
```suggestion
import {useGetCapabilities} from 'web/utils/useCapabilities';
```

PR #2050, opened 2020-03-19
Add useCapabilities to detailslink

saberlynx:
**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/components/link/detailslink.js
+++ b/gsa/src/web/components/link/detailslink.js
@@ -20,9 +20,13 @@ import React from 'react';
 import PropTypes from 'web/utils/proptypes';
 import withCapabilities from 'web/utils/withCapabilities';
 
+import {useCapabilities} from 'web/utils/useCapabilities';
+
 import Link from './link';
 
-const DetailsLink = ({capabilities, id, type, textOnly = false, ...props}) => {
+const DetailsLink = ({id, type, textOnly = false, ...props}) => {
+  const capabilities = useCapabilities(props.capabilities);
+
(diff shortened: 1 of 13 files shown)


gsa/src/web/pages/audits/__tests__/__snapshots__/details.js.snap
```diff
@@ -147,6 +147,7 @@ exports[`Audit Details tests should render full audit details 1`] = `
     </h2>
     <div>
       <a
+        capabilities="[object Object]"
```
bjoernricks: This is caused by not removing the capabilities from the props anymore.

gsa/src/web/components/link/detailslink.js
```diff
@@ -20,9 +20,13 @@ import React from 'react';
 import PropTypes from 'web/utils/proptypes';
 import withCapabilities from 'web/utils/withCapabilities';
 
+import {useCapabilities} from 'web/utils/useCapabilities';
+
 import Link from './link';
 
-const DetailsLink = ({capabilities, id, type, textOnly = false, ...props}) => {
+const DetailsLink = ({id, type, textOnly = false, ...props}) => {
+  const capabilities = useCapabilities(props.capabilities);
```
bjoernricks: IMHO this should work too.
```suggestion
const DetailsLink = ({capabilities, id, type, textOnly = false, ...props}) => {
  capabilities = useCapabilities(capabilities);
```

saberlynx:
> Not sure if we should merge this pr and update it afterwards or if we should drop it for now.

agreed

PR #2051, opened 2020-03-19
Adjust parsing of filter strings

sarahd93:
Parsing filter strings that include substrings in double quotes is causing issues. Adjust parsing functions to be able to handle double quotes.

**Checklist**:

- [x] Tests


sarahd93:
I already started writing tests :)

You are right about "foo bar"=def, I am going to check again if I can catch some more of those edge cases.

PR #2055, opened 2020-03-20
Get rid of warnings in tests and components

swaterkamp:
afaics this fixes all warnings coming up during tests except "setting state on unmounted component" for SvgIcon and onActivate and isActive props on DOM elements.

**Checklist**:

- [X] Tests


gsa/src/web/pages/reports/details/applicationstable.js
```diff
@@ -105,7 +105,15 @@ const Row = ({entity, links = true, onToggleDetailsClick, ...props}) => {
 };
 
 Row.propTypes = {
-  entity: PropTypes.model.isRequired,
+  entity: PropTypes.shape({
```
bjoernricks: Nice :+1:

PR #2063, opened 2020-03-23
No state update in unmounted SvgIcons

swaterkamp:
This PR introduces two custom hooks to the SvgIcon component.

useStateWithMountCheck() replaces the normal useState() hook in the SvgIcon to set isLoading. It checks whether the component is actually mounted before trying to set state.

Whether the component is mounted is stored in a ref (with useRef()). ref.current will be consistent across renders and will not define a new handleClick() if the icon is rerendered, leading to an inconsistent state use, basically breaking the isLoading functionality.

The whole idea is based in the clean-up function at the end of the hook. Normally, the state update on unmounted components is fixed with a useEffect() hook, like this:

useEffect(() => {
  let isMounted = true;
  if (isMounted) {setState(...)};
  return () => isMounted = false;
})

The definition of handleClick would have to be placed inside the if, as the handler's setState takes place there.
So the useEffect does not work when wrapped around the handleClick, as both should be defined top-level in SvgIcon.

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/components/icon/svgicon.js
+++ b/gsa/src/web/components/icon/svgicon.js
@@ -56,12 +56,12 @@ const Styled = styled.span`
   }
 `;
 
-const useIsMountedRef = () => {
+export const useIsMountedRef = () => {
   const ref = useRef();
   ref.current = true;
   // if the ref changes, which is the case when the component unmounts
   // set the ref.current to false in order to prevent state updates in
-  // useStateWithRefCheck()
+  // useStateWithMountCheck()
   useEffect(() => {
     return () => {
       ref.current = false;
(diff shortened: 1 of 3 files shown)


gsa/src/web/components/icon/svgicon.js
```diff
@@ -56,6 +56,34 @@ const Styled = styled.span`
   }
 `;
 
+const useIsMountedRef = () => {
+  const ref = useRef();
+  ref.current = true;
+  // if the ref changes, which is the case when the component unmounts
+  // set the ref.current to false in order to prevent state updates in
+  // useStateWithRefCheck()
```
sarahd93:
```suggestion
  // useStateWithMountCheck()
```

swaterkamp: Oops... I'm going to include this in the coming PR update.

PR #2070, opened 2020-03-26
Adjust task icons to work for audits and remove audit icons

sarahd93:
Remove audit icons and use task icons for both tasks and audits. This will reduce unnecessary duplicate code and fix problems with the audit icon tooltips.

**Checklist**:

- [x] Tests


gsa/src/web/pages/audits/__tests__/actions.js
```diff
@@ -229,6 +229,91 @@ describe('Audit Actions tests', () => {
...
+        onAuditStartClick={handleAuditStart}
+        onAuditStopClick={handleAuditStop}
+        onReportDownloadClick={handleReportDownload}
+      />,
+    );
+
+    const icons = getAllByTestId('svg-icon');
+
+    fireEvent.click(icons[0]);
+    expect(handleAuditStart).not.toHaveBeenCalledWith(audit);
```
swaterkamp: Maybe test this for ´´´.not.toHaveBeenCalled()´´´, because it should not have been called with anything instead of just with this specific audit. This would also affect the following tests below.

gsa/src/web/pages/audits/__tests__/actions.js
```diff
@@ -229,6 +229,91 @@ describe('Audit Actions tests', () => {
...
+    expect(icons[1]).toHaveAttribute(
+      'title',
+      'Permission to resume audit denied',
+    );
+
+    fireEvent.click(icons[2]);
+    expect(handleAuditDelete).not.toHaveBeenCalledWith(audit);
+    expect(icons[2]).toHaveAttribute(
+      'title',
+      'Permission to move Audit to trashcan denied',
```
swaterkamp: Maybe adjust the upper case Audit for these icons as well? I know, I pledged for doing it step by step, but I didn't necessarily mean inconsistencies within audit icons (just across entities maybe).

gsa/src/web/pages/tasks/__tests__/actions.js
```diff
@@ -204,6 +204,79 @@ describe('Task Actions tests', () => {
...
+        onTaskResumeClick={handleTaskResume}
+        onTaskStartClick={handleTaskStart}
+        onTaskStopClick={handleTaskStop}
+      />,
+    );
+
+    const icons = getAllByTestId('svg-icon');
+
+    fireEvent.click(icons[0]);
+    expect(handleTaskStart).not.toHaveBeenCalledWith(task);
```
swaterkamp: The comments I made for audits also work for tasks :)

gsa/src/web/pages/audits/__tests__/actions.js
```diff
@@ -314,6 +314,88 @@ describe('Audit Actions tests', () => {
...
+        onAuditStartClick={handleAuditStart}
+        onAuditStopClick={handleAuditStop}
+        onReportDownloadClick={handleReportDownload}
+      />,
+    );
+
+    const icons = getAllByTestId('svg-icon');
+
+    fireEvent.click(icons[0]);
+    expect(handleAuditStop).not.toHaveBeenCalledWith(audit);
```
swaterkamp: Same as for audits and tasks :)

gsa/src/web/pages/tasks/__tests__/actions.js
```diff
@@ -277,6 +277,73 @@ describe('Task Actions tests', () => {
...
+        onTaskResumeClick={handleTaskResume}
+        onTaskStartClick={handleTaskStart}
+        onTaskStopClick={handleTaskStop}
+      />,
+    );
+
+    const icons = getAllByTestId('svg-icon');
+
+    fireEvent.click(icons[0]);
+    expect(handleTaskStop).not.toHaveBeenCalledWith(task);
```
swaterkamp: And again :)

gsa/public/locales/gsa-de.json
```diff
@@ -873,7 +869,7 @@
   "Latest": "Neueste",
   "Latest Identifiers": "Neueste Identifikatoren",
   "Latest Severity": "Neuester Schweregrad",
-  "License: GNU General Public License version 2 or any later version": "Lizenz: GNU General Public License Version 2 oder jede neuere Version",
+  "License: GNU Affero General Public License version 2 or any later version": "Lizenz: GNU Affero General Public License Version 2 oder jede neuere Version",
```
swaterkamp:
```suggestion
  "License: GNU Affero General Public License version 3 or any later version": "Lizenz: GNU Affero General Public License Version 3 oder jede neuere Version",
```

gsa/src/web/pages/audits/__tests__/actions.js
sarahd93: Inconsistencies across entities are not possible because delete, edit etc are always the entity icons that we use for every entity. That was precisely the reason I asked because if we change one type of icon we have to change all of them to avoid inconsistencies on the same page. But it might be better to do that in a separate PR.

gsa/public/locales/gsa-de.json
```diff
@@ -1615,12 +1607,8 @@
   "Task List": "Aufgabenliste",
   "Task Name": "Name der Aufgabe",
   "Task Wizard": "Aufgaben-Wizard",
-  "Task is a container": "Aufgabe ist eine Container-Aufgabe",
```
swaterkamp: Is this really not needed anymore? Can an Audit be a container?

gsa/src/web/pages/audits/__tests__/actions.js
swaterkamp: Ok, that's fine with me :)

PR #2076, opened 2020-03-31
Task dashboards do not crash

saberlynx:
**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/components/dashboard/dashboard.js
+++ b/gsa/src/web/components/dashboard/dashboard.js
@@ -35,6 +35,7 @@ import {
 
 import {isDefined} from 'gmp/utils/identity';
 import {excludeObjectProps} from 'gmp/utils/object';
+import Setting from 'gmp/models/setting';
 
 import {
   loadSettings,
@@ -62,10 +63,9 @@ import withGmp from 'web/utils/withGmp';
 import compose from 'web/utils/compose';
 
 import {getDisplay} from './registry';
-import {getRows as get_rows} from './utils';
+import {getRows as getRowsUtil} from './utils';
 
 import {useGetSetting} from 'web/utils/useGetSettings';
-import Setting from 'gmp/models/setting';
 
 const log = Logger.getLogger('web.components.dashboard');
 
@@ -96,7 +96,7 @@ const RowPlaceHolder = styled.div`
 
 export const Dashboard = props => {
   const getSetting = useGetSetting();
-  const {data, loading} = getSetting({
+  const {data, loading: isLoading} = getSetting({
     userSettingId: '3d5db3c7-5208-4b47-8c28-48efc621b1e0',
   });
   const {permittedDisplays = []} = props;
@@ -195,7 +195,7 @@ export const Dashboard = props => {
   };
 
   const getRows = defaultRows => {
-    return get_rows(props.settings, defaultRows);
+    return getRowsUtil(props.settings, defaultRows);
   };
 
   const getDisplayState = id => {
@@ -254,7 +254,6 @@ export const Dashboard = props => {
 
   const {
     error,
-    isLoading,
     maxItemsPerRow = DEFAULT_MAX_ITEMS_PER_ROW,
     maxRows = DEFAULT_MAX_ROWS,
     ...others
@@ -267,7 +266,7 @@ export const Dashboard = props => {
   } else {
     rows = getRows();
   }
-  if (isDefined(error) && !loading && !isDefined(dashboardSettings)) {
+  if (isDefined(error) && !isLoading && !isDefined(dashboardSettings)) {
     return (
       <RowPlaceHolder>
         {_('Could not load dashboard settings. Reason: {{error}}', {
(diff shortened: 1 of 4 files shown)


gsa/src/web/components/dashboard/dashboard.js
```diff
@@ -61,7 +61,11 @@ import withGmp from 'web/utils/withGmp';
 import compose from 'web/utils/compose';
 
 import {getDisplay} from './registry';
-import {getRows} from './utils';
+import {getRows as get_rows} from './utils';
+
+import {useGetSetting} from 'web/pages/tasks/graphql';
+import Setting from 'gmp/models/setting';
+import {convertLoadedSettings} from 'gmp/commands/dashboards';
```
swaterkamp: Sorry for nitpicking, but could you sort the imports according to our convention? I know, this is a POC, but we will probably use this code later on as well. Therefore, we should tidy it up and make clean commits to save us more work in the future. Better make it nice from the beginning than spending resources on changing stuff later.

gsa/src/web/components/dashboard/dashboard.js
```diff
@@ -61,7 +61,11 @@ import withGmp from 'web/utils/withGmp';
 import compose from 'web/utils/compose';
 
 import {getDisplay} from './registry';
-import {getRows} from './utils';
+import {getRows as get_rows} from './utils';
```
swaterkamp, line 65: Does this mean, "camelCase API" is not a thing anymore?

bjoernricks: snake_case should not be a thing anymore! All our code use camelCase now. All old code needs to be converted to use camelCase.

swaterkamp: Why the ... as ... import than? Is this necessary at this place? I mean, even in the old code we have one location where we will always find snake_case, namely in the commands...

gsa/src/web/components/dashboard/dashboard.js
```diff
@@ -96,7 +96,7 @@ const RowPlaceHolder = styled.div`
 
 export const Dashboard = props => {
   const getSetting = useGetSetting();
-  const {data} = getSetting({
+  const {data, loading} = getSetting({
```
swaterkamp, line 99: If this is a boolean, it should be named ´´´isLoading´´´.

gsa/src/web/components/dashboard/dashboard.js
saberlynx: It's necessary because we have a method called getRows that calls getRows and react returns an error. It wasn't a problem in a class component. Not sure why, but I suspect it's because in class components it's technically "this.getRows". I can name it something else, but right now it looks like it has to be named differently from the internal dashboard function getRows.

gsa/src/web/components/dashboard/dashboard.js
saberlynx: Dashboard gets an isLoading prop already, passed from a lot of other dashboards that do not have the new backend. I don't know what will happen if I touched that right now. loading is an Apollo exported variable.

gsa/src/web/components/dashboard/dashboard.js
saberlynx: Of course.

gsa/src/web/components/dashboard/dashboard.js
swaterkamp: Then it should be named differently without snake_case. I find it suboptimal, though, that we seemingly had two getRows in one component - one being a function in utils and one being a class method calling this function. Their naming should have been a little more precise from the beginning on.

swaterkamp: Just saw that you already made changes :) Thanks!

gsa/src/web/utils/useGetSettings.js
```diff
@@ -0,0 +1,53 @@
...
+
+import {toFruitfulQuery} from 'web/utils/graphql';
+
+export const GET_SETTING = gql`
+  query UserSetting($userSettingId: UUID!) {
+    userSetting(userSettingId: $userSettingId) {
+      uuid
+      name
+      value
+      comment
```
bjoernricks, line 31: Not sure if we really need the comment. It is not very useful.

gsa/src/web/components/dashboard/dashboard.js
```diff
@@ -90,39 +94,31 @@ const RowPlaceHolder = styled.div`
...
-    });
-
-    this.handleItemsChange = this.handleItemsChange.bind(this);
-    this.handleRowResize = this.handleRowResize.bind(this);
-    this.handleUpdateDisplay = this.handleUpdateDisplay.bind(this);
-    this.handleRemoveDisplay = this.handleRemoveDisplay.bind(this);
-
-    this.getDisplaysById = memoize(rows => getDisplaysById(rows));
-  }
+export const Dashboard = props => {
```
bjoernricks, line 97: I would love if we can destruct all used props already here in the Dashboard function definition.

```javascript
const Dashboard = ({
    permittedDisplays = [],
    defaultDisplays,
   maxItemsPerRow = DEFAULT_MAX_ITEMS_PER_ROW,
   ...
}) => {
```

but i guess it is something for another pr.

saberlynx:
Due to recent API changes, there is another PR being prepared to make `hyperion-gsa` work with it. It would be best to wait until that PR is merged before we merge this one to visually verify that the code will work the same way.

PR #2099, opened 2020-04-14
Improve BPM

swaterkamp:
This includes the following improvements:
- Invert coloring of the active state of the ConditionalColorizationIcon
- Implement a maximum of 50 nodes
- Increase clickable area of edges to 25px width
- Disable zoom icons if max or min zoom is reached

**Checklist**:

- [x] Tests


gsa/src/web/components/processmap/tools.js
```diff
@@ -54,11 +54,12 @@ const IconWrapper = styled.div`
   justify-content: space-around;
   padding: 10px 0 10px 0;
   & svg path {
-    fill: ${Theme.black};
+    fill: ${props => (props.inactive ? Theme.inputBorderGray : Theme.black)};
```
bjoernricks, line 59: Should be isInActive or even better use isActive

swaterkamp: It's active by default, so I chose it to be inactive that needs to be explicitly set. I'd go with isInactive, though.

swaterkamp: Plus, it's a different prop than the isActive used for DrawEdge und ConditionalColorization.

PR #2110, opened 2020-04-21
Update gsa to reflect hyperion api changes

saberlynx:
**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/gmp/models/scanner.js
+++ b/gsa/src/gmp/models/scanner.js
@@ -99,10 +99,12 @@ class Scanner extends Model {
       ret.id = element.uuid;
     }
 
-    if (hasValue(element.scannerType)) {
-      ret.scannerType = scannerTypeInt(element.scannerType);
+    // scanner.type can be one of three formats:
+    // integer 2 (for scanners), '2' (audits) and 'OPENVAS_SCANNER_TYPE' (hyperion)
+    if (isString(element.type) && element.type.length > 2) {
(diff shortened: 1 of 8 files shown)


gsa/src/gmp/models/scanner.js
```diff
@@ -99,10 +99,12 @@ class Scanner extends Model {
       ret.id = element.uuid;
     }
 
-    if (hasValue(element.scannerType)) {
-      ret.scannerType = scannerTypeInt(element.scannerType);
+    // scanner.type can be one of three formats:
+    // integer 2 (for scanners), '2' (audits) and 'OPENVAS_SCANNER_TYPE' (hyperion)
+    if (isString(element.type) && element.type.length > 2) {
```
saberlynx, line 104: maybe we'll have more scanner types in the future but i don't anticipate more than 99...

gsa/src/gmp/models/scanner.js
```diff
@@ -99,10 +99,12 @@ class Scanner extends Model {
       ret.id = element.uuid;
     }
 
-    if (hasValue(element.scannerType)) {
-      ret.scannerType = scannerTypeInt(element.scannerType);
+    // scanner.type can be one of three formats:
+    // integer 2 (for scanners), '2' (audits) and 'OPENVAS_SCANNER_TYPE' (hyperion)
```
bjoernricks, line 103: No take a look at https://github.com/greenbone/gsa/blob/1eaae04e254ab32facb4585797e4828917be70e0/gsa/src/gmp/models/scanner.js#L30-L34 for the meaning of the type int. And that's why I find string constants better then integers :-)

PR #2115, opened 2020-04-22
Fix alterable tasks not listed as alterable

sarahd93:
The alterable state for tasks is not correctly parsed by hyperion-gsa yet. Fix parsing alterable to be able to render alterable icons and change the alterable state in the edit dialog.

**Checklist**:

- [x] Tests


gsa/src/gmp/models/task.js
```diff
@@ -148,8 +147,7 @@ class Task extends Model {
       copy.reportCount.total = parseInt(reportCount.total);
       copy.reportCount.finished = parseInt(reportCount.finished);
     }
-
-    copy.alterable = parseYesNo(object.alterable);
+    copy.alterable = object.alterable === true ? YES_VALUE : NO_VALUE;
```
bjoernricks, line 150: Oh f**k we should change YES_VALUE to true and NO_VALUE to false asap. But nevertheless it is better to change parseYesNo first here. Could you adjust parseYesNo that it returns YES_VALUE for true instead?

```javascript
export const parseYesNo = value =>
  value === '1' || value === 1 || value === true ? YES_VALUE : NO_VALUE;
```
and adjust the tests too? I am really sure this will fail at other places besides alterable too if we don't change parseYesNo.

bjoernricks:
Thanks a lot!

PR #2125, opened 2020-04-29
Auto logout

bjoernricks:
Clear the redux store and remove possible gsad session token if hyperion's session has ended.


bjoernricks:
> Out of curiosity, do you know if `logoutLink.concat(httpLink)` is meaningful? Does the order of concatenation matter? Is there a difference between that and `httpLink.concat(logoutLink)`?

Yes there is a difference. It is nearly the same as the middleware in django and graphene. See https://www.apollographql.com/docs/link/overview/ for the details :-)

PR #2136, opened 2020-05-04
Auto session logout

bjoernricks:
Refactor 401 response handling for auto logout of current user.


gsa/src/web/authorized.js
```diff
@@ -15,108 +15,86 @@
...
+
+      if (!isAuthenticated) {
+        toLoginPage();
+      }
+    }
+    setIsLoading(loading);
+  }, [loading, data]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  if (isLoading) {
+    // to prevent infinite loading if there is an error
```
saberlynx, line 88: I think these comments are no longer relevant and might be confusing at a later date.

PR #2148, opened 2020-05-07
Update license and link on AboutPage

swaterkamp:
- Adjust link to license text to lead to AGPL 3.0+
- Fix closing the dialog when clicking resume

**Checklist**:

- [X] Tests

### DIFF
--- a/gsa/src/web/components/link/__tests__/externallink.js
+++ b/gsa/src/web/components/link/__tests__/externallink.js
@@ -62,6 +62,37 @@ describe('ExternalLink tests', () => {
     expect(dialog).toBeTruthy();
   });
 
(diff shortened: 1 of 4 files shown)


gsa/src/web/components/link/__tests__/externallink.js
```diff
@@ -62,6 +62,33 @@ describe('ExternalLink tests', () => {
     expect(dialog).toBeTruthy();
   });
 
+  test('should close confirmation dialog on resume click', () => {
+    window.open = jest.fn();
```
bjoernricks, line 66: You need to restore the open function after the test. Otherwise this may have side effects.

PR #2151, opened 2020-05-08
Task details shows source interface

saberlynx:
**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/pages/tasks/details.js
+++ b/gsa/src/web/pages/tasks/details.js
@@ -94,15 +94,14 @@ class TaskDetails extends React.Component {
       inAssets,
       reports,
       minQod,
-      preferences,
       scanner,
       target,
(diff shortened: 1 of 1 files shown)


gsa/src/web/pages/tasks/details.js
```diff
@@ -94,15 +94,14 @@ class TaskDetails extends React.Component {
       inAssets,
       reports,
       minQod,
-      preferences,
       scanner,
       target,
       maxChecks,
       maxHosts,
       schedule,
+      sourceIface = '',
```
bjoernricks, line 102: Could we rename that variable to `sourceInterface` in another PR? also in hyperion?

saberlynx: of course. when i first saw that variable i was a little confused too.

PR #2157, opened 2020-05-11
Hyperion renew session

bjoernricks:
This is a first start for renewing the user session via hyperion.


gsa/src/web/components/menu/__tests__/usermenu.js
```diff
@@ -81,46 +75,71 @@ describe('UserMenu component tests', () => {
...
+    const renewDate = '2019-10-10T12:00:00Z';
+    const [queryMock, resultFunc] = createRenewSessionQueryMock(renewDate);
+
+    const renewSession = jest.fn().mockResolvedValue({data: renewDate});
     const gmp = {
       user: {
         renewSession,
       },
+      settings: {
+        isHyperionOnly: false,
```
saberlynx, line 126: I think this is `enableHyperionOnly` in the gmpsettings

bjoernricks: Ups. Should be fixed now. Wasn't the only line...

PR #2170, opened 2020-05-15
Make reload via GMP optional for list pages

bjoernricks:
Before it was required to pass a loadEntities function to
withEntitiesContainer. This commit changes the withEntitiesContainer to
make the loadEntities function optional. As a result not passing a
loadEntities functions deactivates the reloading via the Reload
component.

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/entities/withEntitiesContainer.js
+++ b/gsa/src/web/entities/withEntitiesContainer.js
@@ -107,10 +107,13 @@ const withEntitiesContainer = (
       loadedFilter = filter;
     }
 
+    const reloadFunc = isDefined(loadEntities)
+      ? (newFilter = filter) => loadEntities(newFilter)
+      : undefined;
     return (
       <Reload
         reloadInterval={() => reloadInterval(props)}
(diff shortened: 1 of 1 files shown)


gsa/src/web/entities/withEntitiesContainer.js
```diff
@@ -107,10 +107,13 @@ const withEntitiesContainer = (
...
     }
 
+    const reloadFunc = isDefined(loadEntities)
+      ? (newFilter = filter) => loadEntities(newFilter)
+      : undefined;
     return (
       <Reload
         reloadInterval={() => reloadInterval(props)}
-        reload={(newFilter = filter) => loadEntities(newFilter)}
+        reload={reloadFunc}
```
saberlynx, line 116: If reloadFunc is undefined, would that cause a problem with the Reload component?

bjoernricks: Nope. See https://github.com/greenbone/gsa/blob/hyperion/gsa/src/web/components/loading/reload.js#L223 no reloading is done if the function is undefined.

PR #2171, opened 2020-05-15
Introduce a usePageFilter custom hook

bjoernricks:
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/entities/filterprovider.js
+++ b/gsa/src/web/entities/filterprovider.js
@@ -16,126 +16,22 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, {useEffect, useState} from 'react';
-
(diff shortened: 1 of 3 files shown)


bjoernricks:
I've chosen `usePageFilter` because it resolves the filter for a page. I've considered `useFilter` - i think it is to generic for this purpose - and `useCurrentFilter`, which is not really correct. Personally I also don't like `useConditionalFilter` because there is no hook the filter can influence.

swaterkamp:
OK, the classic problem of "naming things" is always a matter of taste, I guess. I just think that, with the given FilterProvider, usePageFilter reduces what we think the hook does to just one of several possible filters.
As I'm lacking a better idea other than those mentioned above, I'd just approve for now. If someone has a better idea later we can still rename it.

PR #2176, opened 2020-05-18
Hyperion renew session #2

bjoernricks:

### DIFF
--- a/gsa/src/web/pages/credentials/component.js
+++ b/gsa/src/web/pages/credentials/component.js
@@ -15,9 +15,9 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
+import React, {useCallback, useState, useEffect} from 'react';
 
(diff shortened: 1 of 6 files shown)


gsa/src/web/pages/credentials/component.js
```diff
@@ -26,192 +26,196 @@ import {shorten} from 'gmp/utils/string';
...
+  onInstallerDownloadError,
+  onInstallerDownloaded,
+  onSaved,
+  onSaveError,
+}) => {
+  const dispatch = useDispatch();
+  const userDefaultsSelector = useSelector(getUserSettingsDefaults);
+  const userName = useUserName();
+  const gmp = useGmp();
+  const [, renewSessionTimeout] = useUserSessionTimeout();
```
saberlynx, line 61: In our previous PR I think we called this function `renewSession`, though I don't have particularly strong feelings on this matter.

bjoernricks: It's just a variable name here. We could have called it foo here too ;-)

PR #2180, opened 2020-05-19
Implement graphql create and modify targets

saberlynx:
**Checklist**:

- [x] Tests


gsa/src/web/pages/targets/component.js
```diff
@@ -15,406 +15,479 @@
...
+      setInUse(entity.isInUse());
+      setName(entity.name);
+      setPort(
+        isDefined(entity.ssh_credential) ? entity.ssh_credential.port : '22',
+      );
+      setReverseLookupOnly(entity.reverse_lookup_only);
+      setReverseLookupUnify(entity.reverse_lookup_unify);
+      setTargetSource('manual');
+      setTargetExcludeSource('manual');
+      setTargetTitle(_('Edit Target {{name}}', entity));
```
bjoernricks, line 129: Sadly putting every dialog state value into an own use state variable has a serious side effect. Every call to a setXyz function will cause a single re-render. Therefore either [useReducer](https://reactjs.org/docs/hooks-reference.html#usereducer) or a single state object must be used. E.g. i've used something like
```js
const [dialogState, setDialogState] = useState({foo: initialvalue, bar: initialvalue});

....
openDialog = () => {
setDialogState(state => {...state, xyzValue: 1234, foo: '345});
```
}

gsa/src/web/pages/targets/component.js
```diff
@@ -15,406 +15,479 @@
...
+          reverseLookupOnly: reverse_lookup_only,
+          reverseLookupUnify: reverse_lookup_unify,
+          smbCredentialId: smb_credential_id,
+          snmpCredentialId: snmp_credential_id,
+          sshCredentialId: ssh_credential_id,
+        };
+
+        return createTarget(mutationData);
+      })
+      .then(result => onCreated(result), onCreateError)
```
bjoernricks, line 389: can be simplified if the calling function/method passes to correct arguments already.
```suggestion
      .then(onCreated, onCreateError)
```

gsa/src/web/pages/targets/component.js
```diff
@@ -15,406 +15,479 @@
...
+      .then(() => closeTargetDialog());
```
bjoernricks, line 390: same here.
```suggestion
      .then(closeTargetDialog);
```

gsa/src/web/utils/render.js
```diff
@@ -31,6 +31,8 @@ import {typeName, getEntityType} from 'gmp/utils/entitytype';
 export const UNSET_VALUE = '0';
 export const UNSET_LABEL = '--';
 
+export const NULL_VALUE = null;
```
bjoernricks, line 34: Hmm maybe we should just redefine UNSET_VALUE to null? This could possible break gsad based gmp but who cares ;-)

gsa/src/web/pages/targets/component.js
```diff
@@ -56,43 +56,47 @@ const TargetComponent = props => {
   const [portListDialogVisible, setPortListDialogVisible] = useState(false);
   const [targetDialogVisible, setTargetDialogVisible] = useState(false);
 
+  const [targetDialogState, setTargetDialogState] = useState({
+    id: undefined,
```
bjoernricks, line 60: You can skip the undefined object properties if you want. It is not necessary to initialize these properties. They are undefined by default.

PR #2189, opened 2020-05-25
Renew session on refresh

saberlynx:
**Checklist**:

- [ ] Tests


gsa/src/web/authorized.js
```diff
@@ -60,12 +62,14 @@ const Authorized = ({children}) => {
     if (hasValue(isAuthenticated)) {
       dispatch(setIsLoggedIn(isAuthenticated));
 
-      if (!isAuthenticated) {
+      if (isAuthenticated) {
+        renewSession();
```
bjoernricks, line 66: I fear this line gets called to often. I am really not sure about that. It really should only be called once.

PR #2192, opened 2020-05-26
Use gql renewSession for trashcan page and fix trashcan task parsing

saberlynx:
**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/gmp/model.js
+++ b/gsa/src/gmp/model.js
@@ -104,15 +104,8 @@ class Model {
       delete copy.in_use;
     }
 
-    if (isDefined(element.owner)) {
-      if (this.entityType === 'task') {
-        // hyperion tasks don't have owner.name
-        copy.owner = element.owner?.name;
-      }
-
-      if (isEmpty(element.owner.name)) {
-        delete copy.owner;
-      }
+    if (isDefined(element.owner) && isEmpty(element.owner.name)) {
+      delete copy.owner;
     }
 
     copy.summary = parseText(element.summary);
--- a/gsa/src/gmp/model.js
+++ b/gsa/src/gmp/model.js
@@ -105,13 +105,13 @@ class Model {
     }
 
     if (isDefined(element.owner)) {
-      if (isEmpty(element.owner.name)) {
-        delete copy.owner;
-      }
-
       if (this.entityType === 'task') {
         // hyperion tasks don't have owner.name
-        copy.owner = element.owner.name;
+        copy.owner = element.owner?.name;
+      }
+
+      if (isEmpty(element.owner.name)) {
+        delete copy.owner;
       }
     }
 
--- a/gsa/src/gmp/model.js
+++ b/gsa/src/gmp/model.js
@@ -65,7 +65,6 @@ class Model {
 
   static parseElement(element = {}) {
     const copy = parseDefaultProperties(element);
-
     if (isDefined(element.end_time)) {
       if (element.end_time.length > 0) {
         copy.endTime = parseDate(element.end_time);
@@ -105,8 +104,15 @@ class Model {
       delete copy.in_use;
     }
 
-    if (isDefined(element.owner) && isEmpty(element.owner.name)) {
-      delete copy.owner;
+    if (isDefined(element.owner)) {
+      if (isEmpty(element.owner.name)) {
+        delete copy.owner;
+      }
+
+      if (this.entityType === 'task') {
+        // hyperion tasks don't have owner.name
+        copy.owner = element.owner.name;
+      }
     }
 
     copy.summary = parseText(element.summary);
@@ -120,7 +126,6 @@ class Model {
     if (isEmpty(element.comment)) {
       delete copy.comment;
     }
-
     return copy;
   }
 
--- a/gsa/src/web/pages/extras/trashcanpage.js
+++ b/gsa/src/web/pages/extras/trashcanpage.js
@@ -151,9 +151,7 @@ const Trashcan = () => {
       return gmp.trashcan
         .restore(entity)
         .then(getTrash)
-        .catch(error => {
-          setError(error);
-        });
+        .catch(setError);
     },
     [getTrash, gmp.trashcan, handleInteraction],
   );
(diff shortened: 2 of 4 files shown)


gsa/src/web/pages/tasks/row.js
```diff
@@ -110,11 +110,11 @@ const Row = ({
     if (hasValue(observers.users)) {
       obs.user = _('Users {{user}}', {user: observers.users.join(', ')});
     }
-    if (observers.roles.length > 0) {
+    if (observers?.roles?.length > 0) {
```
bjoernricks, line 113: Would be better to use
```suggestion
    if (isDefined(observers?.roles?.length) && observers.roles.length > 0) {
```
to not rely on js' behavior of comparing undefined with 0.

gsa/src/web/pages/tasks/row.js
```diff
@@ -110,11 +110,11 @@ const Row = ({
...
       const role = observers.roles.map(r => r.name);
       obs.role = _('Roles {{role}}', {role: role.join(', ')});
     }
-    if (observers.groups.length > 0) {
+    if (observers?.groups?.length > 0) {
```
bjoernricks, line 117:
```suggestion
    if (isDefined(observers?.roles?.length) && observers.roles.length > 0) {
```

gsa/src/gmp/model.js
```diff
@@ -105,8 +104,15 @@ class Model {
       delete copy.in_use;
     }
 
-    if (isDefined(element.owner) && isEmpty(element.owner.name)) {
-      delete copy.owner;
+    if (isDefined(element.owner)) {
+      if (this.entityType === 'task') {
```
bjoernricks, line 108: I really don't want such if checking in the generic model class. Why is this needed? Maybe just parse the owner in the task model class before calling super or just re-parse it from the original element.

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -15,9 +15,8 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
-
-import {connect} from 'react-redux';
+/* eslint-disable no-shadow */
```
bjoernricks, line 18: Please don't disable no-shadow for the whole file. Shadowing a variable may not be obvious sometimes and can lead to strange bugs.

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -119,105 +115,85 @@ const separateByUsageType = inputList => {
...
-  componentDidMount() {
-    this.getTrash();
-  }
+  const [trash, setTrash] = useState({});
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState();
 
-  getTrash() {
-    const {gmp} = this.props;
+  const getTrash = useCallback(() => {
```
bjoernricks, line 126: Should be renamed to something more meaningful like loadTrashCanData or something like that.

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -119,105 +115,85 @@ const separateByUsageType = inputList => {
...
-      .then(this.getTrash)
-      .catch(error => {
-        this.setState({
-          error: error,
+      return gmp.trashcan
+        .restore(entity)
+        .then(getTrash)
+        .catch(error => {
+          setError(error);
         });
```
bjoernricks, line 156: Can be simplified if you want.
```suggestion
        .catch(setError);
```

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -119,105 +115,85 @@ const separateByUsageType = inputList => {
...
+  const handleDelete = useCallback(
+    entity => {
+      handleInteraction();
+
+      return gmp.trashcan
+        .delete(entity)
+        .then(getTrash)
+        .catch(error => {
+          setError(error);
         });
```
bjoernricks, line 170:
```suggestion
        .catch(setError);
```

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -119,105 +115,85 @@ const separateByUsageType = inputList => {
...
+  );
 
-  handleEmpty() {
-    const {gmp} = this.props;
-    this.handleInteraction();
+  const handleEmpty = useCallback(() => {
+    handleInteraction();
 
-    this.setState({loading: true});
+    setLoading(true);
```
bjoernricks, line 178: I am confused why loading is set during emptying the trash but not when loading the trash data.

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -226,344 +202,312 @@ class Trashcan extends React.Component {
...
+        </TableBody>
+      );
+    },
+    [createContentRow],
+  );
+
+  if (!isDefined(trash) && !isDefined(error)) {
+    return <Loading />;
+  } else if (!isDefined(trash) && isDefined(error)) {
+    setTrash({});
```
bjoernricks, line 313: I don't think this works. You are changing the state during a react render phase.

gsa/src/web/pages/extras/trashcanpage.js
saberlynx: Carelessness. I'll add that in.

gsa/src/web/pages/extras/trashcanpage.js
saberlynx: I have set it to watch for `loading` instead. trash should always be defined since it's initialized as `{}` and never set to undefined in code.

gsa/src/web/pages/extras/trashcanpage.js
```diff
@@ -119,105 +114,81 @@ const separateByUsageType = inputList => {
...
-  }
-
-  handleEmpty() {
-    const {gmp} = this.props;
-    this.handleInteraction();
+  const handleEmpty = useCallback(() => {
+    handleInteraction();
 
-    this.setState({loading: true});
+    setLoading(true);
```
bjoernricks, line 173: I still don't get this loading case. Why is loading set to true until the loading is actually started? Maybe just remove setting loading here in handleEmpty

PR #2199, opened 2020-05-28
Allow tasks selection at list page

bjoernricks:


gsa/src/web/utils/useSelection.js
```diff
@@ -0,0 +1,62 @@
...
+) => {
+  const [selected, setSelected] = useState(
+    initialSelectionType === SelectionType.SELECTION_USER ? [] : undefined,
+  );
+  const [selectionType, setSelectionType] = useState(initialSelectionType);
+
+  const select = useCallback(obj => {
+    // ensure the using component gets re-rendered by creating new array
+    setSelected(prevSelected =>
+      isDefined(prevSelected) && !prevSelected.includes(obj)
```
swaterkamp, line 34: As we have not officially retired IE11 support, do we need to take care of that? Because in that case, Array.protoype.includes needs a polyfill

gsa/src/web/utils/useSelection.js
```diff
@@ -0,0 +1,62 @@
...
+        ? [...prevSelected, obj]
+        : undefined,
+    );
+  }, []);
+
+  const deselect = useCallback(obj => {
+    // ensure the using component gets re-rendered by creating new array
+    setSelected(prevSelected => {
+      if (isDefined(prevSelected) && prevSelected.includes(obj)) {
```
swaterkamp, line 43: Same here

gsa/src/web/utils/useSelection.js
bjoernricks: Oh thanks for the reminder! I've thought about that during development and of course completely forgot it.

PR #2225, opened 2020-06-10
Enable graphql bulk delete for tasks

saberlynx:
**Checklist**:

- [x] Tests


gsa/src/web/pages/tasks/listpage.js
```diff
@@ -156,15 +166,29 @@ const TasksListPage = () => {
...
+
   useEffect(() => {
     // load tasks initially after the filter is resolved
     if (!isLoadingFilter && hasValue(filter) && !called) {
       getTasks({
         filterString: filter.toFilterString(),
         first: filter.get('rows'),
       });
+
+      getAllFiltered({filterString: filter.all().toFilterString()});
```
bjoernricks, line 189: Oh just saw this line. This isn't a good idea. It will always load **ALL** tasks.

bjoernricks: We need to load all tasks only on demand. And only if the selection was filter related. Maybe it is better to implement a deleteAllTasks graphql API to make this easier?

saberlynx: I thought this just sends the current filter with rows=-1?

saberlynx: I'll try this with a useEffect hook. I tried for hours to fetch this only in case 3 of `handleBulkDeleteTask` but don't think the data can be used in the same render, or something.

gsa/src/web/pages/tasks/listpage.js
```diff
@@ -156,6 +166,25 @@ const TasksListPage = () => {
...
+  }, [selectionType, filter, getAllFiltered]);
+
+  const handleBulkDeleteTask = () => {
+    let tasksToDelete;
+    if (selectionType === SelectionType.SELECTION_USER) {
+      tasksToDelete = getEntityIds(selected);
+    } else if (selectionType === SelectionType.SELECTION_PAGE_CONTENTS) {
+      tasksToDelete = getEntityIds(tasks);
+    } else {
+      tasksToDelete = getEntityIds(allFilteredTasks);
```
bjoernricks, line 183: Much better already. But this could result in a race condition when loading of all filtered tasks takes a while. In that case getEntityIds would be called with an empty or old allFilteredTasks variable. I see two possible solutions:

1. query getAllFiltered when clicking on the bulk delete and using the returned promise
2. implement something like deleteAllTasksByFilter in the backend

gsa/src/web/graphql/__tests__/tasks.js
```diff
@@ -44,8 +45,10 @@ import {
   createStartTaskQueryMock,
   createStopTaskQueryMock,
   createDeleteTasksQueryMock,
+  createDeleteFilteredTasksQueryMock,
 } from '../__mocks__/tasks';
 import {isDefined} from 'gmp/utils/identity';
+import {render} from 'react-dom';
```
bjoernricks: That import is wrong. The render func is returned from rendererWith()

saberlynx:
> Eslint complains about something and the render import is wrong.

Yeah, sometimes when I try autocomplete it imports something for me.

gsa/src/web/utils/selectiontype.js
```diff
@@ -22,6 +22,8 @@ const SelectionType = {
   SELECTION_FILTER: '2',
 };
 
+export const getEntityIds = entityArray => entityArray.map(entity => entity.id);
```
bjoernricks, line 25:
```suggestion
export const getEntityIds = (entityArray = []) => entityArray.map(entity => entity.id);
```

bjoernricks: Would be good to get a simple unit test for this function too.

PR #2227, opened 2020-06-11
Improve queued status

sarahd93:
**Checklist**:

- [ ] Tests


gsa/src/web/components/bar/statusbar.js
```diff
@@ -80,15 +81,15 @@ const StatusBar = ({status = 'Unknown', progress = '0'}) => {
...
   } else if (status === TASK_STATUS.running) {
     background = 'run';
-  } else if (status === TASK_STATUS.queued) {
-    background = 'queued';
   }
+
+  const title =
+    status === TASK_STATUS.queued
+      ? 'Task is queued for scanning'
+      : getTranslatableTaskStatus(status);
```
bjoernricks, line 89: Any reason why you didn't extend getTranslatableTaskStatus? At least `'Task is queued for scanning'` needs to be translateable. Therefore it should be `_('Task is queued for scanning')`

sarahd93: Oh right, I have to make it translatable. Not sure about extending getTranslatableTaskStatus, though. That would mean that the text for the queued status would be "Task is queued for scanning" everywhere even in the statusbar itself and in the displays. I think we should stick to single words for the status there. Anything longer would probably not even fit in the space properly.

bjoernricks: Ah maybe you are right. We want only the translation of Queued at some places and the task status here should be longer for the title to have a better explanation.

PR #2239, opened 2020-06-15
GraphQL create and modify tag

saberlynx:
Supports remove, create, edit and toggle tags.
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/graphql/__tests__/tags.js
+++ b/gsa/src/web/graphql/__tests__/tags.js
@@ -18,54 +18,18 @@
  */
 
 import React, {useState} from 'react';
+import {
+  createTagInput,
+  modifyTagInput,
+  createCreateTagMock,
+  createModifyTagMock,
+} from '../__mocks__/tags';
 
 import Button from 'web/components/form/button';
 
 import {rendererWith, fireEvent, wait, screen} from 'web/utils/testing';
 
-import {CREATE_TAG, MODIFY_TAG, useCreateTag, useModifyTag} from '../tags';
-
-const createTagInput = {
-  name: 'foo',
-  resourceType: 'OPERATING_SYSTEM',
-};
-
-const modifyTagInput = {
-  id: '12345',
-};
-
-const createTagResult = jest.fn().mockReturnValue({
-  data: {
-    createTag: {
-      id: '12345',
-      status: 200,
-    },
-  },
-});
-
-const modifyTagResult = jest.fn().mockReturnValue({
-  data: {
-    modifyTag: {
-      ok: true,
-    },
-  },
-});
-
-const createTagMock = {
-  request: {
-    query: CREATE_TAG,
-    variables: {input: createTagInput},
-  },
-  newData: createTagResult,
-};
-
-const modifyTagMock = {
-  request: {
-    query: MODIFY_TAG,
-    variables: {input: modifyTagInput},
-  },
-  newData: modifyTagResult,
-};
+import {useCreateTag, useModifyTag} from '../tags';
 
 const TestComponent = () => {
   const [notification, setNotification] = useState('');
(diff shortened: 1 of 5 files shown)


gsa/src/web/graphql/tags.js
```diff
@@ -0,0 +1,135 @@
...
+  result: 'RESULT',
+  role: 'ROLE',
+  scanconfig: 'SCAN_CONFIG',
+  scanner: 'SCANNER',
+  schedule: 'SCHEDULE',
+  target: 'TARGET',
+  task: 'TASK',
+  tlscertificate: 'TLS_CERTIFICATE',
+  user: 'USER',
+};
```
bjoernricks, line 55: Does a list like this not exist yet?

gsa/src/web/graphql/tags.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+`;
+
+export const useCreateTag = options => {
+  const [queryCreateTag, {data, ...other}] = useMutation(CREATE_TAG, options);
+  const createTag = useCallback(
+    // eslint-disable-next-line no-shadow
+    (inputObject, options) =>
+      queryCreateTag({...options, variables: {input: inputObject}}),
```
bjoernricks, line 76: It's best to return the same data from the promise as it is returned from the hook. 
```suggestion
      queryCreateTag({...options, variables: {input: inputObject}}).then(result => result.data.createTag.id),
```

gsa/src/web/graphql/tags.js
saberlynx: I looked and saw nothing that looks exactly like this. For instance I found one called `CMD_TYPE` in `gmp/utils/entitytype.js` but it has `scanconfig: config` instead of `scanconfig: SCAN_CONFIG`. there IS a similar list named `ENTITY_TYPE`, so I should probably change the name of this list.

gsa/src/web/graphql/tags.js
```diff
@@ -0,0 +1,135 @@
...
+    [queryModifyTag],
+  );
+  return [modifyTag, data];
+};
+
+export const useToggleTag = () => {
+  const [queryModifyTag, data] = useMutation(MODIFY_TAG);
+  const enableTag = useCallback(
+    // eslint-disable-next-line no-shadow
+    tag => queryModifyTag({variables: {input: {id: tag.id, active: 1}}}),
```
bjoernricks, line 106: Should use true and false instead of 1 and 0. We should use a clean API. Booleans should always be true and false.

bjoernricks, line 106:
```suggestion
    tag => queryModifyTag({variables: {input: {id: tag.id, active: true}}}),
```

gsa/src/web/graphql/tags.js
```diff
@@ -0,0 +1,135 @@
...
+    [queryModifyTag],
+  );
+  const disableTag = useCallback(
+    // eslint-disable-next-line no-shadow
+    tag => queryModifyTag({variables: {input: {id: tag.id, active: 0}}}),
```
bjoernricks, line 111:
```suggestion
    tag => queryModifyTag({variables: {input: {id: tag.id, active: false}}}),
```

gsa/src/web/graphql/tags.js
bjoernricks: Got it. It is a conversion between GSA entity types and the enum in the backend.

gsa/src/web/graphql/__tests__/tags.js
```diff
@@ -0,0 +1,137 @@
...
+  newData: createTagResult,
+};
+
+const modifyTagMock = {
+  request: {
+    query: MODIFY_TAG,
+    variables: {input: modifyTagInput},
+  },
+  newData: modifyTagResult,
+};
```
bjoernricks, line 68: Would be best to put this into a mock module. See `web/graphql/__mocks__`

gsa/src/web/pages/tags/component.js
```diff
@@ -175,140 +175,144 @@ class TagComponent extends React.Component {
...
+          })}
+          {dialogVisible && (
+            <TagDialog
+              active={active}
+              comment={comment}
+              id={id}
+              name={name}
+              resource_ids={resource_ids}
+              resource_type={resource_type}
+              resource_types={resource_types}
```
bjoernricks, line 299: Would be nice to change these into camelCase too

gsa/src/web/graphql/__tests__/tags.js
```diff
@@ -0,0 +1,217 @@
...
+  createCreateTagMock,
+  createModifyTagMock,
+  createEnableTagMock,
+  createDisableTagMock,
+  createRemoveTagMock,
+} from '../__mocks__/tags';
+
+import {useCreateTag, useModifyTag, useToggleTag, useRemoveTag} from '../tags';
+
+const tag = Tag.fromElement({_id: '12345'});
```
bjoernricks, line 41: Why not
```suggestion
const tag = Tag.fromObjectt({id: '12345'});
```
?

PR #2248, opened 2020-06-19
Add loading indicator for cves on cpe detailspage

sarahd93:
**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/pages/cpes/detailspage.js
+++ b/gsa/src/web/pages/cpes/detailspage.js
@@ -41,6 +41,8 @@ import Tabs from 'web/components/tab/tabs';
 
 import DetailsLink from 'web/components/link/detailslink';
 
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/cpes/detailspage.js
```diff
@@ -136,8 +138,10 @@ const Details = ({entity, links = true}) => {
               ))}
             </TableBody>
           </Table>
-        ) : (
+        ) : cve_refs === 0 ? (
```
bjoernricks, line 141: I guess cve_refs is a count and not an array right?

sarahd93: Right, cves is the array with all the cves in in it and cve_refs is the count. I totally missed that at first. cve_refs is also used for showing the number of cves on the listpage.

PR #2271, opened 2020-06-26
add automatic reload to task detailspage

sarahd93:
Add automatic reload to task detailspage similar to the reload on the listpage.

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/pages/tasks/detailspage.js
+++ b/gsa/src/web/pages/tasks/detailspage.js
@@ -362,9 +362,8 @@ const Page = ({
     [task, gmpSettings],
   );
 
-  const reloadFunc = useCallback(() => refetch(), [refetch]);
   const [startReload, stopReload, hasRunningTimer] = useReload(
-    reloadFunc,
+    refetch,
     timeoutFunc,
   );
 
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/tasks/detailspage.js
```diff
@@ -344,6 +347,37 @@ const Page = ({
...
+      return gmpSettings.reloadInterval;
+    },
+    [task, gmpSettings],
+  );
+
+  const reloadFunc = useCallback(() => refetch(), [refetch]);
+  const [startReload, stopReload, hasRunningTimer] = useReload(
+    reloadFunc,
+    timeoutFunc,
+  );
```
bjoernricks, line 369: Not sure if I did that wrong at the list page too but there is no need to define a reloadFunc here. We can use refetch directly.
```suggestion
  const [startReload, stopReload, hasRunningTimer] = useReload(
    refetch,
    timeoutFunc,
  );
```

sarahd93: Yeah, I got that from the listpage. I am going to adjust both pages.

PR #2275, opened 2020-06-26
Add mock graphql creator and implement for tags

saberlynx:
**Checklist**:

- [x] Tests


gsa/src/web/graphql/__mocks__/tags.js
```diff
@@ -28,146 +28,89 @@ import {
...
-
-  const queryMock = {
-    request: {
-      query: TOGGLE_TAG,
-      variables: {input: {id: tag.id, active: true}},
-    },
-    newData: enableTagResult,
-  };
-  return [queryMock, enableTagResult];
+export const modifyTagQueryMock = createGenericQueryMock(
```
bjoernricks, line 61: Please use a function to always create new mock data for each test. We need to ensure that all tests can be run independently. It may be possible that a test changes the mocked data and therefore running tests in a undefined order some tests may fail randomly.

PR #2279, opened 2020-06-30
Implement bulk export query and abstract bulk methods into own import

saberlynx:
**Checklist**:

- [x] Tests


saberlynx:
> So we have bulkactions now that contain the BulkTagComponent, useBulkExportEntities and useBulkDeleteEntities, right? There are tests for the BulkTagComponent but I don't see any for useBulkExportEntities and useBulkDeleteEntities. We should test those too if possible.

These are all tested in `web/graphql/__tests__/tasks.js` I think, tested the same way that all other task queries are tested. Bulk delete query tests were implemented in a previous PR. 

> The task listpage contains a test "should allow to export all displayed tasks". Parts of the test are commented out because the bulk actions were still missing. Shouldn't this work now, since bulk export for tasks is implemented? (Also "should allow to export all displayed tasks" is probably not the best title because bulk delete seems to be tested as well.)

I don't think it should because stuff like that used to be from gmp, so we can check from the listpage if they are called. Now it's a graphql function. I should try to see if I can put the mock into the listpage test.

gsa/src/web/entities/bulkactions.js
```diff
@@ -229,4 +234,93 @@ BulkTagComponent.propTypes = {
...
+            fileNameFormat: listExportFileName,
+            resourceType,
+            username,
+          });
+
+          const commandName =
+            'export' + capitalizeFirstLetter(resourceType) + 'ByFilter';
+
+          const xml = response?.data;
+          const exportedEntities = xml[commandName]?.exportedEntities;
```
bjoernricks, line 270: Could that ever happen that response has no data? Wouldn't that be an error that should be shown to the user? I really think the next line will crash in that case.

What about the exportedEntities if they aren't found? will this be handled in onDowload? Can that ever happen?

gsa/src/web/pages/tasks/listpage.js
```diff
@@ -162,6 +158,23 @@ const TasksListPage = () => {
...
+    {counts, tasks, error, loading: isLoading, refetch, called, pageInfo},
+  ] = useLazyGetTasks();
+
+  const exportTasksByFilter = useExportTasksByFilter();
+  const exportTasksByIds = useExportTasksByIds();
+  const bulkExportTasks = useBulkExportEntities();
+
+  const [deleteTask] = useDeleteTask();
+const [deleteTasksByIds] = useDeleteTasksByIds();
+const [deleteTasksByFilter] = useDeleteTasksByFilter();
```
bjoernricks, line 175: Not sure why prettier didn't fix the indentation here.
```suggestion
    const [deleteTasksByIds] = useDeleteTasksByIds();
    const [deleteTasksByFilter] = useDeleteTasksByFilter();
```

gsa/src/web/entities/__tests__/bulkactions.js
```diff
@@ -0,0 +1,439 @@
...
+    const message = getByTestId('message');
+
+    expect(message).toHaveTextContent('Not called');
+
+    const button = screen.getByTestId('load');
+    fireEvent.click(button);
+
+    await wait();
+
+    expect(exportByIdsFunc).toHaveBeenCalled();
```
sarahd93, line 286:
```suggestion
    expect(exportByIdsFunc).toHaveBeenCalled();
    expect(exportByFilterFunc).not.toHaveBeenCalled();
```
I would always check both functions for the sake of completeness so we can be sure that only the correct one is called.

gsa/src/web/entities/__tests__/bulkactions.js
```diff
@@ -0,0 +1,439 @@
...
```
sarahd93, line 307:
```suggestion
    expect(exportByIdsFunc).toHaveBeenCalled();
    expect(exportByFilterFunc).not.toHaveBeenCalled();
```

gsa/src/web/entities/__tests__/bulkactions.js
```diff
@@ -0,0 +1,439 @@
...
+    const message = getByTestId('message');
+
+    expect(message).toHaveTextContent('Not called');
+
+    const button = screen.getByTestId('load');
+    fireEvent.click(button);
+
+    await wait();
+
+    expect(exportByFilterFunc).toHaveBeenCalled();
```
sarahd93, line 328:
```suggestion
    expect(exportByFilterFunc).toHaveBeenCalled();
    expect(exportByIdsFunc).not.toHaveBeenCalled();
```

gsa/src/web/pages/tasks/listpage.js
```diff
@@ -162,6 +158,23 @@ const TasksListPage = () => {
...
+    {counts, tasks, error, loading: isLoading, refetch, called, pageInfo},
+  ] = useLazyGetTasks();
+
+  const exportTasksByFilter = useExportTasksByFilter();
+  const exportTasksByIds = useExportTasksByIds();
+  const bulkExportTasks = useBulkExportEntities();
+
+  const [deleteTask] = useDeleteTask();
+    const [deleteTasksByIds] = useDeleteTasksByIds();
+    const [deleteTasksByFilter] = useDeleteTasksByFilter();
```
bjoernricks, line 175: The indentation still seems to be wrong here. Maybe a bug in prettier...

PR #2290, opened 2020-07-02
Deal with duplicate settings in dashboard getSetting()

swaterkamp:
**Checklist**:

- [N/A] Tests

### DIFF
--- a/gsa/src/gmp/commands/dashboards.js
+++ b/gsa/src/gmp/commands/dashboards.js
@@ -87,8 +87,10 @@ class DashboardCommand extends GmpCommand {
       }
 
       if (isArray(setting)) {
-        setting = setting[0]; // use first setting to avoid crash if the same
-        // setting is returned twice in an array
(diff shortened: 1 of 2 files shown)


gsa/src/gmp/commands/dashboards.js
```diff
@@ -80,18 +80,22 @@ class DashboardCommand extends GmpCommand {
...
-      const {setting} = data.get_settings.get_settings_response;
+      let {setting} = data.get_settings.get_settings_response;
 
       if (!isDefined(setting)) {
         return response.setData({});
       }
 
+      if (isArray(setting)) {
+        setting = setting[0]; // use first setting to avoid crash if the same
+        // setting is returned twice in an array
```
bjoernricks, line 91:
```suggestion
        // before https://github.com/greenbone/gvmd/pull/1106 it was possible that
        // a setting with the same UUID is added twice to the db
        // therefore use first setting to avoid crash
        setting = setting[0];
```

PR #2294, opened 2020-07-03
Adjust parsing in cve model

sarahd93:
Fix parsing in cve model. 

This fixes
- missing reference identifiers
- disappearing vulnerable products section
- missing published and last updated dates

It also removes the cwe id as this has not been displayed starting with GOS 5 and seems not to be needed anymore.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/cve.js
+++ b/gsa/src/gmp/models/cve.js
@@ -18,7 +18,7 @@
  */
 import 'core-js/features/object/entries';
 
-import {isDefined} from 'gmp/utils/identity';
+import {isArray, isDefined} from 'gmp/utils/identity';
 import {isEmpty} from 'gmp/utils/string';
 import {map} from 'gmp/utils/array';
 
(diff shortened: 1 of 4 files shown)


gsa/src/gmp/models/cve.js
```diff
@@ -124,17 +124,13 @@ class Cve extends Info {
...
-        ret.cweId = entry.cwe._id;
-      }
-
-      ret.publishedTime = parseDate(entry['published-datetime'].__text);
-      ret.lastModifiedTime = parseDate(entry['last-modified-datetime'].__text);
+      ret.publishedTime = parseDate(entry['published-datetime']);
+      ret.lastModifiedTime = parseDate(entry['last-modified-datetime']);
 
       ret.references = map(entry.references, ref => ({
         name: ref.reference.__text,
```
swaterkamp, line 131: This is still working with __text and ret.source is not?

sarahd93: The problem is not that __text is not working anymore. The difference is having <foo>text</foo> and <foo id=123>text</foo> or <foo>text <bar /></foo>. For the first one __text does not work anymore but for the other two it does work and is actually needed.

Issue #652, opened 2018-05-13
make base path configurable instead of hardcoded /


Fregf:
I want to use gsad behind an Apache reverse proxy, in a subdirectory of the root of a domain. This mostly works with this Apache configuration:
```
<Location /openvas>
        ProxyPass http://localhost:9392
	ProxyPassReverse http://localhost:9392
        ProxyHTMLEnable On
        ProxyHTMLURLMap /       /openvas/
	ProxyAddHeaders On
</Location>
```
However even then there are some places where ProxyHTMLURLMap fails to rewrite the path of some image resources (maybe in Javascript?). It would be useful if there were a way to define a base path in gsad itself, and that it would automatically make all resources available in that path instead of in the root directory.

bjoernricks:
Is this about master or gsa-7.0?

root360-AndreasUlm:
Hi,

I tried the same with openvas9 on Ubuntu 18.04 and it still is not possible to change the base-url.
It would be a nice feature to be able to configure the base-url.

The expected behaviour would be:
BASE_URL='/' => http://domain/
BASE_URL='/bla' => http://domain/bla/

Can you implement this configuration option?

Regards,
Andreas

bjoernricks:
This won't get fixed for GSA 7 and 8 anymore. I am open for a PR to master.

Ethnical:
Hi, 
Anyone know if this is fixed on the last release ? 
thank

bjoernricks:
Nobody did provide a PR. Therefore it is still not supported.

Virsacer:
This change is also included in my Docker-image:
https://hub.docker.com/r/virsacer/gvm

PR #2307, opened 2020-07-12
Allow relative paths for reverse proxy

Virsacer:
Support relative paths for reverse proxy

Issue https://github.com/greenbone/gsa/issues/652

For example:
`	<Location "/GVM">`
`		ProxyPass https://127.0.0.1:8443`
`		ProxyPassReverse https://127.0.0.1:8443`
`		RedirectMatch ^/GVM$ /GVM/`
`	</Location>`
`	ProxyPass /gmp https://127.0.0.1:8443/gmp`

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/package.json
+++ b/gsa/package.json
@@ -1,6 +1,7 @@
 {
   "name": "gsa",
   "version": "20.08",
+  "homepage": ".",
   "description": "Greenbone Security Assistant",
   "keywords": [
     "openvas",
--- a/gsa/src/web/routes.js
+++ b/gsa/src/web/routes.js
@@ -17,7 +17,7 @@
  */
 import React from 'react';
 
-import {Router, Route, Switch} from 'react-router-dom';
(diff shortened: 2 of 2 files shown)


Virsacer:
When you set a proxy in location /GVM as in the example above, your URL becomes /GVM/login.
But the application does not know how to handle that.
HashRouter makes that URL /GVM/#/login so the leading stuff does not matter and the application knows, what has been requested.

Second problem ist, that images are still loaded from /img which is not proxied.
Same goes for css, js and also redirects... So you get a blank page.
Setting homepage to "." makes the paths relative, and images are loaded from /GVM/img instead and everything works.

I dont know react and what the state of the art technology is - I found this solution with google, trial and error...
I also read about a basepath attibute for Router, but that seems to hardcode the prefix at compile-time.
Do you have another idea?

PR #2314, opened 2020-07-13
Fix various typos.

cfi-gb:
Found with codespell.

The `msgid "# of Vulnerabilties"` and `msgid "Modifed"` already existed in their correct variant. Hope just deleting these is the correct way of fixing this.

*Edit* Removed the .pot file completely as requested in https://github.com/greenbone/gsa/pull/2314#pullrequestreview-447316167


swaterkamp:
Thanks a lot!

PR #2324, opened 2020-07-16
Show stop icon based on task or audit status

saberlynx:
**Checklist**:

- [x] Tests


saberlynx:
> This is going to change the behaviour on both the detailspage and the listpage, right? Which is actually a good thing, as far as I'm concerned, as I don't think they should work differently. Just wanted to make sure I understand this correctly.

I think so. Just because when I looked at the ToolBarIcons it seems that everything is decided in the StopIcon component (e.g. whether to render, and whether it's active). Detailspage doesn't decide that as far as I can tell.

> I am a little worried that we will have stages were we have more icons showing up than usual and then vanishing again (in the transition from requested to queued to running). Not sure if this is actually happening, though. If everything looks normal, you can just ignore this.

Hasn't happened in my setup. Maybe it bears more testing.

saberlynx:
Not implementing in this version.

PR #2333, opened 2020-07-22
Do not use resultsCounts in filter message

saberlynx:
Currently, if we filter for a nonexistent result under the results tab of the report detailspage and there is an empty report, we get a message like "filter does not match any of the N results" where N is the number of results in the database. This is confusing because ideally, this N should refer to the total number of results in the *report* but due to technical constraints, it is not possible to do that. This will be re-implemented in a future date. For now, we omit this number in the message.

**Checklist**:

- [x] Tests


gsa/src/web/pages/reports/details/emptyresultsreport.js
```diff
@@ -43,7 +43,7 @@ const FilterString = styled.span`
 `;
 
 const EmptyResultsReport = ({
-  all,
+  all, // 'all' is not used right now because it refers to the no. of rows in the results table. Should be re-implemented in future versions.
```
sarahd93, line 46: I think since we do not use "all" anymore we should remove it completely. We don't know when we are going to re-implement it and since the actual implementation will probably be in the resultstab the prop we give to the emptyresultsreport might not even be called "all" then. If we don't use it at the moment it should not be in here.

saberlynx:
A conceptually different implementation was chosen in the end.

PR #2339, opened 2020-07-24
Add filtered links to GVMD_DATA row in feed status page

saberlynx:
This PR adds links to reportformats, scanconfigs, policies and portlists on the feedstatus table, filtered for the gvmd data objects. So if the user makes e.g. a new report format, clicking the link will lead to a listpage filtered by ONLY the received feed objects.

**Checklist**:

- [x] Tests


gsa/src/web/pages/extras/feedstatuspage.js
```diff
@@ -162,6 +171,34 @@ const FeedStatus = () => {
                       </Link>
                     </IconDivider>
                   )}
+                  {feed.feed_type === GVMD_DATA && (
+                    <IconDivider>
+                      <Link to="policies">
+                        <IconDivider align={['start', 'center']}>
+                          <PolicyIcon size="medium" />
+                          <span>Compliance Policies</span>
```
swaterkamp, line 218: These titles should be translatable. While you're at it, please also make the old titles translatable if applicable.

gsa/src/web/pages/extras/__tests__/feedstatuspage.js
```diff
@@ -17,17 +17,22 @@
...
 
 import FeedStatus from '../feedstatuspage';
 import {Feed} from 'gmp/commands/feedstatus';
 
 import Response from 'gmp/http/response';
 
+const mockDate = new Date(1595833720475);
+
+// set mockDate so the feed ages don't keep changing
+global.Date.now = jest.fn(() => mockDate);
```
swaterkamp, line 30: This needs to be reset after the test, or the .now method will be "permanently" overwritten.

PR #2350, opened 2020-07-29
Add reload timer to feedstatuspage and displaying when updating is in progress

saberlynx:
**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/commands/__tests__/feedstatus.js
+++ b/gsa/src/gmp/commands/__tests__/feedstatus.js
@@ -57,40 +57,4 @@ describe('FeedStatusCommand tests', () => {
       expect(data[0].version).toEqual('20190625T1319');
     });
   });
-
-  test('should return null currently_syncing attribute and not undefined', () => {
-    const response = createResponse({
-      get_feeds: {
-        get_feeds_response: {
-          feed: {
-            type: 'NVT',
-            name: 'foo',
-            version: 201906251319,
-            description: 'bar',
-          },
-        },
-      },
-    });
-
-    const fakeHttp = createHttp(response);
-
-    expect.hasAssertions();
-
-    const cmd = new FeedStatus(fakeHttp);
-    return cmd.readFeedInformation().then(resp => {
-      expect(fakeHttp.request).toHaveBeenCalledWith('get', {
-        args: {
-          cmd: 'get_feeds',
-        },
-      });
-
-      const {data} = resp;
-      expect(data[0].feed_type).toEqual('NVT');
-      expect(data[0].name).toEqual('foo');
-      expect(data[0].description).toEqual('bar');
-      expect(data[0].currently_syncing).toBeUndefined();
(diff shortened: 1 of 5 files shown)


gsa/src/web/pages/extras/feedstatuspage.js
```diff
@@ -265,4 +263,39 @@ const FeedStatus = () => {
...
+  feeds: PropTypes.array,
+};
+
+const FeedStatusWrapper = () => {
+  const gmp = useGmp();
+  const [feeds, setFeeds] = useState([]);
+
+  const loadFeeds = () =>
+    // cannot store gmp.feedstatus.readFeedInformation in separate variable
+    // causes undefined "this" in FeedStatus class
```
sarahd93, line 276: I don't quiet get why we would want to do this in the first place. The comment seems unnecessary.

gsa/src/web/pages/extras/feedstatuspage.js
```diff
@@ -265,4 +263,39 @@ const FeedStatus = () => {
...
+      setFeeds(data);
+    });
+
+  const calculateSyncInterval = (feedsArray = []) => {
+    const hasFeed = hasValue(feedsArray[0]);
+
+    return hasFeed && hasValue(feedsArray[0].currently_syncing)
+      ? USE_DEFAULT_RELOAD_INTERVAL_ACTIVE
+      : USE_DEFAULT_RELOAD_INTERVAL;
+  }; // currently, if one feed is syncing then all feeds have the currently_syncing attribute if that changes, this will have to be redone
```
sarahd93, line 288: Maybe we can just use something like `feedsArray.some(feed => feed.currently_syncing)` to check if at least one feed is updating. Then it should still work even if the behaviour changes.

gsa/src/web/pages/extras/feedstatuspage.js
saberlynx: That's because I moved the loadFeeds and setting feeds out of the FeedStatus and into the wrapper. I tried to keep it in there via a variable pointing to gmp.feedstatus.readFeedInformation() but it caused the aforementioned problem. Thought I might explain myself, although if it's clear I can get rid of it.

gsa/src/web/pages/extras/feedstatuspage.js
saberlynx: We can also do that. I was trying to avoid unnecessary looping but i guess with only 4 feeds it shouldn't be a problem. it's not like we're going to add an unlimited number of feeds later.

gsa/src/gmp/commands/__tests__/feedstatus.js
```diff
@@ -30,6 +30,7 @@ describe('FeedStatusCommand tests', () => {
             name: 'foo',
             version: 201906251319,
             description: 'bar',
+            currently_syncing: {timestamp: 'baz'},
```
bjoernricks, line 33: Should have been a camel case variable.

PR #2365, opened 2020-08-06
Fix pagination for report results tab

bjoernricks:
Use the loaded results counts not the result counts from the report. The
results are loaded independently of the report now and therefore the
result counts from the report doesn't fit anymore.

**Checklist**:

- [ ] Tests


gsa/src/web/pages/reports/details/resultstab.js
```diff
@@ -364,6 +364,7 @@ const mapStateToProps = (state, {reportId}) => {
     resultsFilter,
     resultsError: selector.getEntitiesError(resultsFilter),
     results: selector.getEntities(resultsFilter),
+    resultsCounts: selector.getEntitiesCounts(resultsFilter),
```
sarahd93, line 367: Doesn't this revert the changes from https://github.com/greenbone/gsa/pull/2335/files ? That was probably what broke the pagination in the first place. But if we can't do it like that we probably need a different fix for the number in the EmptyResultsReport.

saberlynx: What about this potentially awful idea. Pass the upper resultCount down under a different name and use it for EmptyResultsReport.

bjoernricks: Yes that was my idea too. Done with be2133b8568dbcca39ce8219671ce8b6a9d64746

bjoernricks: Actually in e01e9b424e9b66f4cf10dd50c9c8c064ae55a83f after the rebase.

PR #2366, opened 2020-08-06
Fix report result filter determination

bjoernricks:
Don't use the current stored filter of the results list page for loading results at the report details page.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/reports/detailspage.js
+++ b/gsa/src/web/pages/reports/detailspage.js
@@ -715,6 +715,11 @@ const load = ({
   }
 
   if (!hasValue(filter)) {
+    // use filter from user setting
+    filter = defaultFilter;
+  }
+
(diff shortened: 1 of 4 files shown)


gsa/src/web/pages/reports/detailspage.js
```diff
@@ -714,6 +715,16 @@ const load = ({
     filter = reportFilter;
   }
 
+  if (!hasValue(filter)) {
+    // use filter from store
+    filter = pageFilter;
+  }
```
bjoernricks, line 721: I am not 100 percent sure if these lines are really necessary because the page filter will be included in the defaultFilter by the FilterProvider already. Currently this overrides a possible `?filter=...` url parameter (named locationQueryFilter in FilterProvider).

PR #2388, opened 2020-08-14
fix missing nvt solution type

sarahd93:
NVT solution was not parsed correctly and therefore it was missing on listpage and detailspage.

**Checklist**:

- [ ] Tests


sarahd93:
> I am just wondering aren't there any tests for this?

There are tests but they didn't catch it. 
For the nvts we have an info element that contains an nvt element and the first thing we do in the model is dumping the contents of both together so we have everything in the same element. In the tests, however, we do not make this distinction, the input we use there already has all the content in the same element.
I've already been investigating this and we might want to think about if we want to change this. However, that means we would not only need to adjust the nvt tests but the tests for all other info elements (CVEs, CPEs, etc) as well.

PR #2405, opened 2020-08-20
Add new getOverrides query

swaterkamp:
**What**:
This PR adds getOverrides via graphql to GSA

**Why**:
To use the new backend properly

**How**:
Number of available overrides is shown in the task toolbar icons properly. Test suites pass.

**Checklist**:

- [X] Tests

### DIFF
--- a/gsa/src/web/pages/tasks/detailspage.js
+++ b/gsa/src/web/pages/tasks/detailspage.js
@@ -82,16 +82,14 @@ import CloneIcon from 'web/entity/icon/cloneicon';
 import EditIcon from 'web/entity/icon/editicon';
 import TrashIcon from 'web/entity/icon/trashicon';
 
+import {useLazyGetOverrides} from 'web/graphql/overrides';
+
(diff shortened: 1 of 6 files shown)


sarahd93 (dismissed):

sarahd93 (approved):

gsa/src/web/pages/tasks/detailspage.js
```diff
@@ -383,6 +381,14 @@ const Page = ({
     }
   }, [history]);
 
+  const [loadOverrides, {overrides}] = useLazyGetOverrides({
+    filterString: 'task_id:' + id,
+  });
+
+  useEffect(() => {
+    loadOverrides();
+  }, [task]);
```
bjoernricks, line 390: Shouldn't it be `id` instead of `task`? Don't forget the task object changes with every reload. That means with every reload the overrides would be loaded again too (not considering apollo caching atm)
```suggestion
  }, [id]);
```

bjoernricks (changes requested):

sarahd93 (approved):

PR #2406, opened 2020-08-20
Refactor alert component and alert dialog into function component

saberlynx:
**Checklist**:

- [ ] Tests


bjoernricks (approved):
The whole alert dialog and it's state handling is still way to difficult and hard to understand. I really think we need to come up with a new concept for the alert dialog in future.

PR #2425, opened 2020-08-27
Create alerts via graphql

saberlynx:
**What**:
Implement a `handleSaveAlert` method using the `createAlert` mutation then refetch alerts after creating the alert.

**Why**:

It is possible to create alerts from the task dialog.

**How**:

**Checklist**:

- [x] Tests


gsa/src/web/graphql/__mocks__/alerts.js
```diff
@@ -120,3 +120,30 @@ const mockAlerts = {
...
+
+export const createAlertInput = {
+  name: 'foo',
+  event: 'NEW_SECINFO_ARRIVED',
+  condition: 'ALWAYS',
+  method: 'HTTP_GET',
+  methodData: {
+    URL: 'yourdomain.com',
+    composer_include_notes: 1,
+    composer_include_overrides: 0,
```
bjoernricks, line 139: Should not be send via the api. At least not as methodData. We need to think about the api changes for the content composer sometime in future.

bjoernricks (changes requested):
The whole conversion seems to be unnecessary to me. Why not unify the API and frontend? The frontend should just use the values that the backend expects. And you are free to adjust the backend for the frontend needs.

bjoernricks (approved):
Am really not satisfied with the alert handling and the API but we leave it as it is atm.

PR #2448, opened 2020-09-09
Drop dynamic severity classes

janowagner:
**What**:

The support for multiple severity class ranges is removed.
There is now only one classification scheme (then one previously
called "nist"). Actually there was only one other classification scheme called
"pci-dss". It wasn't used or needed.

For the user, this essentially means that in "My Settings", the option
to select a severity classification scheme is not there anymore and that
the default is now applied mandatory.

This PR requires: https://github.com/greenbone/gvmd/pull/1288

**Why**:

Removing the overhead about a dynamic handling where actually only static
class ranges are used simplifies code.
  

**How**:

Basically all aspects are removed, on gsa side, and on gvmd side.
While gvmd appears pretty clean now, in gsa there are some special
handlings with an old range style due to compatibility with older dashboard elements.

**Checklist**:

- [ ] Tests N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 ### Removed
+- Drop dynamic severity classes [#2448](https://github.com/greenbone/gsa/pull/2448)
 - Removed severity class 'bsi' [#2434](https://github.com/greenbone/gsa/pull/2434)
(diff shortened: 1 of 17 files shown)


bjoernricks:
I am a bit surprised that no tests had to be adjusted. @janowagner did you check all charts if they still work?

janowagner:
> I am a bit surprised that no tests had to be adjusted. @janowagner did you check all charts if they still work?

Indeed this code was no covered by tests. I checked all charts manually and all looked fine.

bjoernricks (approved):

PR #2453, opened 2020-09-14
Load credentials with graphql

sarahd93:
**What**:

Add get_credentials query and corresponding tests.
Use query to load credentials in task component.
Use query to load credentials in target component (because the task dialog can open the target dialog).

Adjusting the alert component will follow in another PR.

**Why**:

To adjust the task page to load everything via graphql.

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/credential.js
+++ b/gsa/src/gmp/models/credential.js
@@ -17,12 +17,12 @@
  */
 import Model, {parseModelFromElement} from '../model';
 
-import {_l} from '../locale/lang';
+import {_, _l} from '../locale/lang';
 
 import {isDefined} from '../utils/identity';
 import {map} from '../utils/array';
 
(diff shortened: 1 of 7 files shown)


saberlynx (approved):

gsa/src/gmp/models/credential.js
```diff
@@ -108,6 +127,18 @@ export const getCredentialTypeName = type => `${TYPE_NAMES[type]}`;
 class Credential extends Model {
   static entityType = 'credential';
 
+  static parseObject(object) {
+    const copy = super.parseObject(object);
+
+    copy.credential_type = parseCredentialType(object.type);
+
+    copy.allow_insecure = parseBoolean(object.allow_insecure);
```
bjoernricks, line 135: Should be camelCase in future

PR #2463, opened 2020-09-22
Edit dialog: only send name and comment if config or policy is in use, and add in use notification

saberlynx:
**What**:

If a config or policy is in use, ONLY send name and comment fields (and id, of course) through the ScanConfigCommand or PolicyCommand. Added a h4 notification in the edit dialog to make users aware that they can only edit name and comment for in use config/policy.

**Why**:

Currently, the edit scan config dialog sends all the form fields (even if unchanged) to the backend, even if only name and comment fields are rendered (in the case of in-use config/policy). This triggers a response from gvmd that "since the entity is in use, only name and comment can be edited". Due to this error, even if the user does everything correctly and only edit the fields available to them, it will trigger a gsad error and the dialog will not close. This is wrong and confusing for the user.

This change avoids triggering the error because nothing outside of the permitted fields are "edited", and the user is made aware why the edit dialog is so short in this case.

**How**:

**Checklist**:

- [x] Tests


gsa/src/web/pages/policies/component.js
```diff
@@ -766,7 +766,16 @@ class PolicyComponent extends React.Component {
                   onEditNvtDetailsClick={this.openEditNvtDetailsDialog}
                   onSave={d => {
                     this.handleInteraction();
-                    return save(d).then(() => this.closeEditPolicyDialog());
+                    // eslint-disable-next-line no-shadow
+                    const {name, comment, id} = d;
```
sarahd93, line 770: Since we need multiple lines for this we probably should not put it all in here but add a handleSavePolicy function instead.

gsa/src/web/pages/scanconfigs/component.js
```diff
@@ -514,7 +514,15 @@ class ScanConfigComponent extends React.Component {
                   onEditNvtDetailsClick={this.openEditNvtDetailsDialog}
                   onSave={d => {
                     this.handleInteraction();
-                    return save(d).then(() => this.closeEditConfigDialog());
+                    const {name, comment, id} = d;
```
sarahd93, line 517: Same as above. I think we should use a handleSaveConfig function.

gsa/src/web/pages/scanconfigs/editdialog.js
```diff
@@ -185,8 +185,13 @@ const EditScanConfigDialog = ({
               onChange={onValueChange}
             />
           </FormGroup>
-
-          {!configIsInUse && (
+          {configIsInUse ? (
+            <h4>
+              {_(
+                'The config or policy is now in use by a task or audit, therefore only name and comment can be modified.',
```
sarahd93, line 191: Most users do not really know that configs and policies are technically the same and ideally they should not know it. So this message might be a little confusing. We should probably show different messages for configs and policies.

sarahd93 (changes requested):

gsa/src/web/pages/scanconfigs/editdialog.js
swaterkamp: We will also need a translation for the new notification.
```suggestion
                'The config or policy is currently in use by a task or audit, therefore only name and comment can be modified.',
```

sarahd93 (approved):

swaterkamp (approved):

PR #2478, opened 2020-09-28
Validate rows per page setting and fix tooltips

swaterkamp:
**What**:
The "Rows Per Page" setting is restricted to use a value greater than 0, because 0 makes no sense in the first place and is showing at least 1 rows in the end. The new validation rule also prohibits an empty field at the same time.

Helping tooltips are now lazy translated because otherwise with the normal translation the error message was undefined.

**Why**:
Improve UX with helping tooltips and fix a bug

**How**:
Tooltips are now shown in the createProcessDialog and the rows per page setting in der userSettings dialog. The setting is invalid if the input is 0 or empty.

**Checklist**:

- [ ] Tests


gsa/src/web/pages/usersettings/dialog.js
```diff
@@ -175,14 +180,27 @@ let UserSettingsDialog = ({
...
+    validityStatus,
+    handleSubmit,
+  } = useFormValidation(validationSchema, userSettingsRules, onSave);
+
   return (
     <SaveDialog
       title={_('Edit User Settings')}
       onClose={onClose}
-      onSave={onSave}
+      onSave={values => handleSubmit(values)}
```
bjoernricks, line 199:
```suggestion
      onSave={handleSubmit}
```

bjoernricks (changes requested):
Just a small change. No need to create a new function on every render.

bjoernricks (approved):

PR #2484, opened 2020-10-01
Implement create and modify port list by graphql

saberlynx:
**What**:

- [x] Refactor portlist component into function
- [x] Use createportlist mutation
- [x] Test createportlist mutation
- [x] Implement into target dialog
- [x] use modifyPortList mutation
- [x] Test modifyPortList mutation
- [x] Implement handleSavePortList method
- [x] Implement redirect for creating portlist from detailspage

Also implemented a file reader component in case we need more of that in the future.

**Why**:

Because it is possible to create a port list from the target dialog, which is accessible from the task dialog.

**How**:

Test + manual test

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/portlists/component.js
+++ b/gsa/src/web/pages/portlists/component.js
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
+import React, {useState, useReducer} from 'react';
 
 import _ from 'gmp/locale';
 
@@ -25,7 +25,8 @@ import {isDefined} from 'gmp/utils/identity';
 import {shorten} from 'gmp/utils/string';
 
 import PropTypes from 'web/utils/proptypes';
-import withGmp from 'web/utils/withGmp';
+import useGmp from 'web/utils/useGmp';
+import reducer from 'web/utils/stateReducer';
 
 import EntityComponent from 'web/entity/component';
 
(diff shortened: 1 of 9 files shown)


gsa/src/web/pages/portlists/component.js
```diff
@@ -25,186 +25,205 @@ import {isDefined} from 'gmp/utils/identity';
...
+}) => {
+  const gmp = useGmp();
+
+  const [state, dispatch] = useReducer(reducer, {
+    importDialogVisible: false,
+    portListDialogVisible: false,
+    portRangeDialogVisible: false,
+  });
+
+  const [createdPortRanges, setCreatedPortRanges] = useState([]);
```
saberlynx, line 61: Is special treatment here unnecessary? Can they be used like normal states?

gsa/src/web/pages/portlists/component.js
```diff
@@ -25,186 +25,205 @@ import {isDefined} from 'gmp/utils/identity';
...
+  const [deletedPortRanges, setDeletedPortRanges] = useState([]);
```
saberlynx, line 62: Same here.

gsa/src/web/pages/portlists/component.js
bjoernricks: Just some background for using `this.foo` for a *global* variable in the past. Sometimes it wasn't necessary to re-render if some variable has been set. E.g. it wasn't displayed after some loading from the backend. Therefore using a global variable instead of updating the state avoid the re-render. I am not sure if this is still the case therefore don't think about re-rendering anymore (at the moment) and don't look back.

sarahd93 (approved):

gsa/src/web/utils/readFileToText.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (isDefined(file)) {
+    fileText = file.text().then(text => {
+      return text;
+    });
+  } else {
+    fileText = Promise.resolve();
+  }
+  return fileText;
+};
```
bjoernricks, line 32: can be written much shorter
```
const readFileToText = file => isDefined(file) ? file.text() : Promise.resolve();
```

gsa/src/web/pages/targets/component.js
```diff
@@ -40,10 +40,12 @@ import TargetDialog from './dialog.js';
 
 import {useLazyGetCredentials} from 'web/graphql/credentials';
 
-import {useLazyGetPortLists} from 'web/graphql/portlists';
+import {useLazyGetPortLists, useCreatePortList} from 'web/graphql/portlists';
 
 import {useCreateTarget, useModifyTarget} from 'web/graphql/targets';
 
+import readFileToText from 'web/utils/readFileToText.js';
```
bjoernricks, line 47: This import should be changed to not include the `.js` suffix.

PR #2486, opened 2020-10-01
Create credential with graphql

sarahd93:
**What**:

Add mutation for createCredential and corresponding test.

Create credentials with graphql in target component.

Create credentials with graphql in alert component.

**Why**:

It is possible to open the target dialog and the alert dialog from the task dialog. Both dialogs offer the option to create a credential. Therefore, the target component and the alert component need to use graphql to create the credentials.

**How**:

Tested in gsa for all credential types in both dialogs.

**Checklist**:

- [x] Tests


bjoernricks (changes requested):
Looks good! just a small thing

bjoernricks (approved):

PR #2494, opened 2020-10-05
Refactor task component and stateReducer

saberlynx:
**What**:

This PR refactors the TaskComponent to unify the state object into one with `useReducer`. It also refactors `stateReducer` to reduce boilerplate, as this function is mostly for replacing `this.setState` for setting large states and will never have any other usecase, I thought all of the boilerplate e.g. `dispatch({type: 'setState', newState: {...}})` is unnecessary. Now we have an action creator `updateState` and pure state updates (with no other logic) can be called with `dispatch(updateState({stateA: 'foo', stateB: 'bar'}))` etc. Adds a test suite for `stateReducer`.

Also, adjusted other components using `stateReducer` to the new syntax. Composed certain (not all. See editnvtdetailsdialog for example of reducer that doesn't need stateReducer logic) custom reducers with stateReducer.

**Why**:

`TaskComponent` was implemented in the early days of the hyperion branch and thus contains some outdated syntax and messy code (like using multiple `useState`s for visibility flags, a separate `dialogState`, etc.). This PR brings the code in line with later refactors for hyperion.

**How**:

Manual and jest tests.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/utils/__tests__/stateReducer.js
+++ b/gsa/src/web/utils/__tests__/stateReducer.js
@@ -1,0 +1,79 @@
+/* Copyright (C) 2020 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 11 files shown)


gsa/src/web/utils/__tests__/stateReducer.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
```
bjoernricks, line 18: oh i just noticed that the license header is wrong. it must be an AGPLv3+ header. Could you check that for the changed files?

bjoernricks (changes requested):
Besides the license header it looks very good to me :+1:

gsa/src/web/utils/__tests__/stateReducer.js
saberlynx: Oops. I'll check all of them.

bjoernricks (approved):

PR #2498, opened 2020-10-09
Drop gmp scanner type from GSA

saberlynx:
**What**:

Drops GMP_SCANNER_TYPE everywhere in GSA.

**Why**:

**How**:

Tests are adjusted accordingly and should pass.

**Checklist**:

- [x] Tests


gsa/src/web/pages/performance/performancepage.js
```diff
@@ -393,10 +386,10 @@ const mapStateToProps = (rootState, {gmp}) => {
 
   if (gmp.settings.enableGreenboneSensor) {
     SLAVE_SCANNER_FILTER = Filter.fromString(
-      'type=' + GMP_SCANNER_TYPE + ' type=' + GREENBONE_SENSOR_SCANNER_TYPE,
+      'type=' + GREENBONE_SENSOR_SCANNER_TYPE,
     );
   } else {
-    SLAVE_SCANNER_FILTER = Filter.fromString('type=' + GMP_SCANNER_TYPE);
+    SLAVE_SCANNER_FILTER = Filter.fromString('');
```
sarahd93, line 392: I would expect this filter to return all scanners and I think we want to return none in this case, right? I might be wrong, though.

bjoernricks: The whole else clause should be removed.

gsa/src/web/pages/performance/performancepage.js
```diff
@@ -376,10 +369,10 @@ const mapDispatchToProps = (dispatch, {gmp}) => {
...
   }
```
bjoernricks, line 376: The else clause should be removed here too

gsa/src/web/pages/audits/row.js
```diff
@@ -109,14 +109,15 @@ const Row = ({
             {entity.alterable === 1 && (
               <AlterableIcon size="small" title={_('Audit is alterable')} />
             )}
-            {isDefined(scanner) && scanner.scannerType === GMP_SCANNER_TYPE && (
```
bjoernricks, line 112: Oh I think this is wrong in 20.08 already and should include the GREENBONE_SENSOR_TYPE in that version too.

saberlynx: Yeah, we forgot to update the audit row when we added the greenbone sensor type

gsa/src/web/pages/performance/performancepage.js
```diff
@@ -376,10 +369,8 @@ const mapDispatchToProps = (dispatch, {gmp}) => {
 
   if (gmp.settings.enableGreenboneSensor) {
```
bjoernricks, line 370: I've taken a look at the code again to be sure we are doing the correct thing with the filter here. IMHO you can just remove the if here and below to always use `Filter.fromString('type=' + GREENBONE_SENSOR_SCANNER_TYPE)`
Maybe also rename the `SLAVE_SCANNER_FILTER` to `SENSOR_SCANNER_FILTER`

gsa/src/web/pages/performance/performancepage.js
```diff
@@ -393,10 +384,8 @@ const mapStateToProps = (rootState, {gmp}) => {
...
```
bjoernricks, line 385: This if should be unnecessary as mentioned above too.

bjoernricks (approved):

PR #2508, opened 2020-10-13
Update versions in CMake files

swaterkamp:
**What**:

**Why**:
To get the correct version into the build

**How**:

**Checklist**:

- [N/A] Tests
- [N/A] [CHANGELOG](https://github.com/greenbone/gsa/blob/master/CHANGELOG.md) Entry


bjoernricks:
Did you test if the version is displayed correctly? `gsad --version` and at the about page?

bjoernricks (approved):

PR #2509, opened 2020-10-13
Implement task wizard via graphql

saberlynx:
**What**:

Implement runQuickFirstScan (Task Wizard) with purely graphql commands.

**Why**:

Because task wizard can be a frontend only feature.

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/graphql/tasks.js
+++ b/gsa/src/web/graphql/tasks.js
@@ -25,6 +25,8 @@ import Task from 'gmp/models/task';
 
 import {isDefined} from 'gmp/utils/identity';
 
+import {CREATE_TARGET} from 'web/graphql/targets';
+
 export const CLONE_TASK = gql`
   mutation cloneTask($id: UUID!) {
     cloneTask(id: $id) {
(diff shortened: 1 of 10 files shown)


gsa/src/web/graphql/tasks.js
```diff
@@ -538,3 +544,51 @@ export const useExportTasksByIds = options => {
 
   return exportTasksByIds;
 };
+
```
sarahd93, line 547: I kinda like this, it's pretty straight forward. Not sure if there is a better way to handle all those nested promises (and there are going to be even more for the advanced wizard, I guess). We might need some error handling in case one of the mutations doesn't work. Not sure how to do this, though.   

And I wonder if we should put this in a separate wizard file. It is more than just task mutations after all and I think the advanced wizard will need even more different entities.

saberlynx: > And I wonder if we should put this in a separate wizard file. It is more than just task mutations after all and I think the advanced wizard will need even more different entities.

That's a good idea. I was noting how long the task file was getting.

gsa/src/web/graphql/wizards.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const runQuickFirstScan = useCallback(
+    data => {
+      const date = dateTimeWithTimeZone(sessionTimeout, userTimezone);
+
+      const {hosts} = data;
+      const targetInputObject = {
+        name: `Target for immediate scan of IP ${hosts} - ${date}`,
+        hosts,
+        portListId: '33d0cd82-57c6-11e1-8ed1-406186ea4fc5', // All IANA assigned TCP
```
bjoernricks, line 47: '33d0cd82-57c6-11e1-8ed1-406186ea4fc5' should be put into a constant variable. (Maybe in the portlist model file?)

gsa/src/web/graphql/wizards.js
```diff
@@ -0,0 +1,77 @@
...
+      };
+
+      return createTarget({
+        variables: {input: targetInputObject},
+      }).then(resp => {
+        const targetId = resp?.data?.createTarget?.id;
+
+        const taskInputObject = {
+          name: `Immediate scan of IP ${hosts}`,
+          configId: 'daba56c8-73ec-11df-a475-002264764cea', // Full and Fast
```
bjoernricks, line 57: I am really sure we already have a constant variable for the full and fast scan config which should be used here

gsa/src/web/graphql/wizards.js
```diff
@@ -0,0 +1,77 @@
...
+          targetId,
+          scannerId: '08b69003-5fc2-4037-a479-93b440211c73', // OpenVAS Default
```
bjoernricks, line 59: same here. we should have a constant for the openvas default scanner.

gsa/src/web/graphql/wizards.js
```diff
@@ -0,0 +1,77 @@
...
+import useUserSessionTimeout from 'web/utils/useUserSessionTimeout';
+import useUserTimezone from 'web/utils/useUserTimezone';
+
+export const useRunQuickFirstScan = () => {
+  const [sessionTimeout] = useUserSessionTimeout();
+  const [userTimezone] = useUserTimezone();
+
+  const [createTarget] = useMutation(CREATE_TARGET);
+  const [createTask] = useMutation(CREATE_TASK);
+  const [startTask] = useMutation(START_TASK);
```
bjoernricks, line 37: didn't we implement custom hooks for these mutations already?

saberlynx: We do. And now I have zero clue why I didn't use them already.

gsa/src/web/graphql/tasks.js
saberlynx: The errors are handled already (will appear in red banner in task wizard dialog), but I wrote some tests to confirm that the entire chain will not go through if there are errors in prior steps of the chain. I didn't test every link in the chain (as they are quite similar) but it would not be difficult to add those test

bjoernricks (approved):
NIce :smile: !

PR #2513, opened 2020-10-16
Replace deprecated sys_siglist with strsignal

tgurr:
Required to work with glibc >= 2.32.

https://sourceware.org/pipermail/libc-announce/2020/000029.html

  The deprecated arrays sys_siglist, _sys_siglist, and sys_sigabbrev
  are no longer available to newly linked binaries, and their declarations
  have been removed from <string.h>. They are exported solely as
  compatibility symbols to support old binaries. All programs should use
  strsignal instead.

**What**: Same as https://github.com/greenbone/gvmd/pull/1280

**Why**: Same as https://github.com/greenbone/gvmd/pull/1280

**How**: Same as https://github.com/greenbone/gvmd/pull/1280

**Checklist**:

- [ ] Tests


bjoernricks (approved):

bjoernricks:
Thanks!

PR #2514, opened 2020-10-16
Fix task wizard and implement modify task wizard via graphql

saberlynx:
**What**:

- Implement runModifyTask (Modify Task Wizard) with purely graphql commands.
- Change the date name to current date (in UTC, in ISO string format with seconds) instead of user session timeout.
- Adjust modify task wizard to have similar logic as schedule dialog e.g. only use `startDate` (`start_minute` and `start_hour` will now be inferred from `startDate`, and the time will dynamically adjust based on `startTimezone`)

TODO
- [x] Dynamically generate input mocks
- [x] Change modify task wizard variable name case
- [x] Clean up code

**Why**:

Because modify task wizard can be a frontend only feature.

**How**:

**Checklist**:

- [x] Tests


gsa/src/web/graphql/wizards.js
```diff
@@ -66,8 +68,108 @@ export const useRunQuickFirstScan = () => {
...
+      reschedule,
+      tasks,
+    }) => {
+      start_date.hours(start_hour);
+      start_date.minutes(start_minute);
+
+      const [task] = tasks;
+      const {name, id: taskId, alerts} = task;
+
+      const rightNow = new Date().toISOString();
```
bjoernricks, line 98: I really thing rightNow could be an optional parameter for runModifyTask too.

gsa/src/web/graphql/wizards.js
```diff
@@ -66,8 +68,108 @@ export const useRunQuickFirstScan = () => {
...
+
+      const event = Event.fromData(
+        {startDate: start_date},
+        start_timezone,
```
bjoernricks, line 102:
```suggestion
        {startDate},
        startTimezone,
```

gsa/src/web/graphql/wizards.js
```diff
@@ -66,8 +68,108 @@ export const useRunQuickFirstScan = () => {
...
+        false, // if true, then the event will be scheduled at the given time, but at the user timezone instead. If I create an event at 22:00 Shanghai time, it will always be scheduled at 22:00 CEST (User timezone)
```
bjoernricks, line 103: As mentioned already I am really sure this is wrong. the passed startDate needs to be already in the correct timezone.

gsa/src/web/graphql/wizards.js
```diff
@@ -66,8 +68,108 @@ export const useRunQuickFirstScan = () => {
...
+      );
+      let schedulePromise;
+
+      if (reschedule) {
+        const createScheduleInput = {
+          name: `Schedule for ${name} - ${rightNow}`,
+          icalendar: event.toIcalString(),
+          timezone: start_timezone,
```
bjoernricks, line 111:
```suggestion
          timezone: startTimezone,
```

gsa/src/web/graphql/wizards.js
```diff
@@ -66,8 +68,108 @@ export const useRunQuickFirstScan = () => {
...
+        schedulePromise = Promise.resolve();
+      }
+
+      return schedulePromise.then(resp => {
+        // Should be undefined if no schedule is created
+        const scheduleId = resp?.data?.createSchedule?.id;
+
+        let alertPromise;
+
+        if (alert_email.length > 0) {
```
bjoernricks, line 125: Personally I would check if alert email is also defined to not crash if it isn't.
```suggestion
        if (isDefined(alertEmail) && alertEmail.length > 0) {
```

gsa/src/web/graphql/wizards.js
```diff
@@ -66,8 +68,108 @@ export const useRunQuickFirstScan = () => {
...
+            name: `Email Alert for ${name} - ${rightNow}`,
+            comment: 'Automatically generated by wizard',
+            event: 'TASK_RUN_STATUS_CHANGED',
+            eventData: {
+              status: 'DONE',
+            },
+            method: 'EMAIL',
+            methodData: {
+              to_address: alert_email,
+              from_address: alert_email,
```
bjoernricks, line 136:
```suggestion
              to_address: alertEmail,
              from_address: alertEmail,
```

bjoernricks (changes requested):

gsa/src/web/graphql/wizards.js
saberlynx: I'll refactor modify task wizard to output variable names in camelCase.

gsa/src/web/graphql/__tests__/wizards.js
```diff
@@ -19,38 +19,57 @@
...
+
+beforeAll(() => {
+  global.Date = jest.fn(() => mockDate);
+  global.Date.now = jest.fn(() => mockDate.getTime());
+});
+
+afterAll(() => {
+  global.Date = RealDate;
+  global.Date.now = RealDate.now;
+});
```
bjoernricks, line 59: Is this still necessary?

gsa/src/web/graphql/__tests__/wizards.js
```diff
@@ -161,3 +154,199 @@ describe('useRunQuickFirstScan tests', () => {
...
+      tasks: [
+        {
+          name: 'myFirstTask',
+          id: '13579',
+          alerts: [{id: '34567'}],
+        },
+      ],
+    })
+      .then(() => setOk(true))
+      .catch(err => setError(err.message));
```
bjoernricks, line 178: you don't need to pass err.message here. The Error class implements to toString() method that returns the message. Therefore the following code would also fit
```suggestion
      .catch(setError);
```

bjoernricks (changes requested):
I would still change `runModifyTask` to allow passing a date. That will be much easier to test then and will not require the mock anymore.

saberlynx:
> I would still change `runModifyTask` to allow passing a date. That will be much easier to test then and will not require the mock anymore.

Ah. So do you mean, allow passing a date so we can pass one when we test, but not necessarily use it in production? (because that would be `rightNow` or something

bjoernricks (approved):
Really really nice now! Thanks a lot for your patience :-)

PR #2520, opened 2020-10-19
Update form fields

bjoernricks:
**What**:

Update form fields to fix a typo, a bug and to remove obsolete behavior.

**Why**:

Cleanup and fixing of form fields.

**How**:

All tests still run, linting is happy and verified that some form fields are still working manually.

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/components/form/textfield.js
+++ b/gsa/src/web/components/form/textfield.js
@@ -29,7 +29,7 @@ const TextField = ({hasError = false, errorContent, title, ...props}) => {
       <Field
         {...props}
         hasError={hasError}
-        title={hasError ? errorContent : title}
+        title={hasError ? `${errorContent}` : title}
         type="text"
       />
       <Marker isVisible={hasError}>×</Marker>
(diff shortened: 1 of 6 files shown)


gsa/src/web/components/form/textfield.js
```diff
@@ -29,7 +29,7 @@ const TextField = ({hasError = false, errorContent, title, ...props}) => {
       <Field
         {...props}
         hasError={hasError}
-        title={hasError ? errorContent : title}
+        title={hasError ? `${errorContent}` : title}
```
saberlynx, line 32: I understand what this does, but I don't quite understand why to do it this way.

bjoernricks: for converting a js object to string you have three choices

* Use string concatenation e.g. `'' + foo` which uses the `toString()` method internally
* Use `toString()` method of the object directly e.g. `foo.toString()`
* Use template literal syntax `` `${foo}` `` which also uses `toString()` internally

Personally I don't think `toString` should be called directly. I see that method more like the `__str__` method in Python which never gets called directly. Additionally I think the template literal syntax is easier to read and more appealing.

saberlynx (approved):

PR #2522, opened 2020-10-22
Update login page in an effort to combine gsa and gmsp themes

janwerremeyer:
**What**:

**Why**:

This is the first step and a proof of concept in making the gsa and gmsp ui more similar

**How**:

- [x ] Tests


gsa/package.json
```diff
@@ -65,7 +67,8 @@
     "redux-thunk": "^2.3.0",
     "styled-components": "^4.4.1",
     "uuid": "^8.3.1",
-    "whatwg-fetch": "^3.4.1"
+    "whatwg-fetch": "^3.4.1",
+    "core-js": "3.6.5"
```
bjoernricks, line 71: Do you really need core-js?

gsa/package.json
```diff
@@ -23,6 +23,8 @@
     "extends": "react-app"
   },
   "dependencies": {
+    "@greenbone/components": "git+https://github.com/chaoa/ui-components.git",
+    "@material-ui/core": "4.11.0",
```
bjoernricks, line 27: Do we really need this import here? Should it be a sub-dependency of @greenbone/components

gsa/package.json
```diff
@@ -23,6 +23,8 @@
     "extends": "react-app"
   },
   "dependencies": {
+    "@greenbone/components": "git+https://github.com/chaoa/ui-components.git",
```
bjoernricks, line 26: I guess the url should be git+https://github.com/greenbone/ui-components/

bjoernricks (changes requested):
Could you remove gsa/package-lock.json and just use yarn v1 for now? maybe we need to discuss using npm and package-lock.json instead of yarn v1 and yarn.lock for the future.

PR #2536, opened 2020-10-28
Added CVSS v3 Calculator.

y0urself:
**What**:

Adding the CVSS v3(.1) Calculator to the GSA Help Page.
It works like the CVSS v2 Calculator (for now), except the score is calculated on the userside and not by gmp.

**Why**:

CVSS v2 will be deprecated and is not up to date anymore.

**How**:

**Checklist**:

- [x] Tests


gsa/src/gmp/models/cve.js
```diff
@@ -21,13 +21,9 @@ import {isArray, isDefined} from 'gmp/utils/identity';
...
 
-import {
-  parseSeverity,
-  parseCvssBaseVector,
-  parseDate,
-  setProperties,
-} from 'gmp/parser';
+import {parseSeverity, parseDate, setProperties} from 'gmp/parser';
 
+import {parseCvssV2BaseVector} from 'gmp/utils/calculator';
```
bjoernricks, line 26: Would be better to put the functions into `gmp/parser/cvss.js` Yes they are used for the calculator but looking only at the functions they are used to parse cvss vectors only.

gsa/src/gmp/parser.js
```diff
@@ -48,16 +46,9 @@ export const parseText = text => {
...
-      textExcerpt: text.__excerpt,
-    };
+    return text.__text;
   }
-
-  return {
-    text,
-    textExcerpt: '0',
-  };
+  return text;
```
bjoernricks, line 51: What's the intention of changing these lines?

y0urself: There was none. Probably this was accidentally ... i can't remember changing it ..

gsa/src/gmp/utils/calculator.js
```diff
@@ -0,0 +1,676 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {isDefined} from 'gmp/utils/identity';
+
+const toFixed1 = value => {
```
bjoernricks, line 21: That function should get a comment about it's purpose and some examples. I am also not sure which type value expects. Is it a Number or a String? Maybe also use the more common naming scheme `parseXyz`.

gsa/src/gmp/utils/calculator.js
```diff
@@ -0,0 +1,676 @@
...
+  let result = +(
+    Math.round(+(value.toString() + 'e' + 1)).toString() +
```
bjoernricks, line 23: Personally I consider toString() as an implementation detail and wouldn't use it directly. It is more like the __str__ method in Python which also doesn't get called directly. Either use string concatenation or template literals

```javascript
'' + foo

`${foo}` 
```

gsa/src/gmp/utils/calculator.js
```diff
@@ -0,0 +1,676 @@
...
+    'e' +
+    -1
+  );
+  // if no fractional part then add a .0 to the number
+  if (result.toString().indexOf('.') === -1) result = result + '.0';
```
bjoernricks, line 28: Please always use curly braces for if clauses. It increases the readability at lot.
```suggestion
  if (!`${result}`.includes('.')) {
    result = result + '.0';
  }
```

bjoernricks: IMHO using includes instead of indexOf is also better for the readability https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

gsa/src/web/pages/extras/cvsscalculatorpage.js
```diff
@@ -359,21 +350,388 @@ const CvssCalculator = ({gmp, onInteraction, ...props}) => {
...
+  return (
+    <Layout flex="column">
+      <ToolBarIcons />
+      <Divider margin="10px" flex="row" align={['center', 'start']} grow>
+        <CvssV2Calculator grow="1" props={props} />
+        <CvssV3Calculator grow="1" props={props} />
+      </Divider>
+    </Layout>
+  );
 };
```
bjoernricks, line 733: Just for your info. You can use the short syntax of the array function if it just returns something
```suggestion
const CvssCalculator = props => (
  <Layout flex="column">
    <ToolBarIcons />
    <Divider margin="10px" flex="row" align={['center', 'start']} grow
      <CvssV2Calculator grow="1" props={props} />
      <CvssV3Calculator grow="1" props={props} />
    </Divider>
  </Layout>
);
```

gsa/src/web/pages/extras/cvsscalculatorpage.js
```diff
@@ -359,21 +350,388 @@ const CvssCalculator = ({gmp, onInteraction, ...props}) => {
...
+      isDefined(scope) &&
+      isDefined(confidentialityImpact) &&
+      isDefined(integrityImpact) &&
+      isDefined(availabilityImpact)
+    ) {
+      /* only override cvss values and vector if user vector has valid input */
+
+      setState(vals => ({...vals, ...cvssValues, cvssVector: userVector}));
+
+      let newScore = calculateV3Score({
```
bjoernricks, line 491: Any reason for not using const here?

y0urself: Already changed it ... but it is not commited ...

gsa/src/web/pages/extras/cvsscalculatorpage.js
```diff
@@ -359,21 +350,388 @@ const CvssCalculator = ({gmp, onInteraction, ...props}) => {
...
+      confidentialityImpact,
+      integrityImpact,
+      availabilityImpact,
+    });
+    setState(vals => ({...vals, cvssScore: newScore}));
+  };
+
+  const handleInteraction = () => {
+    renewSession();
+  };
```
bjoernricks, line 448: HandleInteraction can be replaced by calling renewSession directly. Or if you want to keep the name just do
```suggestion
  const handleInteraction = renewSession;
```

gsa/src/gmp/models/cve.js
y0urself: Is it possible to have an directory with the same name as a file?

gsa/src/gmp/parser.js
y0urself: Is it possible that husky/lint/whatever has done this?! I really did not touch these lines ...

gsa/src/gmp/models/cve.js
bjoernricks: it is very similar as in Python. You can choose a index.js file which acts like the __init__.py file. E.g. https://github.com/greenbone/gsa/blob/master/gsa/src/gmp/locale/index.js where I split the locale.js file into several files and wanted to keep backwards compatibility with existing imports.

gsa/src/gmp/utils/calculator.js
y0urself: I really did not want to use this function. :/
But all my own rounding Implementations didn't round the score as [expected](https://nvd.nist.gov/vuln-metrics/cvss/v2-calculator).

gsa/src/gmp/parser.js
bjoernricks: I don't think so. That's why I am always using `git add -p` ;-)

gsa/src/gmp/parser.js
```diff
@@ -94,7 +95,7 @@ export const YES_VALUE = 1;
 export const NO_VALUE = 0;
 
 export const parseYesNo = value =>
-  value === '1' || value === 1 ? YES_VALUE : NO_VALUE;
+  value === '1' || value === 1 || value === true ? YES_VALUE : NO_VALUE;
```
bjoernricks, line 98: This function should not be changed. It may have serious side effects.
```suggestion
  value === '1' || value === 1 ? YES_VALUE : NO_VALUE;
```

bjoernricks (commented):
Looks very good. Just one invalid change has slipped in.

y0urself:
So this PR implemented the new feature for CVSS v3 calculation.
Maybe we need to adjust the UI later on ...
Maybe we should add a test, that validates the calculated scores for all possible vectors?

bjoernricks:
> Maybe we should add a test, that validates the calculated scores for all possible vectors?

You are always free to add as much tests as you think is required to validate the functionality.

> Maybe we need to adjust the UI later on ...

Maybe we should ask for some feedback internally.

bjoernricks (approved):

PR #2564, opened 2020-11-09
Refactor CVE Tables to single line table

y0urself:
**What**:

Changing the Tables for CVEs to a one-liner (Name, CVSS BaseVector, Published, Severity, Description).

**Why**:

The old CVE Table format was hard to read and unnecessary complex.

**How**:

Adjusted the Table format.

**Checklist**:

- [ ] Tests


bjoernricks (commented):
Does this consider the CVSSv3 from gvmd already? Is it parsed as cve.cvssBaseVector independent of the CVSS version?

y0urself:
> Does this consider the CVSSv3 from gvmd already? Is it parsed as cve.cvssBaseVector independent of the CVSS version?

No. It parses only CVSSv2 ... I think I need to add a lot of things for CVSSv3 ... should I do that? It wasn't part of the Issue

sarahd93 (changes requested):
Was there a reason for the order of the table columns? Otherwise, I think we should probably use the column order from the jira issue: Name, Description, Published, CVSS, Severity. This feels more consistent with the tables of the other secinfo entities with longer text on the left side and the severity on the right side of the table.

sarahd93 (approved):
Thank you, this looks really nice!

PR #2566, opened 2020-11-09
Fix port range check when editing portlists

swaterkamp:
**What**:
Compare port range start and end via their integer value instead of their string value.

**Why**:
Comparing strings is plain wrong here. It probably happened because of the change of the xml parser.

**How**:
Confirmed by adding a new port range that has a higher first digit at the start than the first digit of its end. No error was thrown and no existing automated test failed.

**Checklist**:

- [N/A] Tests

### DIFF
--- a/gsa/src/web/pages/portlists/component.js
+++ b/gsa/src/web/pages/portlists/component.js
@@ -202,7 +202,10 @@ class PortListComponent extends React.Component {
 
   handleTmpAddPortRange(values) {
     const {port_ranges} = this.state;
-    const {port_range_end, port_range_start, port_type} = values;
+    let {port_range_end, port_range_start, port_type} = values;
+
+    port_range_end = parseInt(port_range_end);
+    port_range_start = parseInt(port_range_start);
 
     this.handleInteraction();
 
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/portlists/component.js
```diff
@@ -216,7 +216,7 @@ class PortListComponent extends React.Component {
     }
 
     // reject port ranges with start value lower than end value
-    if (port_range_start > port_range_end) {
+    if (parseInt(port_range_start) > parseInt(port_range_end)) {
```
sarahd93, line 219: Can you check if we can already parse this in the model? There are other comparisons in the portlist component that also seem to assume that port_range_start and port_range_end are integers.

sarahd93 (changes requested):

gsa/src/web/pages/portlists/component.js
swaterkamp: This concerns the _new_ inputs from the dialog that are compared and it's not linked to the model. `start` and `end` are Int in the model already.

bjoernricks: I see. Could you change the code in this handler method to just store both values as integers from the beginning. Some lines below they are parsed as int again.

bjoernricks (approved):
Thanks

sarahd93 (approved):

PR #2568, opened 2020-11-10
add tests for alert detailspage

sarahd93:
**What**:

Add tests for alert detailspage.

**Why**:

We need tests for the alerts detailspage before we can start the refactoring for Hyperion.

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/alerts/__tests__/detailspage.js
+++ b/gsa/src/web/pages/alerts/__tests__/detailspage.js
@@ -1,0 +1,517 @@
+/* Copyright (C) 2019-2020 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
+ * as published by the Free Software Foundation, either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


saberlynx (approved):

gsa/src/web/pages/alerts/__tests__/detailspage.js
```diff
@@ -0,0 +1,517 @@
...
+import {rendererWith, fireEvent, screen, wait} from 'web/utils/testing';
+
+import Detailspage, {ToolBarIcons} from '../detailspage';
+
+setLocale('en');
+
+if (!isDefined(window.URL)) {
+  window.URL = {};
+}
+window.URL.createObjectURL = jest.fn();
```
bjoernricks, line 44: Is this code used for the tests. As far as I can see no download code paths are tested and therefore createObjectURL is not called. But I am not 100 percent sure.

gsa/src/web/pages/alerts/__tests__/detailspage.js
```diff
@@ -0,0 +1,517 @@
+/* Copyright (C) 2019-2020 Greenbone Networks GmbH
```
bjoernricks, line 1: For the next PR a new file should get the copyright for the current year only.

gsa/src/web/pages/alerts/__tests__/detailspage.js
sarahd93: In the "should call commands" test the export icon is clicked which uses the download component. Removing the code would cause the test to fail because window.URL.createObjectURL is not defined.

PR #2569, opened 2020-11-12
Implement bulk actions for alerts via graphQL

saberlynx:
**What**:

In addition to the title, this PR also attempts to
- Bring alerts listpage in line with task listpage. There are some permissions related handlers that are not removed.
- Fix report formats and details parsing since graphql returns null

**Why**:

**How**:

Previous bulk action tests are passing

**Checklist**:

- [x] Tests


sarahd93 (approved):

gsa/src/gmp/models/alert.js
```diff
@@ -129,6 +129,15 @@ class Alert extends Model {
       ret.filter = Filter.fromObject(object.filter);
     }
 
+    const methDatRepForm = ret.method.data.report_formats;
+
+    if (hasValue(methDatRepForm)) {
+      const methDatRepFormSplit = methDatRepForm.value.split(',');
```
bjoernricks, line 135: Btw. the backend should already provide an array here instead of a CSV list. See our new advises page for hyperion.

gsa/src/web/graphql/__mocks__/alerts.js
```diff
@@ -223,3 +226,50 @@ export const createCloneAlertQueryMock = (alertId = '1', newAlertId = '2') =>
     },
     {id: alertId},
   );
+
+const exportAlertsByIdsResult = {
+  exportAlertsByIds: {
+    exportedEntities: '<get_alerts_response status="200" status_text="OK" />',
```
bjoernricks, line 232: What is that? Why it is an XML string of a GMP response?

gsa/src/web/graphql/alerts.js
```diff
@@ -121,14 +129,30 @@ export const DELETE_ALERTS_BY_IDS = gql`
...
+    deleteAlertsByFilter(filterString: $filterString) {
       ok
     }
   }
 `;
 
+export const EXPORT_ALERTS_BY_FILTER = gql`
+  mutation exportAlertsByFilter($filterString: String) {
+    exportAlertsByFilter(filterString: $filterString) {
+      exportedEntities
```
bjoernricks, line 143: Wouldn't it be better to call that property `exportedEntityIDs` because it just returns a list of IDs?

gsa/src/web/graphql/__mocks__/alerts.js
saberlynx: exportedEntities is always a raw xml string of the GMP response that the frontend transforms into an xml file for download so I just mocked the first line here.

gsa/src/web/graphql/alerts.js
saberlynx: exportedEntities? That's supposed return raw xml. Maybe I'm misunderstanding something big here...

gsa/src/gmp/models/alert.js
saberlynx: Hm. OK. It currently still provides a CSV list because the report_formats method data does not have any special parsing in our fields. but maybe that can go on the list of todos.

PR #2571, opened 2020-11-13
Implement get_alert via graphql and preliminary refactor of alert detailspage

saberlynx:
**What**:

This PR refactors the detailspage to use graphQL related commands such as adding clone/delete/export handlers. It also adds a custom hook to handle exporting entities via graphQL in the same vein as the new bulk action custom hooks.

**Why**:

Manual and unit testing. 

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/entity/useExportEntity.js
+++ b/gsa/src/web/entity/useExportEntity.js
@@ -1,0 +1,57 @@
+/* Copyright (C) 2020 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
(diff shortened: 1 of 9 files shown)


gsa/src/web/entity/useExportEntity.js
```diff
@@ -0,0 +1,57 @@
...
+import {useSelector} from 'react-redux';
+
+import {capitalizeFirstLetter} from 'gmp/utils/string';
+
+import {getUserSettingsDefaults} from 'web/store/usersettings/defaults/selectors';
+
+import {generateFilename} from 'web/utils/render';
+import useUserName from 'web/utils/useUserName';
+
+export const useExportEntity = () => {
```
sarahd93, line 29: I wonder if we should also use the default export here same as for usePagination?

sarahd93 (approved):

PR #2572, opened 2020-11-13
Export default usePagination

saberlynx:
**What**:

Make usePagination as default export.

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/entities/usePagination.js
+++ b/gsa/src/web/entities/usePagination.js
@@ -17,7 +17,6 @@
  */
 
 import {useCallback} from 'react';
-import PropTypes from 'web/utils/proptypes';
 
 const usePagination = ({simpleFilter, filter, pageInfo, refetch}) => {
   const getNext = useCallback(() => {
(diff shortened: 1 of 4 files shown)


gsa/src/web/entities/usePagination.js
```diff
@@ -61,3 +62,12 @@ export const usePagination = ({simpleFilter, filter, pageInfo, refetch}) => {
 
   return [getFirst, getLast, getNext, getPrevious];
 };
+
+usePagination.propTypes = {
+  filter: PropTypes.filter.required,
+  pageInfo: PropTypes.object.required,
+  refetch: PropTypes.func.isRequired,
+  simpleFilter: PropTypes.filter.required,
+};
```
bjoernricks, line 71: I don't think propTypes are used for hooks. But i could be wrong.

saberlynx: Hm not quite sure... I can just remove them.

bjoernricks (approved):

PR #2583, opened 2020-11-19
Refactor cve table

y0urself:
**What**:

* The CVE-Info XML changed. These adjustments account the changes and also shows CVSSv3 Vectors in the CVE-Tables page now.

**Why**:

* The CVE-Info XML changed.

**How**:

**Checklist**:

- [x] Tests


sarahd93 (commented):
I think we should change the cve detailspage to use the correct date. Right now it looks like this: 
`
<div>{_('Published:')}</div>
<div>{dateTimeWithTimeZone(publishedTime)}</div>
<div>{_('Modified:')}</div>
<div>{dateTimeWithTimeZone(updateTime)}</div>
<div>{_('Last updated:')}</div>
<div>{dateTimeWithTimeZone(lastModifiedTime)}</div>
`

This does not have a defined timezone (which makes it default to local time I think) and that makes it impossible to control in the test. I think this is just a leftover and we can use the DateTime component here:

`
<div>{_('Published:')}</div>
<div><DateTime date={publishedTime} /></div>
<div>{_('Modified:')}</div>
<div><DateTime date={updateTime} /></div>
<div>{_('Last updated:')}</div>
<div><DateTime date={lastModifiedTime} /></div>
`

That should fix the tests.

sarahd93 (approved):

PR #2587, opened 2020-11-20
Refactor alerts detailspage to not use withEntityContainer

saberlynx:
**What**:

Refactor detailspage to get rid of `withEntityContainer`. Implemented `useEntitiesTimeout` and `useEntityTimeout` for alert listpage and detailspage

**Why**:

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/entities/useEntitiesTimeout.js
+++ b/gsa/src/web/entities/useEntitiesTimeout.js
@@ -15,6 +15,8 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
+/* eslint-disable react/prop-types */
+
 import {useCallback} from 'react';
 import {isDefined} from 'gmp/utils/identity';
 
--- a/gsa/src/web/entities/useEntitiesTimeout.js
+++ b/gsa/src/web/entities/useEntitiesTimeout.js
@@ -1,0 +1,38 @@
+/* Copyright (C) 2020 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
+ * as published by the Free Software Foundation, either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


sarahd93 (approved):

gsa/src/web/entities/useEntitiesTimeout.js
```diff
@@ -0,0 +1,39 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {useCallback} from 'react';
+import {isDefined} from 'gmp/utils/identity';
+
+const useEntitiesTimeout = ({entities, gmpSettings}) => {
```
bjoernricks, line 22: Sorry for being so late. Any reason for not using `useGmpSettings` hook here directly instead of passing the setting?

gsa/src/web/pages/alerts/detailspage.js
```diff
@@ -186,20 +188,21 @@ const Page = ({onChanged, onError, onInteraction, ...props}) => {
...
+      onCreated={goto_entity_details('alert', {history})}
+      onDeleted={goto_list('alerts', {history})}
+      onDeleteError={showError}
       onDownloaded={handleDownload}
-      onDownloadError={onError}
-      onInteraction={onInteraction}
-      onSaved={onChanged}
+      onDownloadError={showError}
+      onInteraction={renewSessionTimeout}
+      onSaved={() => refetchAlert()}
```
bjoernricks, line 199: should be just `onSaved={refetchAlert}`

bjoernricks: Got it from the comment below. Could you please add the comment here too?

gsa/src/web/entities/useEntitiesTimeout.js
bjoernricks: Personally I find the name a bit confusing because we have useSessionTimeout which returns some functions to handle the timing and this hook just calculates the duration for this timing. It just confused my mind when reading the code and I don't know a better name at the moment.

saberlynx: > Sorry for being so late. Any reason for not using useGmpSettings hook here directly instead of passing the setting?

Nope, no reason. Just didn't think of it because I didn't notice until you pointed it out that gmpSettings is only used by this function and nowhere else as I just focused on translating the page at first.

PR #2601, opened 2020-11-26
Merge GSA-21.04 in GSA-master

swaterkamp:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests


gsa/src/gmp/models/event.js
```diff
@@ -361,8 +361,8 @@ class Event {
           }
           retries = 0;
         } catch (err) {
-          // ical.js raises an exception if the same date occurs twice
-          // See https://github.com/mozilla-comm/ical.js/blob/master/lib/ical/recur_iterator.js#L373
+          // ical raises an exception if the same date occurs twice
+          // See https://github.com/mozilla-comm/ical/blob/master/lib/ical/recur_iterator#L373
```
sarahd93, line 365:
```suggestion
          // See https://github.com/mozilla-comm/ical.js/blob/master/lib/ical/recur_iterator.js#L373
```

sarahd93: I just realized that the .js was accidentally removed in the link as well which breaks it as far as I can tell. We have to change that back in 21.04 as well.

sarahd93 (changes requested):

gsa/src/gmp/models/event.js
y0urself: Yeah sorry for that ... was me probably ...

sarahd93 (approved):

PR #2635, opened 2020-12-16
Add tests for DialogError

swaterkamp:
**What**:
Add tests for DialogError

**Why**:

**How**:

**Checklist**:

- [X] Tests
- [N/A] [CHANGELOG](https://github.com/greenbone/gsa/blob/master/CHANGELOG.md) Entry

### DIFF
--- a/gsa/src/web/components/dialog/__tests__/error.js
+++ b/gsa/src/web/components/dialog/__tests__/error.js
@@ -1,0 +1,61 @@
+/* Copyright (C) 2020 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
+ * as published by the Free Software Foundation, either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+import React from 'react';
+
+import {setLocale} from 'gmp/locale/lang';
+
+import {render, fireEvent, screen} from 'web/utils/testing';
+import Theme from 'web/utils/theme';
+
+import DialogError from '../error';
+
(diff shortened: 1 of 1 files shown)


gsa/src/web/components/dialog/__tests__/error.js
```diff
@@ -0,0 +1,63 @@
+/* Copyright (C) 2018-2020 Greenbone Networks GmbH
```
sarahd93, line 1:
```suggestion
/* Copyright (C) 2020 Greenbone Networks GmbH
```

gsa/src/web/components/dialog/__tests__/error.js
```diff
@@ -0,0 +1,63 @@
...
+    const {element} = render(
+      <DialogError error="foo" onCloseClick={handler} />,
+    );
+
+    const button = screen.getByTitle('Close');
+
+    expect(element).toHaveTextContent('foo');
+
+    fireEvent.click(button);
+    expect(element).not.toHaveTextContent();
```
sarahd93, line 58: I worry that this will always return true. Not sure what the expected behaviour is that we want to test here. This closes the dialog but that is not something this test can cover because the handler is just a mock. I think checking if the handler has been called would be enough here.

gsa/src/web/components/dialog/__tests__/error.js
```diff
@@ -0,0 +1,63 @@
...
+
+    expect(element).toHaveTextContent('foo');
+    expect(element).toHaveStyleRule('background-color', Theme.lightRed);
+  });
+
+  test('should not render with undefined error', () => {
+    const {element} = render(<DialogError onCloseClick={() => {}} />);
+
+    expect(element).toBe(null);
+    expect(element).not.toHaveStyleRule();
```
sarahd93, line 43: I am not sure if this can work like this. I think these functions need an input argument or they always return true.

sarahd93 (changes requested):

gsa/src/web/components/dialog/__tests__/error.js
swaterkamp: I guess checking, if the element is null is already enough. null elements wouldn't have any stylerule anyway, so I just removed that statement.

gsa/src/web/components/dialog/__tests__/error.js
swaterkamp: `.toHaveStyleRule()` fails for this test and `.not.toHaveStyleRule()` doesn't, so those functions do not just return true when they don't get a parameter. But you are right that checking whether the handler was called is probably enough, so I removed that statement.

sarahd93 (approved):

PR #2636, opened 2020-12-16
Add tests for report listpage

sarahd93:
**What**:

Add tests for report listpage. Query icons by title instead of test id to avoid the need for counting to find the right icon.

For example to test a click on an icon instead of doing
```
expect(icons[24]).toHaveAttribute('title', 'Delete all filtered');
fireEvent.click(icons[24]);
```
We can just write
`fireEvent.click(screen.getAllByTitle('Delete all filtered')[0]);`

For checking if an icon is there we can use
`expect(screen.getAllByTitle('Help: Reports')[0]).toBeInTheDocument();`

**Why**:

There are no tests for the report listpage yet.

**How**:

**Checklist**:

- [x] Tests


saberlynx:
Looks awesome! I'm excited for this new paradigm :)

saberlynx (approved):

PR #2650, opened 2021-01-08
Implement modify scan config mutations

saberlynx:
**What**:

Implements modify scan config mutations and tests. 

**Why**:

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/graphql/scanconfigs.js
+++ b/gsa/src/web/graphql/scanconfigs.js
@@ -1,4 +1,4 @@
-/* Copyright (C) 2020 Greenbone Networks GmbH
+/* Copyright (C) 2020-2021 Greenbone Networks GmbH
  *
  * SPDX-License-Identifier: GPL-2.0-or-later
  *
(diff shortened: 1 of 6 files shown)


gsa/src/web/graphql/scanconfigs.js
```diff
@@ -448,3 +510,213 @@ export const useDeleteScanConfigsByIds = options => {
...
+    [
+      queryModifyScanConfigSetName,
+      queryModifyScanConfigSetComment,
+      queryModifyScanConfigSetScannerPreference,
+      queryModifyScanConfigSetFamilySelection,
+    ],
+  );
+
+  return modifyScanConfig;
+};
```
bjoernricks, line 629: If name, comment, preferences and family selection are set only in one go the backend api should be changed to have only one mutation to change them at the same time. That would make the frontend api easier to implement and to read. Maybe it was also a bad idea to split that in python-gvm.

y0urself: We could recreate a single call in `python-gvm` and integrate it into the backend easily i guess ...

saberlynx: > If name, comment, preferences and family selection are set only in one go the backend api should be changed to have only one mutation to change them at the same time. That would make the frontend api easier to implement and to read. Maybe it was also a bad idea to split that in python-gvm.

And also, currently scanner preferences can only be sent one at a time so I'm pretty sure pressing "save scan config" necessitates 20+ graphql calls. That is quite suboptimal. Although in my experience, sending one huge command (or requesting a list of scan configs with a lot of fields) also causes timeout sometimes.

PR #2659, opened 2021-01-13
Implement schedule detailspage queries and mutations

saberlynx:
**What**:
Convert schedule detailspage to use graphql and add some tests. Adjusted detailspage tests.

**Why**:

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/schedules/__tests__/detailspage.js
+++ b/gsa/src/web/pages/schedules/__tests__/detailspage.js
@@ -300,24 +300,25 @@ describe('Schedule Detailspage tests', () => {
     expect(resultFunc).toHaveBeenCalled();
     expect(permissionResult).toHaveBeenCalled();
 
-    const icons = screen.getAllByTestId('svg-icon');
-
-    expect(icons[3]).toHaveAttribute('title', 'Clone Schedule');
-    fireEvent.click(icons[3]);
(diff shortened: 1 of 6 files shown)


gsa/src/web/pages/schedules/__tests__/detailspage.js
```diff
@@ -155,179 +145,149 @@ describe('Schedule Detailspage tests', () => {
...
-    const tabs = screen.getAllByTestId('entities-tab-title');
+    expect(renewSessionResult).toHaveBeenCalled();
 
-    expect(tabs[1]).toHaveTextContent('Permissions');
-    fireEvent.click(tabs[1]);
-
-    expect(baseElement).toHaveTextContent('No permissions available');
+    expect(baseElement).toHaveTextContent('No user tags available');
   });
-
```
sarahd93, line 259: I think the "should render permissions tab" tab test case has gone missing.

sarahd93 (changes requested):

gsa/src/web/pages/schedules/__tests__/detailspage.js
saberlynx: Oops. Added it back.

sarahd93 (approved):

PR #2669, opened 2021-01-21
Host detailspage test

sarahd93:
**What**:

Add tests for host detailspage.

**Why**:

There weren't any tests yet and we need them before we can start refactoring for Hyperion.

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/hosts/__tests__/detailspage.js
+++ b/gsa/src/web/pages/hosts/__tests__/detailspage.js
@@ -1,0 +1,582 @@
+/* Copyright (C) 2021 Greenbone Networks GmbH
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/hosts/__tests__/detailspage.js
```diff
@@ -0,0 +1,579 @@
...
+
+// setup
+
+setLocale('en');
+
+if (!isDefined(window.URL)) {
+  window.URL = {};
+}
+window.URL.createObjectURL = jest.fn();
+
```
bjoernricks, line 46: Should be put in the beforeEach function block.

bjoernricks (changes requested):
Just a small change for the new Jest 26 version coming with react-scripts 4.0

gsa/src/web/pages/hosts/__tests__/detailspage.js
bjoernricks: Seems to be not necessary at all.

bjoernricks (approved):

PR #2684, opened 2021-01-27
Implement policy component mutations

saberlynx:
**What**:

**Why**:

**How**:

Manual and auto testing. 

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/graphql/policies.js
+++ b/gsa/src/web/graphql/policies.js
@@ -247,7 +247,7 @@ export const useLazyGetPolicy = () => {
         fetchPolicy: 'no-cache', // do not cache, since this is used when a change is saved
       })
       .then(response => {
-        if (isDefined(response?.data?.policy) {
+        if (isDefined(response?.data?.policy)) {
           policy = Policy.fromObject(response?.data?.policy);
         }
 
--- a/gsa/src/web/graphql/policies.js
+++ b/gsa/src/web/graphql/policies.js
@@ -18,7 +18,13 @@
  */
 import {useCallback} from 'react';
 
-import {gql, useQuery, useMutation, useLazyQuery, useApolloClient} from '@apollo/client';
+import {
+  gql,
+  useQuery,
+  useMutation,
+  useLazyQuery,
+  useApolloClient,
+} from '@apollo/client';
 
 import CollectionCounts from 'gmp/collection/collectioncounts';
 import Policy from 'gmp/models/policy';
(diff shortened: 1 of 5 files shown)


gsa/src/web/graphql/policies.js
```diff
@@ -165,3 +181,56 @@ export const useClonePolicy = options => {
...
+        fetchPolicy: 'no-cache', // do not cache, since this is used when a change is saved
+      })
+      .then(response => {
+        const policy = Policy.fromObject(response?.data?.policy);
+
+        return policy;
+      });
+
+  return loadPolicy;
+};
```
bjoernricks, line 219:
```suggestion
export const useLazyGetPolicy = () => {
  const client = useApolloClient();
  let policy;
  const getPolicy = policyId =>
    client
      .query({
        query: GET_POLICY,
        variables: {id: policyId},
        fetchPolicy: 'no-cache', // do not cache, since this is used when a change is saved
      })
      .then(response => {
        if (isDefined(response?.data?.policy) {
          policy = Policy.fromObject(response?.data?.policy);
        }

        return policy;
      });

  return [getPolicy, policy];
};
```

gsa/src/web/pages/policies/component.js
```diff
@@ -39,6 +41,12 @@ import {shorten} from 'gmp/utils/string';
 
 import EntityComponent from 'web/entity/component';
 
+import {
+  useCreatePolicy,
+  useImportPolicy,
+  useLoadPolicyPromise,
```
bjoernricks, line 47:
```suggestion
  useLazyGetPolicy,
```

gsa/src/web/pages/policies/component.js
```diff
@@ -109,6 +118,10 @@ const PolicyComponent = ({
     importDialogVisible: false,
   });
 
+  const [createPolicy] = useCreatePolicy();
+  const loadPolicyPromise = useLoadPolicyPromise();
```
bjoernricks, line 122:
```suggestion
  const [getPolicy] = useLazyGetPolicy();
```

gsa/src/web/pages/policies/component.js
```diff
@@ -603,12 +625,11 @@ const PolicyComponent = ({
       }),
     );
 
-    return gmp.policy
-      .get({id: policyId})
-      .then(response => {
+    return loadPolicyPromise(policyId)
```
bjoernricks, line 628:
```suggestion
    return getPolicy(policyId)
```

bjoernricks (changes requested):

gsa/src/web/graphql/policies.js
saberlynx: I don't think this works because queries are not promises and we cannot use the .then(() => ...)

gsa/src/web/pages/policies/component.js
```diff
@@ -503,11 +524,12 @@ const PolicyComponent = ({
...
 
-  const handleImportPolicy = data => {
+  const handleImportPolicy = async data => {
     handleInteraction();
 
-    return gmp.policy
-      .import(data)
+    const readUploadedXml = readFileToText(data.xml_file);
+
+    return importPolicy(await readUploadedXml)
```
bjoernricks, line 532: Is there any reason for using async/await here? From looking at the code I would expect that is works without. This is not a criticism of your code or even an intention to rewrite it I am just curious :-) It's the first time we use async/await in our GSA code besides tests.

bjoernricks (approved):

PR #2693, opened 2021-01-29
refactor host component to function component

sarahd93:
**What**:

Host component is a class component. Refactor it to be a function component instead.

**Why**:
This is needed to be able to use hooks.

**How**:

**Checklist**:
Checked gsa to make sure dialogs and other functions on the host listpage and detailspage still work.

- [ ] Tests

### DIFF
--- a/gsa/src/web/pages/hosts/component.js
+++ b/gsa/src/web/pages/hosts/component.js
@@ -34,15 +34,28 @@ import SelectionType from 'web/utils/selectiontype';
 import stateReducer, {updateState} from 'web/utils/stateReducer';
 import useGmp from 'web/utils/useGmp';
 
-const HostComponent = props => {
+const HostComponent = ({
+  children,
+  createtarget,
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/hosts/component.js
```diff
@@ -31,72 +31,66 @@ import TargetComponent from 'web/pages/targets/component';
...
 import PropTypes from 'web/utils/proptypes';
 import SelectionType from 'web/utils/selectiontype';
-import withGmp from 'web/utils/withGmp';
+import stateReducer, {updateState} from 'web/utils/stateReducer';
+import useGmp from 'web/utils/useGmp';
 
-class HostComponent extends React.Component {
-  constructor(...args) {
-    super(...args);
+const HostComponent = props => {
```
saberlynx, line 37: Nice! Maybe destructure props here?

saberlynx (approved):

PR #2694, opened 2021-01-29
Update copyright year

y0urself:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests


swaterkamp (approved):
A random sample of all those files looks ok and it is likely, the whole thing worked correctly for all files. Last time the about page made some problems, but it also seems ok.

y0urself:
> A random sample of all those files looks ok and it is likely, the whole thing worked correctly for all files. Last time the about page made some problems, but it also seems ok.

I guess looking into all 1500 files would fill up your working day ... :D

swaterkamp:
Yes, it probably would...

PR #2702, opened 2021-02-04
Refactor host detailspage

sarahd93:
**What**:

Refactor host detailspage to use graphql queries and mutations.

**Why**:

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/pages/hosts/detailspage.js
+++ b/gsa/src/web/pages/hosts/detailspage.js
@@ -74,7 +74,13 @@ import CreateIcon from 'web/entity/icon/createicon';
 import DeleteIcon from 'web/entity/icon/deleteicon';
 import EditIcon from 'web/entity/icon/editicon';
 
-import {useGetHost} from 'web/graphql/hosts';
+import useExportEntity from 'web/entity/useExportEntity';
+
+import {
+  useDeleteHost,
+  useExportHostsByIds,
+  useGetHost,
+} from 'web/graphql/hosts';
 import {useGetPermissions} from 'web/graphql/permissions';
 
 import PropTypes from 'web/utils/proptypes';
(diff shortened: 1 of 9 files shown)


gsa/src/web/pages/hosts/identifiers.js
```diff
@@ -56,64 +56,70 @@ const filter_identifiers = (identifiers, latest = true) => {
   if (!latest || !isDefined(identifiers) || identifiers.length === 0) {
     return identifiers;
   }
-  const last_id = identifiers[0].source.id;
-  return identifiers.filter(identifier => identifier.source.id === last_id);
+  const last_id = identifiers[0].sourceId;
+  return identifiers.filter(identifier => identifier.sourceId === last_id);
```
bjoernricks, line 60:
```suggestion
  const lastId = identifiers[0].sourceId;
  return identifiers.filter(identifier => identifier.sourceId === lastId);
```

gsa/src/web/pages/hosts/detailspage.js
```diff
@@ -263,93 +274,153 @@ Details.propTypes = {
...
-      onTargetCreateError={onError}
-      onCreated={goto_host}
-      onDeleted={goto_list('hosts', props)}
-      onDownloaded={onDownloaded}
-      onDownloadError={onError}
-      onIdentifierDeleted={onChanged}
-      onIdentifierDeleteError={onError}
-      onInteraction={onInteraction}
-      onSaved={onChanged}
+      onTargetCreated={goto_entity_details('target', {history})}
```
bjoernricks, line 340: Shouldn't it be goto_details here too? I don't know the difference between both functions atm.

bjoernricks (changes requested):

gsa/src/web/pages/hosts/detailspage.js
sarahd93: goto_details is the old function and goto_entity_details is the new one for entities that use graphql mutations (which directly return the id for the new entity). So it is correct for onTargetCreated to use goto_entity_details because creating a target already works with graphql (because we needed it for the task dialog). Since the create mutation for hosts is not implemented in gsa yet we still need to use goto_details for onCreated for now.

bjoernricks: Thanks for the detailed explanation!

bjoernricks (approved):

PR #2704, opened 2021-02-05
Refactor form validation

bjoernricks:
**What**:

Refactor the form validation.

**Why**:

After looking at the rows per page user setting for a bugfix I had a quick look at the useFormValidation hook. The original version had a major downside by tying to synchronize states. This should **always** be avoided. State should be handled at one place only.

**How**:

Added new tests and run all other unit tests.

**Checklist**:

- [x] Tests


saberlynx (approved):
This looks awesome!

PR #2721, opened 2021-02-12
Use graphql for notes entity

swaterkamp:
**What**:
Implement graphql usage and tests for the notes entity.

**Why**:
To use it with the new architecture

**How**:
Automated tests

**Checklist**:

- [X] Tests

### DIFF
--- a/gsa/src/gmp/models/note.js
+++ b/gsa/src/gmp/models/note.js
@@ -15,10 +15,10 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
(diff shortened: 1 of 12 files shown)


gsa/src/gmp/models/note.js
```diff
@@ -68,6 +93,10 @@ class Note extends Model {
 
     return ret;
   }
+
+  isInTrash() {
+    return this.trash === 1;
```
bjoernricks, line 98: should be YES_VALUE

swaterkamp: I'll have to make another PR for this, unfortunately. Let me know if there is something else you would like to have changed.

PR #2733, opened 2021-02-16
Fix ScanConfigs and Policies NVT oid->id

y0urself:
**What**:

* Change occurrences of `oid` in `ScanConfigs` and `Policies` to `id` due to latest changes in Hyperion.

**Why**:

* changed it in Hyperion

**How**:

**Checklist**:

- [ ] Tests


sarahd93 (approved):

bjoernricks (approved):
The whole preference handling in the models (scan configs and policies) looks a bit weird now. IMHO this should be refactored and maybe some parts can also be moved to hyperion already. For example the differentiation between scanner and nvt preferences should be done in the backend already and copying the original nvt should be necessary anymore.

```js
          const nvt = {...pref.nvt};
          pref.nvt = nvt;
          pref.nvt.id = preference.nvt.id;
```
makes no sense to me.

PR #2742, opened 2021-02-20
Refactoring CVEs Listpage to use GraphQL

y0urself:
**What**:

* Change CVEs Listpage to use GraphQL
* Remove withEntitiesContainer of both CVEs list- and detailspage
* Extend tests

**Why**:

**How**:

**Checklist**:

- [x] Tests


gsa/src/web/pages/cves/filterdialog.js
```diff
@@ -25,28 +25,8 @@ const SORT_FIELDS = [
...
-    name: 'confidentiality_impact',
-    displayName: _l('Confidentiality Impact'),
-  },
-  {
-    name: 'integrity_impact',
-    displayName: _l('Integrity Impact'),
-  },
-  {
-    name: 'availability_impact',
-    displayName: _l('Availability Impact'),
```
bjoernricks, line 49: Why has this code been removed?

y0urself: We refactored the CVE table https://github.com/greenbone/gsa/pull/2583 here.
We do not list these values in the table anymore and I missed removing that filter.

bjoernricks (approved):

PR #2779, opened 2021-03-09
Add option to allow to scan simultaneous IPs to targets

saberlynx:
**What**:

Add option to allow to scan simultaneous IPs to targets

**Why**:

It is now possible to enable/diable scanning multiple ips from the same host.

**How**:

Unit tests

**Checklist**:

- [x] Tests


gsa/src/gmp/models/target.js
```diff
@@ -67,6 +67,8 @@ class Target extends Model {
       );
     }
 
+    ret.allow_simultaneous_ips = parseYesNo(element.allow_simultaneous_ips);
```
bjoernricks, line 70: Could we use camelCase for the variable at all places? I know it isn't consistent atm and will create some extra effort in the commands methods but only because we never got the time to refactor all models and dialogs to use camelCase instead of underscore.
```suggestion
    ret.allowSimultaneousIPs = parseYesNo(element.allow_simultaneous_ips);
```

PR #2781, opened 2021-03-09
Fix setting comments of BPM nodes

swaterkamp:
**What**:
Fix setting comments of BPM nodes.

**Why**:
The correct value in the textfield wasn't shown and got overwritten by each single key press.

**How**:
Create a node with a proper comment and edit a node's comment. In both cases the current comment value was shown to the user and after clicking save it was applied to the node.

**Checklist**:

- [N/A] Tests


bjoernricks (commented):
I would love to see a test for this.

swaterkamp:
Afaik we don't have a testing concept for dialogs, yet.

bjoernricks (approved):
I am feeling not very well approving this PR without a test.

swaterkamp:
I'll think about how to test dialogs today. There might be something like a prototype for that in master.

PR #2791, opened 2021-03-15
Don't show empty sections in result details

swaterkamp:
**What**:
Parse empty strings in NVT tags as undefined.

**Why**:
Instead of undefined information we got an empty string before. This fixes showing empty sections on the result details.

**How**:
Check data in dev-tools and visually inspect whether the sections are no longer visible if there is no information. Added an automated test for the empty string case.

**Checklist**:

- [X] Tests


bjoernricks (approved):
I am fine with this change too but could you please also fix the models to set undefined if an empty string is used in these tags. Such things should to be fixed in the parsing part (model) to avoid having to implement this meaning of empty string at several places in the view.

bjoernricks (changes requested):
Clicked on the wrong item. Should have been a Request changes.

gsa/src/gmp/models/nvt.js
```diff
@@ -33,7 +33,8 @@ const parseTags = tags => {
     const splitted = tags.split('|');
     for (const t of splitted) {
       const [key, value] = split(t, '=', 1);
-      newTags[key] = value;
+      const newValue = value === '' ? undefined : value;
```
bjoernricks, line 36:
```suggestion
      const newValue = isEmpty(value) ? undefined : value;
```

bjoernricks (changes requested):

sarahd93 (dismissed):

sarahd93 (approved):

PR #2792, opened 2021-03-15
Change parsing and display of solution for log NVTs

swaterkamp:
**What**:
Parsing and display of solution is changed to be able to:
- show type and description independently
- don't show unknown type for log NVTs
- parse empty string as undefined instead

**Why**:
Log NVTs don't always have a solution, sometimes only an unknown solution type, or there is just a description but no type. This PR makes showing these information more dynamic, reducing unnecessary information in the details.

**How**:
Visual inspection and updated tests for NVT model

**Checklist**:

- [X] Tests


bjoernricks (approved):

sarahd93 (changes requested):
After discussing this we came to the conclusion that this is not the behaviour we actually want. Since some log results can have a solutiontype we do want to show the solution section in this case. But for undefined solution types the solution section should not be shown for log results as well as other severities.

gsa/src/gmp/models/nvt.js
```diff
@@ -157,6 +149,25 @@ class Nvt extends Info {
...
+          description: solutionText === '' ? undefined : solutionText,
+          method: solutionMethod === '' ? undefined : solutionMethod,
+        };
+      } else {
+        ret.solution = {
+          type: solutionType,
+          description: solutionText,
+          method: solutionMethod,
+        };
+      }
```
bjoernricks, line 168: Please keep the existing functions for such cases in mind and keep in mind we should have not several meanings for the same thing. No solution type, text, method should mean it has an undefined value.
```suggestion
      ret.solution = {
        type: isEmpty(solutionType) ? undefined : solutionType,
        description:  isEmpty(solutionText) ? undefined : solutionText,
        method: isEmpty(solutionMethod) ? undefined : solutionMethod,
      };
```

bjoernricks (dismissed):

sarahd93 (dismissed):

sarahd93 (dismissed):

sarahd93 (approved):

Issue #427, opened 2018-03-02
NG: Formatting is not preserved in "Log Method" case


cfi-gb:
This is a follow-up of #390 and the PR #397 solving it:

There is another edge/corner case where a NVT is using the ``script_tag(name:"vuldetect", value:""`` tag but is only sending a ``log_message()`` (0.0 severity).

Once this is happening instead of the static and already correctly formatted ``Vulnerability Detection Method`` a ``Log Method`` (static as well) without the formatting is shown like seen below:

![screenshot_2018-03-02_08-42-49](https://user-images.githubusercontent.com/34644702/36889193-3f9783b8-1df9-11e8-9063-ed1fe2eca1ec.png)

bjoernricks:
@cfi-gb is this fixed in gsa-8.0 already?

cfi-gb:
I don't have a gsa-8.0 setup at hand to do any tests for this but on master this is still happening:

![Screenshot_2019-04-09_09-56-17](https://user-images.githubusercontent.com/34644702/55783060-e3d8f780-5aad-11e9-8610-1129ad7e0607.png)

where the reproducer is a scan against a HTTP server with the following patch applied:

```
diff --git a/scripts/http_version.nasl b/scripts/http_version.nasl
index d1091b9c62..968488ae5d 100644
--- a/scripts/http_version.nasl
+++ b/scripts/http_version.nasl
@@ -52,6 +52,12 @@ if(description)
 
   script_tag(name:"summary", value:"This detects the HTTP Server's type and version.");
 
+  script_tag(name:"vuldetect", value:"This is a test:
+
+  - Test 1
+
+  - Test 2");
+
   script_tag(name:"qod_type", value:"remote_banner");
 
   exit(0);
@@ -118,6 +124,8 @@ report = 'The remote web server type is :\n\n';
 
 port = get_http_port( default:80 );
 
+log_message(port:port, data:"This is a test");
+
 # nb: Always keep http_get() before http_open_socket() as the first could
 # fork with multiple vhosts and the child's would share the same socket
 # causing race conditions and similar.
```

cfi-gb:
For the records, this has / should have been fixed in #2793.

PR #2793, opened 2021-03-15
Change the HTML Tag of detection method to Preformatted

y0urself:
**What**:

* Change the HTML Tag of Detection Method to Preformatted, so it does not remove the newlines ...

image

**Why**:

* Was badly formatted before ...

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Deleting a single entity now removes its ID from store [#1839](https://github.com/greenbone/gsa/pull/1839)
 
 ### Fixed
+- Fixed formatting of Detection Method in Results [#2793](https://github.com/greenbone/gsa/pull/2793)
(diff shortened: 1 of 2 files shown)


swaterkamp (approved):

bjoernricks (changes requested):
Please please add a test for this change. I know it can be cumbersome for this small change but we need to ensure not to break this formatting in future accidentally.

y0urself:
> Please please add a test for this change. I know it can be cumbersome for this small change but we need to ensure not to break this formatting in future accidentally.

Sorry I missed to do this a draft ... was not ready!

y0urself:
> Please please add a test for this change. I know it can be cumbersome for this small change but we need to ensure not to break this formatting in future accidentally.

The merged test from https://github.com/greenbone/gsa/pull/2794 should cover these lines.

bjoernricks (approved):
I don't see a test that checks the formatting of the test but nevertheless this PR should be merged now.

cfi-gb:
Fixes: #427

PR #2813, opened 2021-03-23
Use params_value instead of params_value_bool for allow_simultaneous_ips

ArnoStiefvater:
**What**:

Use `params_value()` instead of `params_value_bool()` for getting `allow_simultaneous_ips`.

**Why**:

Although the underlying value is bool we should not use params_value_bool for retrieving the value of reverse_lookup_only because it would break the mechanism of setting a default value for the setting.

params_value() returns a string ("0" or "1") if there is an value and NULL if not. params_value_bool() return 0 if the value is "0"
OR there is no value. This means that if we want to be able to set a default if there is no value (NULL) we need to use params_value() else the information is lost.

Fix crash.

**How**:

Created target with and without PR. Without PR gsad crashes. 

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 - Added CVSS date to NVT details [#2802](https://github.com/greenbone/gsa/pull/2802)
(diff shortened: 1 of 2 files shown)


bjoernricks (approved):
My fault. Thanks for fixing it.

PR #2823, opened 2021-03-26
Refactor results detailspage

sarahd93:
**What**:
Refactor results detailspage to use graphql queries instead of gmp commands.
The overrides and delta results section of the detailspage have not been refactored yet and will follow in later PRs.

**Why**:
The results detailspage is still using gmp commands.
Overrides will be added once they are working correctly again and the behaviour can be tested properly. Delta results will be added in another separate PR together with the deltareportspage.
WithEntityContainer is still used for the export until useEntityExport has been adjusted for single entity exports. 
CVE results might still cause problems and will be adjusted after changes to the NVTs.

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/gmp/models/result.js
+++ b/gsa/src/gmp/models/result.js
@@ -54,32 +54,35 @@ class Result extends Model {
 
     const {
       detectionResult,
+      host = {},
       name,
       notes,
       nvt = {},
       overrides,
-      report,
+      originalSeverity,
+      qod = {},
+      severity,
       task,
       tickets,
     } = object;
 
+    copy.host = {
+      name: host.ip,
+      id: isDefined(host.id) && !isEmpty(host.id) ? host.id : undefined,
+      hostname: isDefined(host.hostname) ? host.hostname : '',
+    };
+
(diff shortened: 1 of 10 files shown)


gsa/src/gmp/models/result.js
```diff
@@ -49,6 +49,79 @@ export class Delta {
...
+      originalSeverity,
+      qod = {},
+      severity,
+      task,
+      tickets,
+    } = object;
+
+    copy.host = {
+      name: host.ip,
+      id: isDefined(host.id) && !isEmpty(host.id) ? host.id : undefined,
```
saberlynx, line 71: Should this be `hasValue`? I'm under the impression that hyperion backend always sends `null` and never `undefined`. `null` will return true with `isDefined` I think.

bjoernricks: isEmpty should return true for null

bjoernricks: yes currently i doesn't check the null case. it should be extend to support the null case.

saberlynx (dismissed):

saberlynx (approved):

PR #2866, opened 2021-04-22
Adjust GSA for result object changes in hyperion

saberlynx:
**What**:

Adjust results parsing to account for changes found here https://github.com/greenbone/hyperion/pull/92 Only affects detailspage for now as listpage is not implemented for hyperion.

**Why**:

Because result detailspage will be broken without adjusting for backend changes.

**How**:

**Checklist**:

- [x] Tests


gsa/src/web/pages/results/details.js
```diff
@@ -102,14 +102,14 @@ DerivedDiff.propTypes = {
...
   const result = entity;
 
-  const {nvt} = result;
-  const {id: nvtId, tags, solution} = nvt;
+  const {information} = result;
+  const {id: infoId, tags, solution} = information;
 
-  const is_oval = hasValue(nvtId) && nvtId.startsWith('oval:');
-  const hasDetection = hasValue(result.detectionResult);
+  const is_oval = hasValue(infoId) && infoId.startsWith('oval:');
```
bjoernricks, line 108: all variables we are touching with code changes should be converted to camelCase. Could you change it to `isOval`?

bjoernricks (changes requested):
Looks very good to me. Only the underscore variable should be renamed additionally.

sarahd93 (approved):

PR #2891, opened 2021-04-30
Fix setting Radius secret key in dialog

swaterkamp:
**What**:
Use value of dialog internal state for secret key in radius dialog

**Why**:
The value was fixed and unchangeable.

**How**:

**Checklist**:

- [x] Tests


swaterkamp:
This is a draft it has yet to be determined.

sarahd93 (approved):

PR #2898, opened 2021-05-04
Adjust target credentials to reflect changes in Hyperion

saberlynx:
**What**:

Adjust GSA based on changes here https://github.com/greenbone/hyperion/pull/113

**Why**:

**How**:

Manually tested that target credentials are parsed correctly on target list and detailspage, adjusted unit tests and that creating and modifying targets with credentials is still possible.

**Checklist**:

- [ ] Tests


gsa/src/gmp/models/target.js
```diff
@@ -109,13 +109,18 @@ class Target extends Model {
...
-      if (hasValue(cred) && hasValue(cred.id)) {
-        ret[name] = parseModelFromObject(cred, 'credential');
-      } else {
-        delete ret[name];
+    if (hasValue(ret.credentials)) {
+      // in some instances we do not query for target credentials
+      for (const key of Object.keys(HYPERION_TARGET_CREDENTIAL_NAMES)) {
+        const cred = ret.credentials[HYPERION_TARGET_CREDENTIAL_NAMES[key]];
+        if (hasValue(cred) && hasValue(cred.id)) {
+          ret[key] = parseModelFromObject(cred, 'credential');
```
sarahd93, line 117: We should use the new structure for the model as well. No need to add separate fields for the different credentials again. It should be the same structure in all components.

sarahd93 (changes requested):

sarahd93 (dismissed):

sarahd93 (approved):

PR #2914, opened 2021-05-17
Fix unable to edit schedules with number only name

nichtsfrei:
When adding a new schedule with a numbers only name `1234` and daily
re occurrence and trying to edit the new schedule than a:
```
TypeError: e.replace is not a function
```
occurs because name is a number and not a string.

To fix that name.toString() is called to explicitly cast it to string.

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/web/pages/schedules/dialog.js
+++ b/gsa/src/web/pages/schedules/dialog.js
@@ -328,7 +328,10 @@ class ScheduleDialog extends React.Component {
         interval: isPreDefined ? 1 : interval,
         monthdays: setMonthydays ? monthdays : undefined,
         weekdays: setWeekdays ? weekdays : undefined,
-        summary: name,
+        // convert name to string explicitly to not run into:
+        // `TypeError: e.replace is not a function`
+        // when name is just numbers.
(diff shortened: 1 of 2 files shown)


swaterkamp (changes requested):
Please add new changelog entries to the top of the list

swaterkamp (dismissed):

gsa/src/web/pages/schedules/dialog.js
```diff
@@ -328,7 +328,10 @@ class ScheduleDialog extends React.Component {
         interval: isPreDefined ? 1 : interval,
         monthdays: setMonthydays ? monthdays : undefined,
         weekdays: setWeekdays ? weekdays : undefined,
-        summary: name,
+        // convert name to string explicitly to not run into:
+        // `TypeError: e.replace is not a function`
+        // when name is just numbers.
+        summary: name.toString(),
```
bjoernricks, line 334:
```suggestion
        summary: `${name}`,
```

bjoernricks (approved):

PR #2949, opened 2021-05-28
Properly space and linebreak roles and groups in users table row

swaterkamp:
**What**:
Properly space and linebreak roles and groups in users table row by not using the Divider component, but HorizontalSep with a new wrap prop.

**Why**:
Adding too many roles and/or groups to a user grew the table row too wide and move parts of it out of screen. It was also not very readable even without growing > 100%/width

**How**:
Visually inspect how roles and groups are displayed. Add test for wrapped HorizontalSep

**Checklist**:

- [X] Tests
- [X] Labels for ports to other branches

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Set SameSite=strict for the session cookie to avoid CSRF [#2948](https://github.com/greenbone/gsa/pull/2948)
 
 ### Changed
+- Properly space and linebreak roles and groups in users table row [#2949](https://github.com/greenbone/gsa/pull/2949)
+- Make HorizontalSep component wrappable [#2949](https://github.com/greenbone/gsa/pull/2949)
 - Use greenbone sensor as default scanner type when opening the dialog if available [#2867](https://github.com/greenbone/gsa/pull/2867)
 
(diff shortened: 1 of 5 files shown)


bjoernricks (changes requested):
> Properly space and linebreak roles and groups in users table row by not using the Divider component, which prevents wraps in flexbox. Instead add a ,&nbsp; to each entry, except the last one, and let flexbox do it's magic (using the Layout component).

Why not using the [wrap](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#flex-wrap) prop at the divider?

swaterkamp:
This might be a slightly misleading commit message. The `<Divider>` is a `<Layout>` and should be able to handle the `wrap`, of course. When introducing the comma between the items, however, I used `<Layout>` instead, because otherwise I would have had to wrap each item and its comma into another span or so, to get the spacing right. Switching to `<Layout>` kept the component tree a bit simpler and easier to read.

bjoernricks (changes requested):
For me your solutions looks to complicated. If this can't be fixed with CSS (especially flexbox) alone it should be abstracted into a Divider like component.

swaterkamp:
Using `,` is just making this so complicated. I now decided on using a wrappable HorizontalSep, to keep the solution as simple as possible.

sarahd93 (approved):

PR #2954, opened 2021-06-01
Allow to add ssh elevate credential to target dialog and display elevate credential in details

saberlynx:
**What**:

- Allow creating SSH elevate credentials if SSH credential is set
- If we unset an SSH credential, the SSH elevate credential will be unset as well
- If a target is in use, neither SSH nor SSH elevate credential will be editable
- Can change SSH elevate credential if target is not in use
- Show SSH elevate credentials in targets
- Allow creating new SSH elevate credentials if target is not in use

**Why**:

**How**:

**Checklist**:

Visual confirmation. Will add tests in following PR.

- [ ] Tests


gsa/src/web/pages/targets/dialog.js
```diff
@@ -107,6 +118,10 @@ const NEW_SNMP = {
   types: SNMP_CREDENTIAL_TYPES,
 };
 
+const ElevatePrivilegeText = styled(Layout)`
+  color: ${Theme.darkRed};
```
sarahd93, line 122: Actually, this should not be red. I talked about this with @bjoernricks and we realized that was an error in the mockup.

sarahd93 (changes requested):

gsa/src/web/pages/targets/dialog.js
saberlynx: Will just regular black work?

bjoernricks: yes just use the standard black.

bjoernricks (approved):

sarahd93 (approved):
Awesome work, thank you!

PR #2955, opened 2021-06-01
Add origin and date from NVT severities to parseObject() in NVT model

swaterkamp:
**What**:
Add origin and date from NVT severities to parseObject() in NVT model.

**Why**:
They are not parsed properly and don't show up in the NVT details

**How**:
Visual inspection, if the info is shown. Automatic tests are still passing.

**Checklist**:

- [x] Tests


saberlynx (changes requested):
Can you add some tests for the parsing you wrote? Thanks!

saberlynx (approved):

PR #2981, opened 2021-06-10
Actions for all branches | Drop CircleCI

bjoernricks:
**What**:

Prepare workflow files for using them in all branches. We need to backport the files to the branches afterwards.

**Why**:

We want to use github actions for all branches.

**How**:

**Checklist**:

- [ ] Tests


y0urself:
Thx for taking over this!

y0urself (dismissed):

y0urself:
@bjoernricks I am actually not sure if we should set `path` in the workflows ...
When we require a workflow, that will never be triggered because the code in the set `path` has not been changed, the check will never complete.
(At least I think that it is the problem here ...)

bjoernricks:
> @bjoernricks I am actually not sure if we should set `path` in the workflows ...
> When we require a workflow, that will never be triggered because the code in the set `path` has not been changed, the check will never complete.
> (At least I think that it is the problem here ...)

I really thought github would be intelligent enough to get that right. We really need to think about splitting gsad and gsa into two separate repos.

y0urself:
> > @bjoernricks I am actually not sure if we should set `path` in the workflows ...
> > When we require a workflow, that will never be triggered because the code in the set `path` has not been changed, the check will never complete.
> > (At least I think that it is the problem here ...)
> 
> I really thought github would be intelligent enough to get that right. We really need to think about splitting gsad and gsa into two separate repos.


I am not sure if this is really the case, but thats the only thing that makes sense to me, because these checks are not running at all ...

Yeah. We really should get that split up done. In the end with a change of the backend we need to get gsad out of gsa anyway, so why not seperate right now.

y0urself (approved):

Issue #2847, opened 2021-04-11
TLS Certificates identified by issuer?


Virsacer:
### Expected behavior

Use the actual CN of the certificates as name.

### Actual behavior

Only the issuer is shown:

TLS Certificate: CN=R3,O=Let's Encrypt,C=US

Name	CN=R3,O=Let's Encrypt,C=US
Valid	Yes
Activates	Fri, Mar 26, 2021 ...
Expires	Thu, Jun 24, 2021 ...
SHA-256 Fingerprint	...
MD5 Fingerprint	...

The issuer is a nice to have information, but the CN should be essential...
To find the CN I have to search my certificates and match the serialnumber...

### Steps to reproduce

Have not-self-signed-certificates and go to "TLS Certificates"

### GSA version

20.8.1

PR #3023, opened 2021-06-23
Fixes: TLS Certificates identified by issuer? #2847

Virsacer:
Show subject DN as name and put issuer DN into detail view

**Why**:

Because certificates can not reliably be identified by the issuer DN

**How**:

Compiled and listed the certificates in the webgui

**Checklist**:

- [ ] Tests

### DIFF
--- a/gsa/src/gmp/models/tlscertificate.js
+++ b/gsa/src/gmp/models/tlscertificate.js
@@ -51,8 +51,7 @@ class TlsCertificate extends Model {
       ? element.certificate.__text
       : undefined;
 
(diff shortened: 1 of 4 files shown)


Virsacer:
Before:
Before

After:
After

swaterkamp:
Thank you very much for your contribution, Virsacer! We are currently evaluating your changes and subsequent workflows.

timopollmeier:
Because this PR is based on an outdated branch, I've created a new PR with similar changes: #3820

PR #2660, opened 2021-01-13
Nvt detailspage test

y0urself:
**What**:

* Added tests for NVT Details Pages.
* No permissions as NVTs are Global Objects
* No Copy, Clone, Edit or Delete stuff for NVTs

While there:
* Updating the DateTime format for Note and Override Boxes

**Why**:

* Tests have been missing so far ....

**How**:

**Checklist**:

- [x] Tests
- [ ] [CHANGELOG](https://github.com/greenbone/gsa/blob/master/CHANGELOG.md) Entry

PR #3024, opened 2021-06-24
Add listpage and detailspage tests

saberlynx:
**What**:

Add tests from master branch and port all changes from https://github.com/greenbone/gsa/pull/2660

**Why**:

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/gsa/src/web/entity/__tests__/box.js
+++ b/gsa/src/web/entity/__tests__/box.js
@@ -17,7 +17,7 @@
  */
 import React from 'react';
 
-import date, {setLocale} from 'gmp/models/date';
+import Date, {setLocale} from 'gmp/models/date';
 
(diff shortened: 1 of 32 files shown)


gsa/src/web/entity/__tests__/box.js
```diff
@@ -17,22 +17,29 @@
  */
 import React from 'react';
 
-import Date, {setLocale} from 'gmp/models/date';
+import date, {setLocale} from 'gmp/models/date';
```
swaterkamp, line 20: This was upper case on purpose, because it's a component model, just like we have Result, Target etc in upper case as well. Putting this into lower case might give the impression, this is some kind of transformation function.

swaterkamp (approved):

PR #3124, opened 2021-08-06
[AP1032] Show dead hosts as possible reason in empty report when scan finishes

swaterkamp:
**What**:
Show possibility of dead hosts also for finished scans, not only `progress < 0`

**Why**:
The notification was not shown after scanning a dead host and was reported missing

**How**:
1. Create a target with a non-existing host.
2. Scan that target to produce an empty report.
3. Check that the notification is listed in the Results tab in the report details.

**Checklist**:

- [N/A] Tests
- [X] Labels for ports to other branches

### DIFF
--- a/gsa/src/web/pages/reports/details/emptyreport.js
+++ b/gsa/src/web/pages/reports/details/emptyreport.js
@@ -53,12 +53,28 @@ const EmptyReport = ({
       />
       <Divider wrap>
         {!isActiveReport && (
-          <ReportPanel
-            icon={props => <TaskIcon {...props} />}
-            title={_('The scan did not collect any results')}
-          >
-            {_('If the scan got interrupted you can try to re-start the task.')}
-          </ReportPanel>
+          <React.Fragment>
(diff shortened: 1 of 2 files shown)


gsa/src/web/pages/reports/details/emptyreport.js
```diff
@@ -78,7 +78,7 @@ const EmptyReport = ({
             {_('Just wait for results to arrive.')}
           </ReportPanel>
         )}
-        {progress < 1 && hasTarget && (
+        {!isActiveReport && hasTarget && (
```
sarahd93, line 81: I think we have to keep the "progress < 1" too. I would assume it covers the case where a scan is already running for a while but there isn't any progress yet because the scanned hosts are dead.

swaterkamp: I have added the dead host note to conditions `progress < 1` and `!isActiveReport`

sarahd93 (approved):

PR #3140, opened 2021-09-02
Revert version update of react-router-dom and history

swaterkamp:
**What**:
Revert react-router-dom dependency version from 5.2.1 to 5.2.0
Revert history dependency version from 5.0.1 to 4.10.1

**Why**:
react-router-dom breaks our login page presumably due to a change of default behavior (https://github.com/remix-run/react-router/issues/5362)
history breaks navigation almost completely

**How**:
Login and navigation is possible again

**Checklist**:

- [ ] Tests


guidobouman:
Great to hear your extensive thoughts. All the best!

PR #3160, opened 2021-09-17
[AP1571] Add license command and update page

swaterkamp:
**What**:
This is the next PR to the new license feature including updates to the license page, adding the command and some things more.
This is still a feature in development and deals with a dummy-license and ignores user permissions to access licenses.

**Why**:
New feature

**How**:

**Checklist**:

- [X] Tests


sarahd93 (changes requested):
Looks good. The tests are still failing because we forgot to set the timezone via the store. When we test a page with a date we always have to use setTimezone to ensure the test will always use the same timezone.

gsa/src/web/pages/page.js
```diff
@@ -46,16 +44,58 @@ import Footer from 'web/components/structure/footer';
...
 
+import InfoPanel from 'web/components/panel/infopanel';
+
+import PropTypes from 'web/utils/proptypes';
+import withGmp from 'web/utils/withGmp';
+import compose from 'web/utils/compose';
+
 const log = logger.getLogger('web.page');
 
+const LICENSE_EXPIRATION_THRESHOLD = 30; // TODO fix this
```
sarahd93, line 55:
```suggestion
const LICENSE_EXPIRATION_THRESHOLD = 30;
```

sarahd93: Have checked the requirements again and 30 days is actually the correct number.

sarahd93 (changes requested):

sarahd93 (approved):

PR #3164, opened 2021-09-21
Require node.js >= 14

swaterkamp:
**What**:
Use a newer node.js version as minimum requirement, because 10 is EOL.

**Why**:
Support for it should outlast GVM 21.10 and we have the first 3rd party dependencies requiring node.js >= 12

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -95,7 +95,8 @@ Prerequisites for using the GUI:
 
 Install node.js >= 14 on Debian GNU/Linux:
 ```sh
-curl -fsSL https://deb.nodesource.com/setup_14.x | bash -
+curl --silent --show-error https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
+echo "deb https://deb.nodesource.com/node_14.x buster main" > /etc/apt/sources.list.d/nodesource.list
 apt-get install -y nodejs
 ```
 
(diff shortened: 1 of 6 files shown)


bjoernricks (changes requested):
Could you check if
```sh
curl --silent --show-error https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -
echo "deb https://deb.nodesource.com/node_14.x buster main" | sudo tee /etc/apt/sources.list.d/nodesource.list
```
works instead of downloading and executing the https://deb.nodesource.com/setup_14.x shell script? Downloading and executing an arbitrary script from the internet without checking its integrity is very very dangerous.

P.S.: That similar to what we have used in the paste too (https://github.com/greenbone/gsa/blob/gsa-9.0/INSTALL.md#compiling-greenbone-security-assistant-gui)

INSTALL.md
```diff
@@ -90,9 +90,14 @@ Compiling Greenbone Security Assistant GUI
 ------------------------------------------
 
 Prerequisites for using the GUI:
-* node.js >= 8.0
+* node.js >= 14.0
 * Either yarn >= 1.0 or npm. yarn is faster and more reliable, but younger.
 
+Install node.js >= 14 on Debian GNU/Linux:
+```sh
+curl -fsSL https://deb.nodesource.com/setup_14.x | bash -
```
bjoernricks, line 98:
```suggestion
curl --silent --show-error https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
echo "deb https://deb.nodesource.com/node_14.x buster main" > /etc/apt/sources.list.d/nodesource.list
```

bjoernricks (approved):

PR #3176, opened 2021-10-04
Fix number of languages in tests

sarahd93:
**What**:
Adjust the number of languages in the language tests.

**Why**:
After adding the Traditional Chinese language option the tests were failing because they had not been updated.

**How**:

**Checklist**:

- [x] Tests


bjoernricks (approved):
I really don't understand why this test didn't failed in the PR for merging the change. Should not have been possible to merge the PR.

PR #3317, opened 2022-01-19
Adjust coverage tests

bjoernricks:
**What**:

Lower number of test workers to reduce memory consumption on our self hosted runners.

**Why**:

The memory consumption goes crazy otherwise.

**How**:

**Checklist**:

- [ ] Tests
- [ ] PR merge commit message adjusted

### DIFF
--- a/.github/workflows/ci-js.yml
+++ b/.github/workflows/ci-js.yml
@@ -11,9 +11,9 @@ jobs:
     env:
       REPORT_DIR: reports
     name: Testing and upload coverage
-    runs-on: 
+    runs-on:
       - self-hosted
-      - gsa_runner
+      - gsa-runner
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-js.yml
```diff
@@ -11,9 +11,9 @@ jobs:
     env:
       REPORT_DIR: reports
     name: Testing and upload coverage
-    runs-on: 
+    runs-on:
       - self-hosted
-      - gsa_runner
+      - gsa-runner
```
y0urself, line 16: Unfortunately the runners label is `gsa_runner` not `gsa-runner`.

bjoernricks: this repo runner will be removed. so i think we should remove the whole line

y0urself (approved):

PR #3323, opened 2022-01-27
remove self-hosted runner where not needed

Dexus:
**What**:

remove self-hosted runner

**Why**:

no internal task that need this runner

### DIFF
--- a/.github/workflows/ci-js.yml
+++ b/.github/workflows/ci-js.yml
@@ -11,9 +11,7 @@ jobs:
     env:
       REPORT_DIR: reports
     name: Testing and upload coverage
-    runs-on:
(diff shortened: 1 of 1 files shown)


Dexus:
as internal noted, this speeds up

bjoernricks (approved):
To improve the speed further we should try to increase the number of processes. We limited them in the past just because CircleCI didn't allow to run more then one process.

y0urself:
> To improve the speed further we should try to increase the number of processes. We limited them in the past just because CircleCI didn't allow to run more then one process.

Also the GitHub Runner only have 2 threads as a limiting factor. Our self-hosted runner run with at least 4 threads I think!

PR #3362, opened 2022-03-04
Move LicenseNotification into own module

swaterkamp:
**What**:
Extract logic and code behind the LicenseNotification out of the page.js into its own module

**Why**:
The code will become more complex and this gives a good basis to keep page.js as clean and simple as possible

**How**:
Manual tests were performed if the notification's behavior changed after moving the code.

**Checklist**:

- [x] Tests (currently not possible/senseful, because the test would need a reference to current server time, which obviously changes)
- [ ] PR merge commit message adjusted (Use: "Change: Move LicenseNotification into own module")

### DIFF
--- a/src/web/components/provider/licenseprovider.js
+++ b/src/web/components/provider/licenseprovider.js
@@ -27,24 +27,30 @@ const LicenseProvider = ({children}) => {
 
   const [license, setLicense] = useState({});
   const [licenseError, setLicenseError] = useState();
+  const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
     updateLicense();
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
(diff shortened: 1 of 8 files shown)


src/web/components/provider/licenseprovider.js
```diff
@@ -0,0 +1,49 @@
...
+const LicenseProvider = ({children}) => {
+  const gmp = useGmp();
+
+  const [license, setLicense] = useState({});
+
+  useEffect(() => {
+    updateLicense();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const updateLicense = () => {
```
sarahd93, line 34: We have to check how we can include error handling here.

sarahd93 (changes requested):

sarahd93 (approved):

PR #3523, opened 2022-10-13
Change: Update some dev dependencies

y0urself:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] PR merge commit message adjusted

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "storybook": "NODE_PATH=src start-storybook",
     "build-storybook": "NODE_PATH=src build-storybook",
     "i18n-extract": "NODE_ENV=development babel -f .babelrc 'src/**/*.{js,jsx,ts,tsx}'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -89,11 +90,11 @@
     "@types/jest": "^29.1.2",
     "babel-plugin-i18next-extract": "^0.9.0",
     "eslint-config-prettier": "^8.5.0",
-    "husky": "^2.7.0",
+    "husky": "^8.0.1",
```
swaterkamp, line 93: Updating husky across these versions will break linting, which is the reason, why this is still 2.7

swaterkamp (changes requested):

package.json
y0urself: Ok according to confluence the reason was:

> Because Debian oldstable and stable don’t provide a git package newer then 2.20.1 husky should only get updates for patch versions if there will be any.

PR #3634, opened 2023-02-20
Rebuild yarn.lock and add missing prop validations

timopollmeier:
## What

This replaces the yarn.lock file with one rebuilt from just the current `package.json` and adds some missing prop validation for alert conditions and events.

## Why

To make it simpler to synchronize it and the `package.json` in the stable and main branches.

### DIFF
--- a/src/web/pages/alerts/condition.js
+++ b/src/web/pages/alerts/condition.js
@@ -31,6 +31,8 @@ import {
   CONDITION_DIRECTION_INCREASED,
 } from 'gmp/models/alert';
 
+import PropTypes from 'web/utils/proptypes';
+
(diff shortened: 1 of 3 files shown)


src/web/pages/alerts/condition.js
```diff
@@ -98,6 +100,11 @@ const Condition = ({condition = {}, event}) => {
   return condition.type;
 };
 
+Condition.propTypes = {
+  condition: PropTypes.object,
```
swaterkamp, line 104: Condition seems to be optional, as it receives a default in the function header

swaterkamp (approved):

PR #3657, opened 2023-03-10
Update README.md

enderbee:
## What
Changed to Greenbone AG
Added SPDX Identifier

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,6 @@ github](https://github.com/greenbone/gsa/issues) first.
 
 ## License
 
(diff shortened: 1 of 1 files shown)


swaterkamp (changes requested):
I think it would be more productive to make one pull request for all changes of GmbH to AG. That includes for example the About page, the copyright footer component and all of the >1700 files with a copyright header. If we start to change single files we will lose the overview of what has been changed already and what is missing. In my opinion it would be best to have a script doing this, possibly across all our repos!

swaterkamp (approved):

PR #3687, opened 2023-04-19
Fix: Show correct auth method in user details

timopollmeier:
## What
When viewing the details of a user, the correct authentication method is now shown.

## Why
Previously only the table showed the correct method while the details always showed "Local".

## References
GEA-27


bjoernricks (approved):
Nice catch! I should have used camelCase from the beginning 😞

PR #3706, opened 2023-05-05
Fix: Get/save TippingPoint TLS workaround correctly

timopollmeier:
## What
The alert dialog now uses the correct field for TLS workaround option of the TippingPoint SMS method.

## Why
Previously the wrong field was used, so the dialog did not work correctly.

## References
GEA-181

## Checklist

- [ ] Tests


swaterkamp (approved):
Looks like it should work, but I have no possibility to check the actual function. I assume this was done during development.

PR #3707, opened 2023-05-05
Added a confirmation dialog when saving the own user data.

jhelmold:
## What
A confirmation dialog was added, when a user in the user dialog tries to save his own user data. This is only possible for Super Admins. So the user can't be logged out accidentally.

## Why
This is a bug-fix.

## References
GEA-37

## Checklist

- [ ] Tests


src/web/pages/users/dialog.js
```diff
@@ -48,45 +52,73 @@ import Divider from 'web/components/layout/divider';
...
+
   handleRoleIdsChange(value) {
     this.setState({
       noRoleConfirmed: false,
       roleIds: value,
     });
   }
 
-  handleSaveClick(onSave, d) {
+  handleSaveClick(user, onSave, d) {
```
swaterkamp, line 111: I know this has been here before, but now would be a good time to change it: no one-letter variables, please :)

src/web/pages/users/dialog.js
```diff
@@ -48,45 +52,73 @@ import Divider from 'web/components/layout/divider';
 import Layout from 'web/components/layout/layout';
 
 class Dialog extends React.Component {
+  dataToSave = false;
```
swaterkamp, line 55: Is this only the initialization of the variable? Please make sure, the type matches whatever the variable will receive later on. I'd suspect that `{}` would better fit. However, this could lead to the situation where you send an empty object, instead of the actual user data. Same is true for the `false` by the way.
I'm kind of missing where the real data is send here.

swaterkamp (changes requested):
It's currently unclear to me, how the correct user data is set and send. Other than that this looks like it works. Please consider my other review comments.

public/locales/gsa-de.json
```diff
@@ -1164,6 +1164,7 @@
...
   "Please note: You are about to change your license! This might have severe influence on the feature set available to you. Are you sure you want to continue?": "Bitte beachten Sie: Sie sind dabei Ihre Lizenz zu ändern! Dies kann einen starken Einfluss auf das verfügbare Feature-Set haben. Sind Sie sicher, dass Sie fortfahren wollen?",
+  "Please note: You are about to change your own personal user data as Super Admin! It is not possible to change the login name. If you have modified the login name, neither the login name nor any other changes made will be saved. If you have made any modifications other than the login name, the data will be saved when clicking OK, and you will be logged out immediately.": "Bitte beachten Sie: Sie sind dabei ihre eigenen persönlichen Nutzerdaten als Super Administrator zu ändern! Es ist nicht möglich, den Login Namen zu ändern. Wenn Sie den Login Namen modifiziert haben, werden weder der Login Name noch jedwede anderen Änderungen gespeichert. Wenn Sie andere Änderungen vorgenommen haben, werden die Daten beim Klick auf OK gespeichert und Sie werden umgehend ausgeloggt.",
```
swaterkamp, line 1167: Please fix typos :)

timopollmeier (approved):

PR #3792, opened 2023-06-30
Fix links to the manual and GMP documentation

bjoernricks:
## What

Fix links to the manual and GMP documentation

## Why
Despite the fact we are using 22.x.y as versions for our components now the GMP version will stay on 22.04. The URL to the manual requires the language now.


timopollmeier (approved):

cfi-gb:
Note that this could have introduced a regression:
- #3814 
- https://forum.greenbone.net/t/ce-container-help-url-broken/15273

PR #3793, opened 2023-06-30
Adjust GSA description at the about page

bjoernricks:
## What

Adjust GSA description at the about page

## Why
Try to improve the about page for using our current naming an terms.

## References

GEA-72


timopollmeier (commented):
Could you also update the German translation for the new text?

swaterkamp:
Please do not run the translation scripts. If you HAVE to add translations, you need to add them manually and correctly ordered to the json

bjoernricks:
I've added German translations for the changes of the about page.

timopollmeier (approved):

PR #3819, opened 2023-07-28
release-pontos: also create gsa-node-modules release artifact

Flowdalic:
For Gentoo, we ideally want to build the node modules as part of our "emerge" process. That means we do not want to consume the -dist artifact. What we instead want is the gsa source and the contents of the node_modules/ directory.

Therefore, add a second release artifact to release-ponto, that only contains the content of the node_modules/ directory.

Please consider this addition, as it would reduce the friction of packaging gsa in Gentoo.

Note that this is untested, since I do not know how to test github workflows, but the change seems trivial enough.

### DIFF
--- a/.github/workflows/release-pontos.yml
+++ b/.github/workflows/release-pontos.yml
@@ -81,12 +81,13 @@ jobs:
         run: yarn install --prefer-offline
       - name: Build dist files
         run: yarn build
-      - name: Create tarball
+      - name: Create tarballs
         run: |
           tar -C build -czvf gsa-dist-${{ needs.release.outputs.release-version }}.tar.gz .
-      - name: Upload dist file to release
+          XZ_OPT='-T0 -9' tar -acf gsa-node-modules-${{ needs.release.outputs.release-version }}.tar.xz node_modules
+      - name: Upload dist files to release
         run: |
-          gh release upload ${{ needs.release.outputs.git-release-tag }} gsa-dist-${{ needs.release.outputs.release-version }}.tar.gz
+          gh release upload ${{ needs.release.outputs.git-release-tag }} gsa-dist-${{ needs.release.outputs.release-version }}.tar.gz gsa-node-modules-${{ needs.release.outputs.release-version }}.tar.xz
         env:
           GH_TOKEN: ${{ secrets.GREENBONE_BOT_TOKEN }}
 


.github/workflows/release-pontos.yml
```diff
@@ -81,12 +81,14 @@ jobs:
...
-      - name: Create tarball
+      - name: Create tarballs
         run: |
           tar -C build -czvf gsa-dist-${{ needs.release.outputs.release-version }}.tar.gz .
-      - name: Upload dist file to release
+          XZ_OPT='-T0 -9' tar -acf gsa-node-modules-${{ needs.release.outputs.release-version }}.tar.xz node_modules
+      - name: Upload dist files to release
         run: |
           gh release upload ${{ needs.release.outputs.git-release-tag }} gsa-dist-${{ needs.release.outputs.release-version }}.tar.gz
+          gh release upload ${{ needs.release.outputs.git-release-tag }} gsa-node-modules-${{ needs.release.outputs.release-version }}.tar.xz
```
bjoernricks, line 91: The node modules should be uploaded additionally to the dist files.
```suggestion
          gh release upload ${{ needs.release.outputs.git-release-tag }} gsa-dist-${{ needs.release.outputs.release-version }}.tar.gz gsa-node-modules-${{ needs.release.outputs.release-version }}.tar.xz
```

bjoernricks (changes requested):
Would be nice if you can change your commit message title to `Add: Also create gsa-node-modules release artifact`. In that case it gets included into the release changelog automatically.

I am curious why you are not using the dist files. It is for reproducibility?

Flowdalic:
> I am curious why you are not using the dist files. It is for reproducibility?

Mostly because [Gentoo likes to perform the `yarn build` step itself](https://github.com/gentoo/gentoo/blob/ec7630170f1f7c47826d6c0738f705ca516cf237/net-analyzer/gsa/gsa-22.5.3.ebuild#L41-L42). Granted, it is not the same as building C/C++ source, but the next best equivalent. This allows users to fine-tune the resulting packages and allows testing with different yarn versions, etc.

bjoernricks:
Doesn't really make much sense for us but anyway I am happy to support Gentoo here.

bjoernricks (approved):

Flowdalic:
> Doesn't really make much sense for us but anyway I am happy to support Gentoo here.

Thanks, your support is appreciated. :)

PR #3831, opened 2023-08-07
Switch to npm

bjoernricks:
## What

Drop yarn in favor of npm

`yarn` --> `npm install`
`yarn build` --> `npm run build`
`yarn test` --> `npm run test`

## Why

Using yarn with newer versions of nodejs caused build errors. Therefore to allow using current maintained nodejs version 18 and 20 drop yarn in favor of npm.

When we did choose yarn at the beginning of the GSA JavaScript version, yarn was much faster and more reliable. But in the meanwhile npm got on par to yarn's features and speed. Additionally the yarn project changed it's direction and re-implemented yarn which resulted in the 2.x and later versions and abandoned 1.x which we use(d).

Therefore drop yarn and switch to npm which allows to use maintained software again.


y0urself (commented):
LGTM ...

timopollmeier (approved):

PR #3857, opened 2023-09-05
Change: request edit_config_family_all alongside edit_config_family

mattmundell:
## What

When getting scan config data, run the new gsad cmd `edit_config_family_all` concurrently with the existing `edit_config_family` cmd.

## Why

Each of these cmds calls a separate GMP GET_NVTS command, so they can run simultaneously to save time.  This speeds up the config family edit dialog.

## References

Requires greenbone/gsad/pull/146.


mattmundell:
I'm guessing a bit with the test commit, but it seems to be correct.

gsad side is ready to merge when this PR is accepted.

timopollmeier (approved):

PR #3867, opened 2023-09-15
Fix: Add CVSS missing range (0.1-0.9)

a-h-abdelsalam:
## What

The range (0.1-0.9) is now shown in the CVSS chart and the corresponding tooltip and filter are consistent with the range.

Fixed the filter ranges for severity values 9 and 10. Value 9 should exclude 10 and Value 10 filter should filter only for 10 and not for > 10.

## Why

The CVSS range (0.1-0.9) was missing.

## References

GEA-266

### DIFF
--- a/src/web/components/dashboard/display/cvss/cvssdisplay.js
+++ b/src/web/components/dashboard/display/cvss/cvssdisplay.js
@@ -52,7 +52,7 @@ class CvssDisplay extends React.Component {
 
     let statusFilter;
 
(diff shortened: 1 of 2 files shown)


bjoernricks (approved):

a-h-abdelsalam:
Hey @bjoernricks thanks for taking a look on this PR. We might need to wait on merging this one. I disabled auto-merge

timopollmeier (approved):

PR #3877, opened 2023-09-26
Change: Update d3 and jest-dom versions

a-h-abdelsalam:
## What

Update the versions of d3 libraries and jest-dom.
- import of jest-dom had to be updated since extend-expect is removed starting from jest-dom 6.0.0. See [Releases](https://github.com/testing-library/jest-dom/releases)
- d3 node modules need to be transformed so that jest can parse them. See [3.0.1 doesn't work with Jest](https://github.com/d3/d3-interpolate/issues/96)

## Why

To keep the dependencies updated to the latest versions.

## References

GEA-309

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,14 +18,14 @@
         "@vx/shape": "^0.0.199",
         "babel-loader": "9.1.3",
         "core-js": "^3.32.2",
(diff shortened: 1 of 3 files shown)


bjoernricks (approved):

bjoernricks:
Nice!

PR #3890, opened 2023-10-05
Enable Dependabot update grouping

timopollmeier:
## What
Grouping of all Dependabot updates that only increase the patch level is enabled so they will be added in a single commit.

## Why
To simplify the update process.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,12 @@ updates:
     schedule:
       interval: monthly
       time: '04:00'
+    groups:
+      patch-updates:
+        patterns:
+          - "*"
+        update-types:
+          - "patch"
     open-pull-requests-limit: 10
     versioning-strategy: increase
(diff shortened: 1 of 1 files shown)


bjoernricks:
I am not sure if grouping all npm dependencies is a good idea. It seems to me that npm dependency updates are break to often and therefore a single grouped PR would not work.

bjoernricks:
Btw. are the ignore settings for npm updates still valid?

a-h-abdelsalam:
What about grouping at least the patch updates together `update-types: patch`? There is a higher chance they are not breaking changes.

bjoernricks:
yes that's a good solution!

timopollmeier:
Okay, I'll update the commit accordingly.

bjoernricks (approved):

bjoernricks:
Oh please add grouping for github-actions too

PR #3894, opened 2023-10-08
i18n lose zh_TW

qpowjohn:
I'm Traditional Chinese user, when i change interface language to chinese but not work. i search container and source code. i think source code lose something.

## Checklist

- [ ] Tests


bjoernricks:
Hi @qpowjohn, thanks a lot for your contribution!

I've tested your changes but sadly wasn't enough to get the translations to work again. Therefore I did spend some time on the issue and created #3895 which replaces your PR.

PR #3900, opened 2023-10-16
Change: Add missing apt-get install line to README.md

cfi-gb:
## What
If the nodejs package isn't installed it can't be used later.

Note: https://deb.nodesource.com/ is using different install instructions, we might want to adjust it completely to that:

```
~$ sudo apt-get update && sudo apt-get install -y ca-certificates curl gnupg
~$ curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
~$ NODE_MAJOR=20
~$ echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
~$ sudo apt-get update && sudo apt-get install nodejs -y
```

## Why
Better documentation

## References

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ gpg --no-default-keyring --keyring "$KEYRING" --list-keys
 
 echo "deb [signed-by=$KEYRING] https://deb.nodesource.com/$VERSION $DISTRIBUTION main" | sudo tee /etc/apt/sources.list.d/nodesource.list
 echo "deb-src [signed-by=$KEYRING] https://deb.nodesource.com/$VERSION $DISTRIBUTION main" | sudo tee -a /etc/apt/sources.list.d/nodesource.list
+
+sudo apt-get update && sudo apt-get install nodejs
 ```
 
 Change into the gsa source directory and delete the possible existing build output
--- a/README.md
+++ b/README.md
@@ -56,17 +56,16 @@ Prerequisites for GSA:
 To install nodejs the following commands can be used
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,8 @@ gpg --no-default-keyring --keyring "$KEYRING" --list-keys
 
 echo "deb [signed-by=$KEYRING] https://deb.nodesource.com/$VERSION $DISTRIBUTION main" | sudo tee /etc/apt/sources.list.d/nodesource.list
 echo "deb-src [signed-by=$KEYRING] https://deb.nodesource.com/$VERSION $DISTRIBUTION main" | sudo tee -a /etc/apt/sources.list.d/nodesource.list
+
+sudo apt-get update && sudo apt-get install nodejs
```
bjoernricks, line 69:
```suggestion
sudo apt update && sudo apt install nodejs
```

cfi-gb: I thought about that as well but then used the strings from `https://deb.nodesource.com/` as i wasn't sure if there is a reason why they used `apt-get` vs. just `apt`.

bjoernricks (approved):

README.md
bjoernricks: Not sure why they still use apt-get. I tend to replace it with just apt since some years.

PR #3902, opened 2023-10-17
Change: Handle delta results with changed severity, qod and hostname

a-h-abdelsalam:
## What
Handle delta results with changed severity, qod and hostname. In addition to description differences, differences in severity, qod or hostname are considered as a change in delta results comparison.

## Why
Within the new delta reports handling, results that have changes only in severity, qod or hostname are now considered changes instead of gone/new.

## References
GEA-320

### DIFF
--- a/src/web/components/icon/deltadifferenceicon.js
+++ b/src/web/components/icon/deltadifferenceicon.js
@@ -1,4 +1,4 @@
-/* Copyright (C) 2018-2022 Greenbone AG
+/* Copyright (C) 2023 Greenbone AG
  *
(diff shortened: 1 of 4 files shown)


src/web/components/icon/deltadifferenceicon.js
```diff
@@ -0,0 +1,26 @@
+/* Copyright (C) 2018-2022 Greenbone AG
```
bjoernricks, line 1: If that is a new file it should be
```suggestion
/* Copyright (C) 2023 Greenbone AG
```

jhelmold (approved):

PR #3903, opened 2023-10-20
Remove: Removed the logic for the client certificates credential type.

jhelmold:
## What
The client certificates credential type logic has been removed.

The credentials of type "client certificate" are not touched and can still be accessed. But they
are of no use anymore. They can be deleted manually.

## Why
The client certificates credential type was once used for (custom) OSP scanners,
a functionality that was removed with GOS 22.04.0. So it is not used any more and
it confused the users.

## References
GEA-333

## Checklist
Manually tested on my development environment.

- [x] Tests


src/web/pages/credentials/__tests__/dialog.js
```diff
@@ -123,7 +123,6 @@ describe('CredentialsDialog component tests', () => {
     );
 
     const formGroups = getAllByTestId('formgroup-title');
-    const selectedValue = getByTestId('select-selected-value');
```
timopollmeier, line 126: I think you should add this part of the test again...

src/web/pages/credentials/__tests__/dialog.js
```diff
@@ -134,7 +133,6 @@ describe('CredentialsDialog component tests', () => {
     expect(commentInput).toHaveAttribute('value', 'blah');
 
     expect(formGroups[2]).toHaveTextContent('Type');
-    expect(selectedValue).toHaveTextContent('Client Certificate');
```
timopollmeier, line 137: ... and also add this again with the changed credential type.

src/web/pages/credentials/__tests__/listpage.js
```diff
@@ -166,7 +166,6 @@ describe('CredentialPage tests', () => {
 
     expect(row[1]).toHaveTextContent('credential 1');
     expect(row[1]).toHaveTextContent('(blah)');
-    expect(row[1]).toHaveTextContent('Client Certificate(cc)');
```
timopollmeier, line 169: This should also be added to the test again with the changed credential type.

src/web/pages/scanners/dialog.js
```diff
@@ -125,10 +119,7 @@ const ScannerDialog = ({
...
     if (onScannerTypeChange) {
       value = parseInt(value);
-      const scan_credentials = filter(
-        credentials,
-        client_cert_credentials_filter,
-      );
+      const scan_credentials = filter(credentials);
 
       onScannerTypeChange(value, name);
       onScannerTypeChange(
```
timopollmeier, line 125: Is this and the declaration of `scan_credentials` still needed or can it be removed?

src/web/pages/scanners/dialog.js
```diff
@@ -167,15 +158,9 @@ const ScannerDialog = ({
...
-    credentials,
-    client_cert_credentials_filter,
-  );
+  const scanner_credentials = filter(credentials);
   const isInUse = isDefined(scanner) && scanner.isInUse();
-  const show_cred_info =
-    isDefined(scanner) &&
-    isDefined(scanner.credential) &&
-    scanner.credential.credential_type === CLIENT_CERTIFICATE_CREDENTIAL_TYPE;
+  const show_cred_info = isDefined(scanner) && isDefined(scanner.credential);
```
timopollmeier, line 163: I think this can be removed as well as the `scanner_credentials` declaration and the conditional `<CertStatus>` element below.
If the ` CertStatus` component is used nowhere else, it can also be removed.

timopollmeier (changes requested):
I think there are some parts of the code made obsolete by the removal that can still be cleaned up while in the tests there are still parts that should not have been removed, but adjusted to the changed credential type instead.

timopollmeier (approved):

PR #3915, opened 2023-11-10
Fix: Fix invalid DOM props and other minor JS console errors

timopollmeier:
## What
This fixes various minor errors that do not visibly affect the content visible to users but appear in the JS console.

This includes:
- internal props are no longer passed to the DOM
- invalid prop types are fixed
- better handling of models and missing data
- tr elements are nested in the correct parent elements
- keys in reference lists are made unique
- correct functions are defined for gmp mock in tests

## Why
Fixing these errors makes it easier to spot errors that do have an effect on the visible content.

## References
GEA-350

## Checklist
- [x] Tests


bjoernricks (commented):
Really nice! Something I would loved to be fixed years ago already.

Maybe make it even simpler to exclude props by adding something like
```js
const styledExcludeProps = (styledComponent, excludedProps) => styledComponent.withConfig(
  excludePropsConfig(excludedProps)
);

usage
```js
styledExcludeProps(styled.div`
....
`, ["foo", "bar"])
```

bjoernricks (approved):

PR #3916, opened 2023-11-13
Change: Show status information for container tasks and reports

a-h-abdelsalam:
## What
Container tasks now show "Processing" when uploading reports or creating assets and show "Interrupted" when upload is interrupted. Reports show "Uploading" during upload  and "Interrupted" if upload is interrupted.

## Why
If a report is uploaded to a container task it will appear done (with the "Container" status in GSA) once it is in the phase of processing the assets. This can be misleading, especially if the processing takes a long time, as is the case for uploading multiple larger reports.

## References
GEA-353

### DIFF
--- a/src/gmp/models/task.js
+++ b/src/gmp/models/task.js
@@ -64,6 +64,7 @@ export const TASK_STATUS = {
   interrupted: 'Interrupted',
   container: 'Container',
   uploading: 'Uploading',
+  uploadinginterrupted: 'Uploading Interrupted',
   processing: 'Processing',
   done: 'Done',
 };
@@ -84,6 +85,7 @@ const TASK_STATUS_TRANSLATIONS = {
   Done: _l('Done'),
   Queued: _l('Queued'),
   Processing: _l('Processing'),
+  'Uploading Interrupted': _l('Interrupted'),
 };
 /* eslint-disable quote-props */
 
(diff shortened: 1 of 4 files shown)


src/gmp/models/task.js
```diff
@@ -84,6 +85,7 @@ const TASK_STATUS_TRANSLATIONS = {
   Done: _l('Done'),
   Queued: _l('Queued'),
   Processing: _l('Processing'),
+  'Uploading Interrupted': _l('Interrupted'),
```
bjoernricks, line 88: Is that correct? Shouldn't it be
```suggestion
  'Uploading Interrupted': _l('Uploading Interrupted'),
```

src/web/pages/reports/row.js
```diff
@@ -106,7 +106,9 @@ const Row = ({
   if (isDefined(task)) {
     if (task.isContainer()) {
       status =
-        status === TASK_STATUS.running
+        status === TASK_STATUS.interrupted
+          ? TASK_STATUS.uploadinginterrupted
+          : status === TASK_STATUS.running || status === TASK_STATUS.processing
           ? TASK_STATUS.uploading
           : TASK_STATUS.container;
```
timopollmeier, line 113:
```suggestion
            ? TASK_STATUS.uploading
            : TASK_STATUS.container;
```
Could you add a level of indentation here to make it more obvious which ternary operators belong together?

timopollmeier (approved):

Issue #3914, opened 2023-11-06
[22.8.1] - twitching hover experience on scan report navbar


tgurr:
### Expected behavior
Smooth and non twitching like it was before.

### Actual behavior
![output](https://github.com/greenbone/gsa/assets/414984/b73ce40b-4a8b-48ad-8f60-7dc3b5519c6a)

### Steps to reproduce
1. upgrade to 22.8.1
2. open a scan report
3. hover the navigation bar

### GVM versions

**gsa:** 22.07.0

**gvm:** 23.0.1

**openvas-scanner:** 22.7.6

**gvm-libs:** 22.7.3

### Environment

**Operating system:** Linux

**Installation method / source:** packages (source based distribution)

cfi-gb:
Seems this is handled via #3920 now

PR #3919, opened 2023-11-22
Change: Revert styled-components version to 5.3.11

a-h-abdelsalam:
## What
Revert styled-components version from 6.1.0 to 5.3.11. 
Undo manually enabling vendor prefixes since this was only needed from styled-components v6.0.0 onward.
Test snapshots are updated due to version change.

## Why

Hovering on the tabs of an open report causes jumping behavior after updating styled-components from v4.4.1 to v6.1.0. Same happens with the tabs of alerts.

## References

GEA-382

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "gsa",
-  "version": "22.8.2-dev1",
+  "version": "22.9.1-dev1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "gsa",
-      "version": "22.8.2-dev1",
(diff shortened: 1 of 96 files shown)


bjoernricks:
Hi,

did you find the cause of the issue? Reverting to an old styled-components version can only be a temporary solution.

bjoernricks:
> Related to #3914 ?

As far as I understand this PR should fix this issue.

a-h-abdelsalam:
Found the root cause of the issue and created a new [PR](https://github.com/greenbone/gsa/pull/3920). Closing this one.

PR #3920, opened 2023-11-24
Fix: Fix jumping behavior when hovering over tabs of open report/alert.

a-h-abdelsalam:
## What
Prefix pseudo classes with &.

## Why

Pseudo selectors no longer work when not prefixed with & in styled-components 6 which causes jumping behavior when hovering over tabs in open reports or alerts. See https://github.com/styled-components/styled-components/issues/4177

## References
GEA-382
Closes: https://github.com/greenbone/gsa/issues/3914

### DIFF
--- a/src/web/components/tab/tab.js
+++ b/src/web/components/tab/tab.js
@@ -47,13 +47,13 @@ const StyledDiv = styledExcludeProps(styled.div, ['active'])`
   margin-bottom: ${props => (props.active ? '-2px' : undefined)};
   border-top: ${props =>
     props.active ? '2px solid ' + Theme.green : '2px solid ' + Theme.white};
-  :hover {
+  &:hover {
     border-top: ${props =>
       props.active
-        ? '2px solid ' + Theme.white
+        ? '2px solid ' + Theme.green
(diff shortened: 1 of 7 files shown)


cfi-gb:
You could also add something like e.g. the following to the initial issue description and the issue would be then automatically closed once this PR is merged:

Closes: #3914

src/web/components/tab/tab.js
```diff
@@ -47,13 +47,13 @@ const StyledDiv = styledExcludeProps(styled.div, ['active'])`
   margin-bottom: ${props => (props.active ? '-2px' : undefined)};
   border-top: ${props =>
     props.active ? '2px solid ' + Theme.green : '2px solid ' + Theme.white};
-  :hover {
+  &:hover {
     border-top: ${props =>
       props.active
-        ? '2px solid ' + Theme.white
+        ? '2px solid ' + Theme.green
```
bjoernricks, line 53: Is this change intended?

a-h-abdelsalam: Yes in order to keep the top border green when a tab is highlighted.

timopollmeier (approved):

PR #3930, opened 2023-12-12
Fix: Use correct oldest result date in vulns

timopollmeier:
## What
When parsing the result dates for vulnerabilities, the correct field is now used for the oldest result.

## Why
When date parsing was added to the Vulnerability model class, the newest date was used accidentally.

## References
GEA-414

## Checklist

- [ ] Tests


bjoernricks (approved):
Would be nice to get a test for that

PR #3943, opened 2024-01-04
Change: fetch trash resources concurrently

mattmundell:
## What

Run the commands that get trash resources concurrently.

## Why

This is faster because it causes gsad to run the GMP requests concurrently.

For example with a trashcan of 1000 alerts, configs and targets:
```
main: 60s
this PR: 52s
```
The time is down to the slowest request (getting the configs).

## References

Requires greenbone/gsad/pull/162

Similar PR for configs: greenbone/gsa/pull/3857


a-h-abdelsalam (commented):
Looks good! You may want to resolve the conflicts to include report_configs.

mattmundell:
> Looks good! You may want to resolve the conflicts to include report_configs.

Thanks, resolved.

a-h-abdelsalam (approved):

PR #3973, opened 2024-03-19
Change: Drop ui-components as a dependency for GSA

a-h-abdelsalam:
## What

Drop ui-components as a dependency and instead use existing GSA form components for the login.

## Why

The ui-components are only used by GSA for the button and input components in the login form. It makes sense to move those components to GSA instead of having to maintain a separate repository.

## References
GEA-393


bjoernricks (commented):
looks good to me

bjoernricks (approved):

PR #3975, opened 2024-03-27
Change: Replace vx with visx

bjoernricks:
## What

Replace vx with visx

## Why

Use the airbnb vixs library instead of the unmaintained vx. visx seems to be an one to one replacement of vx. vx is stick to react 16 and this keeps us from updating to react 18 too.


a-h-abdelsalam:
Nice! Looks good to me

a-h-abdelsalam (approved):

PR #3980, opened 2024-03-28
Fix "Error: Not implemented: navigation (except hash changes)" during tests

bjoernricks:
## What

Fix "Error: Not implemented: navigation (except hash changes)" during tests

## Why

JSDOM doesn't implement the navigation API and therefore raises an error in the console if `window.location` is used and anchors are clicked. As we are using a lot of anchors in our pages this error is raised often during test runs. By mocking the HTMLAnchorElement click handler the console error is silenced.


jhelmold (approved):
Looks good to me.

PR #3997, opened 2024-04-10
Add: New multi_selection report format param type

timopollmeier:
## What
A new parameter type for report formats and report configs is added that allows selecting multiple predefined options.

Some small style changes have been made to make params with multi-line values more readable and to make the params in report formats and report configs more consistent.

## Why
The new param type can later be used to add new customization options to report formats like a column selection for CSV.

## References
GEA-403

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/reportconfigs/dialog.jsx
+++ b/src/web/pages/reportconfigs/dialog.jsx
@@ -68,6 +68,8 @@ const Param = ({
     field = (
       <YesNoRadio
         convert={parseBoolean}
+        yesValue={true}
+        noValue={false}
         name={name}
         value={field_value}
         onChange={onPrefChange}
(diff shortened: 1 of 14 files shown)


src/web/pages/reportconfigs/dialog.jsx
```diff
@@ -68,6 +68,8 @@ const Param = ({
     field = (
       <YesNoRadio
         convert={parseBoolean}
+        yesValue={true}
+        noValue={false}
```
a-h-abdelsalam, line 72: This does not seem to work for me as expected. The default value is not selected when opening the dialog for the first time and trying to set a value does not seem to work (No is always selected). Setting `yesValue={1} noValue={0}` and using the default convert seems to work for me. But re-opening the dialog after saving the values did not show the saved value.

a-h-abdelsalam (changes requested):

src/web/pages/reportconfigs/dialog.jsx
timopollmeier: a32a6b0 should fix this

a-h-abdelsalam (approved):

PR #4030, opened 2024-05-16
Add: Added report config options to alerts.

jhelmold:
## What

Report config options have been added to the methods: Email, SCP, Send to host, SMB and verinice.PRO Connector.
The Alemba vFire method is even more complex and will be handled separately.

## Why

Alert methods that use a report format should also have an option to choose an optional report config. 

## References
GEA-474

## Checklist

- [ ] Tests


src/web/pages/alerts/component.jsx
```diff
@@ -681,6 +720,7 @@ class AlertComponent extends React.Component {
 
           const result_filter_id = selectSaveId(result_filters);
           const report_format_id = selectSaveId(report_formats);
+          const report_config_id = selectSaveId(report_configs);
```
a-h-abdelsalam, line 723: Perhaps `UNSET_VALUE` could be passed here so that we don't get the first report config in the list. 

I noticed when running locally that if I open `New Alert` dialog and change the `Method` to `SCP` for example, I see `Report Config` displaying the first value in the list of the report configs even though it does not match the selected `Report Format`.

src/web/pages/alerts/sendmethodpart.jsx
```diff
@@ -32,12 +33,37 @@ import TextField from 'web/components/form/textfield';
 
 const SendMethodPart = ({
   prefix,
+  reportConfigs,
   reportFormats,
   sendHost,
   sendPort,
+  sendReportConfig,
   sendReportFormat,
+  onSave,
```
a-h-abdelsalam, line 42: I think this one was missed.
```suggestion
```

src/web/pages/alerts/sendmethodpart.jsx
```diff
@@ -73,11 +107,14 @@ const SendMethodPart = ({
...
 SendMethodPart.propTypes = {
   prefix: PropTypes.string,
+  reportConfigs: PropTypes.array,
   reportFormats: PropTypes.array,
   sendHost: PropTypes.string.isRequired,
   sendPort: PropTypes.string.isRequired,
+  sendReportConfig: PropTypes.id,
   sendReportFormat: PropTypes.id,
   onChange: PropTypes.func,
+  onSave: PropTypes.func,
```
a-h-abdelsalam, line 117:
```suggestion
```

PR #4035, opened 2024-05-23
fix powerfilter input parsing

daniele-mng:
## Why

It returns an error t.lowerCase is not a function

## References

GEA-448

## Checklist

- [ ] Tests


src/gmp/utils/identity.js
```diff
@@ -15,6 +15,9 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
+
+import {parseFloat} from 'gmp/parser';
```
bjoernricks, line 19: Hi, could we avoid this import because it created a cyclic dependency?

src/gmp/models/filter/convert.js
```diff
@@ -87,6 +87,10 @@ const convert = (keyword, value, relation) => {
     return {value, relation};
   }
 
+  if (isNumberOrNumberString(keyword)) {
+    return {value: `${keyword}${relation}${value}`, relation: '~'};
```
a-h-abdelsalam, line 91: After checking how the filter is split in gvmd, adding quotes here will ensure gvmd does not interpret the keyword as a column name
```suggestion
    return {value: `"${keyword}${relation}${value}"`, relation: '~'};
```

src/gmp/models/filter/__tests__/convert.js
```diff
@@ -248,6 +248,20 @@ describe('convert tests', () => {
...
+  test('should return correct value and relation when keyword is a number or number string', () => {
+    expect(convert('123', '456', '=')).toEqual({
+      value: '123=456',
+      relation: '~',
+    });
+  });
+
+  test('should return correct value and relation when keyword is a number', () => {
+    expect(convert(123, 456, ':')).toEqual({
+      value: '123:456',
```
a-h-abdelsalam, line 261:
```suggestion
      value: '"123:456"',
```

src/gmp/models/filter/__tests__/convert.js
```diff
@@ -248,6 +248,20 @@ describe('convert tests', () => {
       value: 'regexp',
     });
   });
+
+  test('should return correct value and relation when keyword is a number or number string', () => {
+    expect(convert('123', '456', '=')).toEqual({
+      value: '123=456',
```
a-h-abdelsalam, line 254:
```suggestion
      value: '"123=456"',
```

src/gmp/utils/identity.js
daniele-mng: Yes, you are right. Thanks for pointing it out.

src/gmp/models/filter/convert.js
daniele-mng: Thanks for the changes, I've applied them.

a-h-abdelsalam (approved):

PR #4052, opened 2024-06-06
Add a react hook for storing instance variables

bjoernricks:
## What

Add a react hooks for storing instance variables

## Why
An instance variable stores the value directly and doesn't cause re-renders if it is changed. Variables returned from this hooks are comparable to instance variables for class components.

Required for HOC less entities pages.

## Checklist

- [x] Tests


daniele-mng (approved):
LGTM!

PR #4054, opened 2024-06-06
Add a useShallowEqualSelector hook

bjoernricks:
## What

Add a useShallowEqualSelector hook

## Why

The useShallowEqualSelector hooks allows to avoid re-renders if an object is selected from the redux store but its value(s) didn't change. With the standard selector which uses `===` comparison even updating an object's value to the same value will cause a re-render (because a new state object is created).

This pattern can be found at https://react-redux.js.org/api/hooks#recipe-useshallowequalselector

## Checklist

- [x] Tests


daniele-mng (approved):
LGTM

PR #4055, opened 2024-06-06
Use page filter hook

bjoernricks:
## What

Implement a usePageFilter hook for getting the current applied filter of a page and also to change, reset and remove this filter.

Some background:

Every (entities) page can have a filter for querying the data. The filter can be specified by
1. the user via the powerfilter or some button that modifies the current filter
1. the filter query argument of the current page via the browser URL (`http://...?filter=...`)
1. the last used filter when returning to the page which is stored in the redux store (aka. page filter)
1. the default filter of the entity (which is stored in the database at the user settings and can be received via a specific UUID)
1. the default filter settings for all pages (for example the rows per page setting)

The precedence of the applied filter is specified in the listed order

The new usePageFilter hook implements this behavior as a hook. It can now replace the `FilterProvider` component that implements a similar behavior but via render props. Internally `FilterProvider` is refactored to use `usePageFilter` now.

## Why

The usePageFilter hook is a necessary step for a possible implementation of an entities pages as a function component with hooks.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/hooks/usePageFilter.js
+++ b/src/web/hooks/usePageFilter.js
@@ -53,23 +53,40 @@ const useDefaultFilter = pageName =>
  */
 const usePageFilter = (
   pageName,
-  {fallbackFilter, locationQueryFilterString} = {},
+  {
+    fallbackFilter,
+    locationQueryFilterString: initialLocationQueryFilterString,
+  } = {},
 ) => {
   const gmp = useGmp();
   const dispatch = useDispatch();
   const location = useLocation();
   const history = useHistory();
-
-  // only use location directly if locationQueryFilterString is undefined
-  // use null as value for not set at all
-  if (!isDefined(locationQueryFilterString)) {
-    locationQueryFilterString = location?.query?.filter;
-  }
-
-  let returnedFilter;
-
   const [defaultSettingFilter, defaultSettingsFilterError] =
     useDefaultFilter(pageName);
+  let [rowsPerPage, rowsPerPageError] = useShallowEqualSelector(state => {
+    const userSettingDefaultSel = getUserSettingsDefaults(state);
+    return [
+      userSettingDefaultSel.getValueByName('rowsperpage'),
+      userSettingDefaultSel.getError(),
+    ];
+  });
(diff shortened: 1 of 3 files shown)


src/web/hooks/usePageFilter.js
```diff
@@ -0,0 +1,184 @@
...
+  if (!isDefined(locationQueryFilterString)) {
+    locationQueryFilterString = location?.query?.filter;
+  }
+
+  let returnedFilter;
+
+  const [defaultSettingFilter, defaultSettingsFilterError] =
+    useDefaultFilter(pageName);
+
+  useEffect(() => {
```
daniele-mng, line 74: Nitpick: Could we reorder and group hooks and `consts`, so we have all the `useState` ,  `variables` ,`useEffects` ? Would be easier to follow.

src/web/hooks/usePageFilter.js
```diff
@@ -0,0 +1,184 @@
...
+) => {
+  const gmp = useGmp();
+  const dispatch = useDispatch();
+  const location = useLocation();
+  const history = useHistory();
+
+  // only use location directly if locationQueryFilterString is undefined
+  // use null as value for not set at all
+  if (!isDefined(locationQueryFilterString)) {
+    locationQueryFilterString = location?.query?.filter;
```
daniele-mng, line 66: Question: Would it be possible to avoid mutating the directly `locationQueryFilterString`?
Something like this maybe.
`initialLocationQueryFilterString` passed as `arg` to the hook
```jsx
  let locationQueryFilterString = isDefined(initialLocationQueryFilterString)
    ? initialLocationQueryFilterString
    : location?.query?.filter;
```

src/web/hooks/usePageFilter.js
bjoernricks: I put the variable and hook definitions to the top.

daniele-mng (approved):

src/web/hooks/usePageFilter.js
daniele-mng: Thanks!

PR #4056, opened 2024-06-07
Add a usePreviousValue hook

bjoernricks:
## What

Add a usePreviousValue hook

## Why

Add a new hooks that allows to get the previous value after the value is changed. For example this hook can be used to check whether a filter has changed.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/hooks/usePreviousValue.js
+++ b/src/web/hooks/usePreviousValue.js
@@ -1,0 +1,35 @@
+/* SPDX-FileCopyrightText: 2024 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import {useEffect, useRef} from 'react';
+
+/**
+ * React custom hook to store the value of the previous rendering
+ *
+ * It allows for checking if a value has changed since last render.
+ *
+ * Code Snippet
+ *
+ * ```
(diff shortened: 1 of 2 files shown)


src/web/hooks/usePreviousValue.js
```diff
@@ -0,0 +1,35 @@
...
+ * const [value, setValue] = useState(initialValue);
+ * const previousValue = usePrevious(value);
+ *
+ * if (value !== previousValue) { // value has changed
+ *     doSomething()
+ * }
+ * ```
+ */
+const usePreviousValue = value => {
+  const ref = useRef(value); // set value in initial rendering
```
daniele-mng, line 25: Question:  Maybe `useRef` should be initialized with a `previous` value like `undefined`, not sure if having the `initialValue` and `previousValue` the same at first render could lead to bugs.
```suggestion
  const ref = useRef(); // set value in initial rendering
```

bjoernricks: yes you are right. didn't think about that but that in detail but if there wasn't a previous value because of the initial rendering it should return undefined.

bjoernricks: Done

daniele-mng (approved):
LGTM :rocket:

PR #4057, opened 2024-06-07
Implement a useTiming hook

bjoernricks:
## What

Implement a useTiming hook

## Why

The hook can be used to run a function after a specific amount of time for example for doing a reload of data.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/hooks/__tests__/useTiming.jsx
+++ b/src/web/hooks/__tests__/useTiming.jsx
@@ -1,0 +1,141 @@
+/* SPDX-FileCopyrightText: 2024 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+/* eslint-disable react/prop-types */
+
+import {useState} from 'react';
+
+import {describe, test, expect, testing} from '@gsa/testing';
+
+import {act, fireEvent, render, screen} from 'web/utils/testing';
+
+import useTiming from '../useTiming';
+
+const TestComponent = ({doFunc}) => {
+  const [value, setValue] = useState(0);
+  const timingFunc = () => {
+    setValue(value => value + 1);
+    return doFunc();
+  };
+  const [startTimer, clearTimer, isRunning] = useTiming(timingFunc, 900);
+  return (
+    <>
+      <button onClick={startTimer} data-testid="startTimer"></button>
+      <button onClick={clearTimer} data-testid="clearTimer"></button>
+      <span data-testid="value">{value}</span>
+      <span data-testid="isRunning">{'' + isRunning}</span>
+    </>
+  );
+};
+
+const runTimers = async () => {
+  await act(async () => {
+    await testing.advanceTimersToNextTimerAsync();
+  });
+};
+
+describe('useTiming', () => {
+  test('should start a timer', async () => {
+    testing.useFakeTimers();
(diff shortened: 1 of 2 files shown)


src/web/hooks/useTiming.js
```diff
@@ -0,0 +1,109 @@
...
+ * Hook to start a timer that calls a function after a given timeout
+ *
+ * @param {Function} doFunc The function to call
+ * @param {Number|Function} timeout The timeout in milliseconds or a function that returns the timeout
+ * @returns Array of startTimer function, clearTimer function and boolean isRunning
+ */
+const useTiming = (doFunc, timeout) => {
+  const timer = useInstanceVariable({});
+  const [timerId, setTimerId] = useState(); // store timerId in state too to trigger re-render if it changes
+  const isRunning = !!timerId;
```
daniele-mng, line 26: Nitpick: I would avoid `!!`
```suggestion
  const isRunning = Boolean(timerId);
```

bjoernricks: Any reason for avoiding `!!`? GSA uses `!!` for boolean conversion at several places at the moment.

src/web/hooks/useTiming.js
```diff
@@ -0,0 +1,109 @@
...
+              timer.startTimer();
+            })
+            .catch(() => {
+              updateTimerId(undefined);
+            });
+        } else {
+          updateTimerId(undefined);
+        }
+      }, timeoutValue),
+    );
```
daniele-mng, line 71: Maybe `try/catch` and `async/await` are a bit more readable. Something like this?

```suggestion
    updateTimerId(
      setTimeout(async () => {
        log.debug('Timer with id', timer.timerId, 'fired.');
    
        try {
          const promise = await timer.doFunc();
          updateTimerId(undefined);
          if (promise) {
            timer.startTimer();
          }
        } catch  {
          updateTimerId(undefined);
        }
      }, timeoutValue),
    );
```

bjoernricks: Currently it is not necessary that `timer.doFunc` is a promise. It can be a normal function. Therefore I am using ` if isDefined(promise?.then)` to check whether it is a promise or not. Maybe something like this is possible
```js
        try {
          const promise = timer.doFunc();
          updateTimerId(undefined);
          if (promise?.then) {
            await promise;
            timer.startTimer();
          }
        } catch  {
          updateTimerId(undefined);
        }
      }, timeoutValue),
```
?

daniele-mng: Ok, yeah, got it. Looks already much more readable with the try/catch :rocket:

src/web/hooks/__tests__/useTiming.jsx
```diff
@@ -0,0 +1,142 @@
...
+      <button onClick={clearTimer} data-testid="clearTimer"></button>
+      <span data-testid="value">{value}</span>
+      <span data-testid="isRunning">{'' + isRunning}</span>
+    </>
+  );
+};
+
+const runTimers = async () => {
+  await act(async () => {
+    // await testing.runAllTimersAsync();
```
daniele-mng, line 35:
```suggestion
    
```

daniele-mng (approved):
LGTM!

PR #4060, opened 2024-06-11
Change: withIconSize HOC to hook

daniele-mng:
## What

 Refactored HOC `withIconSize` to hook `useIconSize`

## Why

To improve the code readability. 

## References

GEA-568

## Checklist

- [ ] Tests


bjoernricks (commented):
Nice! Could you do me a favor and just drop all the snapshot tests? They aren't really useful at the end and we have to remove them when switching to the new UI anyway. The simplest test should just check if the component renders and if it is in the DOM. That should catch major rendering issues when doing some refactoring.

bjoernricks (approved):

PR #4071, opened 2024-06-17
add: new UI error boundary

daniele-mng:
## What

- Make error boundary UI similar to opensight

## References

GEA-598

## Checklist

- [ ] Tests


src/web/components/error/errormessage.jsx
```diff
@@ -22,7 +22,8 @@ import {isDefined} from 'gmp/utils/identity';
 import Divider from 'web/components/layout/divider';
 import Layout from 'web/components/layout/layout';
 
-import StNonAvailableIcon from 'web/components/icon/stnonavailableicon';
+import {AlertCircle} from 'lucide-react';
```
bjoernricks, line 25: Is 'lucide-react' already a direct dependency of GSA?

daniele-mng: I think through the opensight UI lib.

bjoernricks: IMHO dependencies used in imports should be added as direct dependencies in the package.json. If it is added only via opensight ui it is a indirect dependency.

bjoernricks (approved):

PR #4073, opened 2024-06-18
Change: Use `.js` as file suffix for hooks

bjoernricks:
## What

Use `.js` as file suffix for hooks

## Why

Hooks usually don't contain JSX and therefore should use `.js` for the file suffix. Currently this is mixed up between `.jsx` and `.js`.


src/web/components/form/ErrorMarker.jsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import React from 'react';
+
+import styled from 'styled-components';
+
+import {styledExcludeProps} from 'web/utils/styledConfig';
+import Theme from 'web/utils/theme';
+
+const StyledMarker = styledExcludeProps(styled.div, ['isVisible'])`
```
daniele-mng, line 13: Alternatively, to  `styledExcludeProps` it should be possible to do `$prop`

bjoernricks: I cleaned up the component a bit instead of leaving it as it is.

daniele-mng (approved):

PR #4074, opened 2024-06-18
Refactor PortLists page to a HOC less entities page

bjoernricks:
## What

Refactor PortLists page to a HOC less entities page

## Why

Use the PortLists page as an example on howto refactor an entities page to use the new hooks instead of the withEntitiesContainer HOC. Despite having some more lines of code it should be easier to understand then before.

## References
HOC less entities pages

### DIFF
--- a/src/web/pages/portlists/listpage.jsx
+++ b/src/web/pages/portlists/listpage.jsx
@@ -3,37 +3,59 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import React from 'react';
+import React, {useCallback, useEffect, useState} from 'react';
 
-import _ from 'gmp/locale';
+import {useDispatch} from 'react-redux';
 
 import {PORTLISTS_FILTER_FILTER} from 'gmp/models/filter';
 
-import IconDivider from 'web/components/layout/icondivider';
-import PageTitle from 'web/components/layout/pagetitle';
+import {isDefined, hasValue} from 'gmp/utils/identity';
+
+import useCapabilities from 'web/utils/useCapabilities';
+import useUserSessionTimeout from 'web/utils/useUserSessionTimeout';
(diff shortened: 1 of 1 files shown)


src/web/pages/portlists/listpage.jsx
```diff
@@ -48,80 +70,273 @@ const ToolBarIcons = withCapabilities(
...
 
-export default withEntitiesContainer('portlist', {
-  entitiesSelector,
-  loadEntities,
-})(PortListsPage);
+const PortListsPage = () => {
+  const [_] = useTranslation();
+  const gmp = useGmp();
+  const dispatch = useDispatch();
+  const [tagsDialogVisible, setTagsDialogVisible] = useState(false);
```
daniele-mng, line 97:
```suggestion
  const [isTagsDialogVisible, setIsTagsDialogVisible] = useState(false);
```

src/web/pages/portlists/listpage.jsx
```diff
@@ -48,80 +70,273 @@ const ToolBarIcons = withCapabilities(
...
+      promise = entitiesCommand.delete(selectedEntities);
+    } else if (selectionType === SelectionType.SELECTION_PAGE_CONTENTS) {
+      promise = entitiesCommand.deleteByFilter(filter);
+    } else {
+      promise = entitiesCommand.deleteByFilter(filter.all());
+    }
+
+    renewSession();
+
+    return promise.then(() => {
```
daniele-mng, line 208: Nitpick: Could we use try/catch and async/await instead of `.then`.

bjoernricks: Could you check twice if the code with async/await is correct now? I am used to promises ;-)

daniele-mng (approved):

PR #4082, opened 2024-06-27
Fix: Allow applying report configs to delta reports.

a-h-abdelsalam:
## What

Allow applying report configs when exporting delta reports. The missing props were passed to the component `DownloadReportDialog` from `DeltaReportDetails` similar to how it is done from `ReportDetails`.

## Why
Trying to export delta reports from GSA was throwing an error caused by report configs not being defined.

## References
GEA-609


daniele-mng (approved):
LGTM!

PR #4095, opened 2024-07-09
Add: conditional route component for feature flag

daniele-mng:
## What

The `ConditionalRoute` component utilizes this hook to control access to routes. It decides whether a particular route should be accessible or if it should redirect to a "not found" page, depending on the feature's enablement status.

## Why

The component `ConditionalRoute`, it is reusable across the application and the provides a declarative way to apply these toggles to route accessibility.

## References

GEA-643

## Checklist

- [x] Tests

### DIFF
--- a/src/web/components/conditionalRoute/ConditionalRoute.jsx
+++ b/src/web/components/conditionalRoute/ConditionalRoute.jsx
@@ -1,0 +1,29 @@
+/* SPDX-FileCopyrightText: 2024 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import PropTypes from 'prop-types';
+
+import {Route, Redirect} from 'react-router-dom';
+import useCapabilities from 'web/hooks/useCapabilities';
+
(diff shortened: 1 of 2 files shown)


src/web/components/conditionalRoute/ConditionalRoute.jsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import PropTypes from 'prop-types';
+
+import {Route, Redirect} from 'react-router-dom';
+import useCapabilities from 'web/hooks/useCapabilities';
+
+const ConditionalRoute = ({component: Component, feature, ...rest}) => {
+  const capabilities = useCapabilities();
+  const isEnabled = capabilities._featuresEnabled[feature];
```
a-h-abdelsalam, line 13: I think you might want to use the `featureEnabled` method instead.
```suggestion
  const isEnabled = capabilities.featureEnabled(feature);
```

daniele-mng: Yes, makes sense. Thanks!

a-h-abdelsalam (approved):

PR #4120, opened 2024-08-05
Add feedback notification for feed sync

daniele-mng:
What

- Added a new Feed Sync Notification component to improve user messaging during feed synchronization or error fetching the feed.
- Updated the feed status pages to display the current feed synchronization status.
- Disabled the task scanner start and resume buttons if the feed is currently syncing.

Why

These changes are necessary because if the feed is syncing, the scanner tasks would not run, and users would not be informed about it. By providing clear notifications and disabling the relevant buttons, users are better informed about the current state of the system and can avoid attempting actions that would not succeed.

## References

GEA-417

## Checklist

- [ ] Tests


a-h-abdelsalam (approved):

4D617274696E (approved):
Reviewed German translations, looks good.

PR #4123, opened 2024-08-13
Add confirmation dialog for bulk trashcan or delete table action

daniele-mng:
## What

- Add a confirmation dialog for table bulk actions move to trashcan/delete

## Why

- To prevent accidental deletions by ensuring users confirm their actions before proceeding. This adds an extra layer of safety.

## References

GEA-412

## Checklist

- [x] Tests


src/web/pages/audits/__tests__/listpage.jsx
```diff
@@ -146,7 +147,7 @@ describe('AuditPage tests', () => {
     expect(tableBody.querySelectorAll('tr').length).toEqual(1);
   });
 
-  test('should call commands for bulk actions', async () => {
+  test.only('should call commands for bulk actions', async () => {
```
a-h-abdelsalam, line 150:
```suggestion
  test('should call commands for bulk actions', async () => {
```

a-h-abdelsalam (changes requested):
Looks good to me! However, I see the ticket suggests the confirmation dialog to be optional (set in user settings) and to keep the default behavior without confirmation.

Was this already discussed ?

a-h-abdelsalam (approved):

PR #4144, opened 2024-09-02
Add user setting time date format

daniele-mng:
## What

- Introduced user setting for custom and default system date and time.

## Why

- These changes are necessary to provide users with customizable and localized options for displaying date and time.

## References

GEA-430

## Checklist

 

- [x] Tests


src/web/pages/overrides/detailspage.jsx
```diff
@@ -109,9 +108,11 @@ ToolBarIcons.propTypes = {
   onOverrideEditClick: PropTypes.func.isRequired,
 };
 
-const Details = connect(rootState => ({
-  timezone: getTimezone(rootState),
-}))(({entity, timezone, ...props}) => {
+const Details = connect(rootState => {
+  return {
```
a-h-abdelsalam, line 112: Is this change to explicit return intentional ?

src/gmp/commands/users.js
```diff
@@ -250,9 +255,12 @@ export class UserCommand extends EntityCommand {
 
   saveSettings(data) {
     log.debug('Saving settings', data);
+
     return this.httpPost({
       cmd: 'save_my_settings',
       text: data.timezone,
+      [PARAM_KEYS.DATE]: data.dateFormat,
+      [PARAM_KEYS.TIME]: data.timeFormat,
```
a-h-abdelsalam, line 263: I think you may want to update the values here
```suggestion
      [PARAM_KEYS.DATE]: data.userInterfaceDateFormat,
      [PARAM_KEYS.TIME]: data.userInterfaceTimeFormat,
```

src/web/pages/usersettings/usersettingspage.jsx
```diff
@@ -237,23 +237,45 @@ class UserSettings extends React.Component {
...
-          : userInterfaceLanguage,
+      this.handleInteraction();
+      await gmp.user.saveSetting(
+        'd9857b7c-1159-4193-9bc0-18fae5473a69',
+        data.userInterfaceDateFormat,
+      );
+      await gmp.user.saveSetting(
+        '11deb7ff-550b-4950-aacf-06faeb7c61b9',
+        data.userInterfaceTimeFormat,
       );
```
a-h-abdelsalam, line 254: You may want to remove the explicit `saveSetting` here and instead let `saveSettings` handle the saving. Since this will require the `gsad` updates, the changes in this PR would need to be ported to gsa main as well.
```suggestion
```

a-h-abdelsalam (commented):
Looks good! Only added a few comments.
I cannot judge the German translations so maybe someone else can take a look on that.

src/web/pages/overrides/detailspage.jsx
daniele-mng: No, it's not needed. Reverted.

src/gmp/commands/users.js
daniele-mng: Good catch!

4D617274696E (approved):
Reviewed German translations.

a-h-abdelsalam (approved):

PR #4151, opened 2024-09-05
fix: default filter selection in details page

daniele-mng:
## What

- Default filter should be applied on first render when details page is opened.

## Why

- When details page report is opened, the wrong filter was being applied.
 The default fallback filter was being selected instead of the user default filter.

## References

GEA-675

## Checklist

- [x] Tests


src/web/hooks/usePageFilter.js
```diff
@@ -47,12 +47,14 @@ const useDefaultFilter = pageName =>
  *
  * @param {String} pageName Name of the page
  * @param {Object} options Options object
+ * @param {String} gmpName Name of the GMP for filtering
```
a-h-abdelsalam, line 50: I think GMP name may be more clear
```suggestion
 * @param {String} gmpName GMP name for filtering
```

a-h-abdelsalam (commented):
Looks good. Nice catch!

a-h-abdelsalam (approved):

PR #4161, opened 2024-09-11
Update languages setting

lyace2010:
## What

Add Simplified Chinese in browser language options

## Why

There is already Simplified Chinese in the localized folder, but there is no option to enable it

## Checklist

- [ ] Tests


bjoernricks:
Hi, the current Simplified Chinese translations are heavily outdated. Thus we hesitate to enable them.

PR #4184, opened 2024-10-16
Change: Enable EPSS toggle

a-h-abdelsalam:
## What
Enable EPSS toggle by default

## Why
EPSS should be enabled in 24.10

## References
GEA-718


daniele-mng (approved):

PiRomant:
So, how is this work? I have a `N/A` in every vulnerability

PR #4200, opened 2024-11-05
fix: filtering error on first render for create display

daniele-mng:
##What

Updated the `usePageFilter` hook. Introduced `useRef` and `useState` to manage the `returnedFilter` to avoid directly modifying the redux store.

##Why

- Redux error in the charts on first render when resetting the filter to default. 
- Error example
Screenshot from 2024-11-05 18-02-58
## Checklist

- [ ] Tests

### DIFF
--- a/src/web/hooks/usePageFilter.js
+++ b/src/web/hooks/usePageFilter.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import {useCallback, useEffect, useState} from 'react';
+import {useCallback, useEffect, useState, useRef} from 'react';
 
 import {useSearchParams} from 'react-router-dom';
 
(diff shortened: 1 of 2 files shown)


src/web/hooks/usePageFilter.js
```diff
@@ -140,7 +144,11 @@ const usePageFilter = (
     rowsPerPage = DEFAULT_ROWS_PER_PAGE;
   }
 
-  if (!returnedFilter.has('rows') && isDefined(rowsPerPage)) {
+  if (
+    !returnedFilter.has('rows') &&
+    isDefined(rowsPerPage) &&
+    rowsPerPage === NO_ROW_LIMIT
```
a-h-abdelsalam, line 150: Question: Just for my understanding, what is the reason of adding the check for `rowsPerPage = -1` ?

a-h-abdelsalam (approved):

PR #4232, opened 2024-11-28
Fix: QM first observations

daniele-mng:
## What

- Fixed performance calendar
- Changed import icon
- Fixed Login form casing
- Centered Not found logo
- Fixed Permission Description missing
- Fixed role dialog selection for Create Permission and Create Super Permission
- Updated import icon

## Why

QM First observations

## References

GEA-752
GEA-753
GEA-754
GEA-755
GEA-757
GEA-758
GEA-760

## Checklist

- [ ] Tests


timopollmeier (changes requested):
I've found a few things that should be fixed:
- The translation files should be updated to match the new "Sign in as Guest" text
- Maybe the normal "Sign In" button should also be changed to "Sign in"
- The performance page starts with an empty time range (start and end date/time are the same)
- Clicking the "Report for Last" links on the performance page does not update the time selector.

timopollmeier (approved):

PR #4273, opened 2025-01-07
Fix dockerfile linting issue with using different casing

bjoernricks:
## What

Fix dockerfile linting issue with using different casing.

## Why

See https://docs.docker.com/reference/build-checks/from-as-casing/ for details. This change fixes an warning wich is shown in every PR.

## References

https://github.com/greenbone/gsa/pull/4266/files

grafik


y0urself (approved):

y0urself:
cAsE cLoSeD?

PR #4276, opened 2025-01-09
Bug edit credentials dialog fields not disabled

daniele-mng:
## What

- Migrated the component to RFC
- Fixed the field being disabled on first render if the checkbox is not selected
- Updated test

## Why

- In the credential dialog when in edit mode the fields for replacing `password`, `passphrase`,  `privacy_password`  are not disabled if the checkbox for replacing is not selected on first render.

## References

GEA-834

## Checklist

- [x] Tests

Screenshot from 2025-01-09 12-23-03

### DIFF
--- a/src/gmp/models/credential.js
+++ b/src/gmp/models/credential.js
@@ -71,7 +71,7 @@ export const vFire_credential_filter = credential =>
 export const SNMP_AUTH_ALGORITHM_MD5 = 'md5';
 export const SNMP_AUTH_ALGORITHM_SHA1 = 'sha1';
 
-export const SNMP_PRIVACY_ALOGRITHM_NONE = '';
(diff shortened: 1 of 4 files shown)


src/gmp/models/credential.js
```diff
@@ -71,7 +71,7 @@ export const vFire_credential_filter = credential =>
 export const SNMP_AUTH_ALGORITHM_MD5 = 'md5';
 export const SNMP_AUTH_ALGORITHM_SHA1 = 'sha1';
 
-export const SNMP_PRIVACY_ALOGRITHM_NONE = '';
+export const SNMP_PRIVACY_ALGORITHM_NONE = '';
```
bjoernricks, line 74: nice catch :D

bjoernricks (approved):

PR #4278, opened 2025-01-09
Fix and update translations

bjoernricks:
## What

Fix and update translations

## Why

Use new translation extraction mechanism and get up to date language files, which include all translatable strings.

## References

GEA-835


daniele-mng (approved):
Nice! LGTM!

PR #4281, opened 2025-01-10
change: cve.js for new data structure

daniele-mng:
## What

- Adjust `cve.js` and `cpe.js` to new data structure.

## Why

- Data structure change.

## References

GEA-730

## Checklist

- [ ] Tests

### DIFF
--- a/src/gmp/models/cpe.js
+++ b/src/gmp/models/cpe.js
@@ -39,12 +39,17 @@ class Cpe extends Info {
       delete ret.update_time;
     }
 
-    if (isDefined(ret.raw_data) && isDefined(ret.raw_data['cpe-item'])) {
-      const cpeItem = ret.raw_data['cpe-item'];
-      if (isDefined(cpeItem._deprecated_by)) {
-        ret.deprecatedBy = cpeItem._deprecated_by;
-      }
+    /*
+     * This code includes a backup check for deprecated field `raw_data`.
+     * Once `raw_data` is removed from the API, this backup check can be removed.
+     */
+
+    if (ret.deprecate === 1 && isDefined(ret.deprecated_by)) {
+      ret.deprecatedBy = ret.deprecated_by;
+    } else if (isDefined(ret.raw_data?.['cpe-item']?._deprecated_by)) {
+      ret.deprecatedBy = ret.raw_data['cpe-item']._deprecated_by;
     }
+
     return ret;
   }
 }
--- a/src/gmp/models/cpe.js
+++ b/src/gmp/models/cpe.js
@@ -44,8 +44,8 @@ class Cpe extends Info {
      * Once `raw_data` is removed from the API, this backup check can be removed.
      */
 
-    if (ret.deprecate === 1 && isDefined(ret.deprecated_by)) {
-      ret.deprecatedBy = ret.deprecated_by;
+    if (ret.deprecated === 1 && isDefined(ret.deprecated_by)) {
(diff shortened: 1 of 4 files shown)


src/gmp/models/cpe.js
```diff
@@ -39,12 +39,17 @@ class Cpe extends Info {
...
-      const cpeItem = ret.raw_data['cpe-item'];
-      if (isDefined(cpeItem._deprecated_by)) {
-        ret.deprecatedBy = cpeItem._deprecated_by;
-      }
+    /*
+     * This code includes a backup check for deprecated field `raw_data`.
+     * Once `raw_data` is removed from the API, this backup check can be removed.
+     */
+
+    if (ret.deprecate === 1 && isDefined(ret.deprecated_by)) {
```
a-h-abdelsalam, line 47:
```suggestion
    if (ret.deprecated === 1 && isDefined(ret.deprecated_by)) {
```

src/gmp/models/cpe.js
```diff
@@ -39,12 +39,17 @@ class Cpe extends Info {
...
+      ret.deprecatedBy = ret.deprecated_by;
```
a-h-abdelsalam, line 48: May be here you need to pick the `cpe_id` ?
This is how it looks like in the response:
```
<cpe>
  ...
  <deprecated>1</deprecated>
  <deprecated_by cpe_id="cpe:..."/>
</cpe>
```

src/gmp/models/cve.js
```diff
@@ -154,10 +160,47 @@ class Cve extends Info {
           ret.products = [];
         }
       }
-
       delete ret.raw_data;
     } else {
+      ret.publishedTime = parseDate(ret['published-datetime']);
```
a-h-abdelsalam, line 165: I think the priority here is to pick the JSON fields and if they don't exist fallback to the `raw_data`.

src/gmp/models/cve.js
```diff
@@ -154,10 +160,47 @@ class Cve extends Info {
...
+
+      if (
+        ret.products.length === 0 &&
+        isDefined(element.cve?.configuration_nodes?.node)
+      ) {
+        const nodes = isArray(element.cve.configuration_nodes.node)
+          ? element.cve.configuration_nodes.node
+          : [element.cve.configuration_nodes.node];
+        nodes.forEach(node => {
+          if (isDefined(node.match_string?.matched_cpes?.cpe)) {
```
a-h-abdelsalam, line 192: I think here we need to include the cpes only for the `match_string` whose `vulnerbale` is set to 1. Otherwise, you can skip it.

src/gmp/models/cve.js
```diff
@@ -154,10 +160,47 @@ class Cve extends Info {
           ret.products = [];
         }
       }
-
       delete ret.raw_data;
     } else {
+      ret.publishedTime = parseDate(ret['published-datetime']);
+      ret.lastModifiedTime = parseDate(ret['last-modified-datetime']);
```
a-h-abdelsalam, line 166: Published and Last update time are empty for `cves` to me in GSA, but they work for `cpes`.

a-h-abdelsalam: I think the issue is that `ret` here is the parsed 'cve' element which does not have the published and modified time. So these are always undefined.
Instead `web/pages/cves/detailspage.jsx` needs to be updated with the correct names `creationTime` and `modificationTime`. See `web/pages/cpes/detailspage.jsx` as an example

PR #4284, opened 2025-01-13
Fix: Allow to set the filter via URL query parameter

bjoernricks:
## What

Drop `usePageFilter` `locationQueryFilterString` because the query param filter gathered from `useSearchParams` is the only valid value.

Drop `FilterProvider` `locationQuery` prop because a string was passed instead of the expected query object and because usePageFilter already handles its use case internally via `useSearchParams`.

## Why
Allow to pass filters via URL query argument like `/tasks?filter=name=bar`.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/hooks/__tests__/usePageFilter.jsx
+++ b/src/web/hooks/__tests__/usePageFilter.jsx
@@ -5,7 +5,7 @@
 
 import {describe, test, expect, testing} from '@gsa/testing';
 import Filter, {DEFAULT_FALLBACK_FILTER} from 'gmp/models/filter';
-import {vi} from 'vitest';
+import {beforeEach, vi} from 'vitest';
 import {pageFilter} from 'web/store/pages/actions';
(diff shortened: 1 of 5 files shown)


src/web/hooks/__tests__/usePageFilter.jsx
```diff
@@ -3,35 +3,43 @@
...
 import {loadingActions} from 'web/store/usersettings/defaults/actions';
 import {rendererWith, waitFor} from 'web/utils/testing';
 
 import usePageFilter from '../usePageFilter';
 
-
-const mockUseNavigate = testing.fn();
+let mockSearchParams = {};
+const mockUseNavigate = vi.fn();
+const mockUseSearchParams = vi
```
daniele-mng, line 18:
```suggestion
const mockUseNavigate = testing.fn();
const mockUseSearchParams = testing
```

daniele-mng (approved):

PR #4286, opened 2025-01-14
Fix nvtpreference dialog

daniele-mng:
## What

- Refactored `nvtpreference.jsx` to RFC
- Added test for `nvtpreference.jsx`
- `Radio.jsx` assures that label is a string
- In `nvtpreference.jsx` modified `onChange` to use inline arrow function
   -  Added styling for long strings `TableData`
- Improved layout of the `editnvtdetailsdialog.jsx`

## Why

- If the radio component expects a string, if a value number 0 is passed it results in some unexpected UI issues.
- Not using an arrow in the handler, it will not receive the specific value.
- Long strings in `TableData` created h-overflow.

## References

GEA-841

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/nvts/nvtpreference.jsx
+++ b/src/web/pages/nvts/nvtpreference.jsx
@@ -26,7 +26,7 @@ const StyledTableData = styled(TableData)`
   word-break: break-word;
 `;
 
-const noop_convert = value => value;
+const noopConvert = value => value;
 
 const NvtPreference = ({preference, value = '', onChange}) => {
   const [checked, setChecked] = useState(false);
@@ -50,7 +50,7 @@ const NvtPreference = ({preference, value = '', onChange}) => {
   if (type === 'checkbox') {
     input = (
       <YesNoRadio
-        convert={noop_convert}
+        convert={noopConvert}
         noValue="no"
         value={value}
         yesValue="yes"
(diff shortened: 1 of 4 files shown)


src/web/pages/nvts/nvtpreference.jsx
```diff
@@ -15,127 +16,116 @@ import TextField from 'web/components/form/textfield';
...
+const StyledTableData = styled(TableData)`
+  overflow-wrap: break-word;
+  white-space: normal;
+  word-break: break-word;
+`;
 
-class NvtPreference extends React.Component {
-  constructor(...args) {
-    super(...args);
+const noop_convert = value => value;
```
bjoernricks, line 29: Would you mind converting noop_convert into camel case too in your PR? i think we should convert old stuff to camel case wherever we touch code
```suggestion
const noopConvert = value => value;
```

src/web/pages/nvts/nvtpreference.jsx
```diff
@@ -15,127 +16,116 @@ import TextField from 'web/components/form/textfield';
...
-        <YesNoRadio
-          convert={noop_convert}
-          noValue="no"
-          value={value}
-          yesValue="yes"
-          onChange={this.onPreferenceChange}
+  if (type === 'checkbox') {
+    input = (
+      <YesNoRadio
+        convert={noop_convert}
```
bjoernricks, line 53:
```suggestion
        convert={noopConvert}
```

bjoernricks (approved):

PR #4312, opened 2025-01-30
Fix: remove Radio component title string conversions

daniele-mng:
## What

- Ensure that label is a string

## Why

- Radio component doesn't always have the prop `title`, ex. if `title` is missing, it would display `undefined`.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/components/form/radio.jsx
+++ b/src/web/components/form/radio.jsx
@@ -30,7 +30,7 @@ const Radio = ({
       {...props}
       checked={checked}
       disabled={disabled}
-      label={String(title)}
+      label={title}
       name={name}
       value={value}
       onChange={handleChange}


src/web/components/form/radio.jsx
```diff
@@ -30,7 +31,7 @@ const Radio = ({
       {...props}
       checked={checked}
       disabled={disabled}
-      label={String(title)}
+      label={isString(title) && title}
```
bjoernricks, line 34: Maybe we could check if title has a `toString` method? Something like this?
```suggestion
      label={title?.toString && String(title)}
```

daniele-mng: We could encounter issues with `booleans` and `truthy` values. If we want the component to convert it just in case then we will need.

```js
  const getLabel = title => {
    if (typeof title === 'string' || typeof title === 'number') {
      return String(title);
    }
    return null;
  };
```

bjoernricks: sadly this wont work with the `_l` lazy translation function :-( so we should be really careful what to expect.

bjoernricks (approved):

PR #4335, opened 2025-02-12
Add: Kerberos to Credential and Target

daniele-mng:
## What

- From the credential it's possible to manage the Kerberos credential.
- From the target section it's possible to manage with new Kerberos credential.

## Why

- Support Kerberos

## References

GEA-777

## Checklist

- [x] Tests


src/gmp/commands/__tests__/credential.js
```diff
@@ -0,0 +1,93 @@
...
+    const credential = Credential.fromElement(elem);
+
+    expect(credential.scanners.length).toEqual(1);
+
+    const [scanner] = credential.scanners;
+    expect(scanner).toBeInstanceOf(Model);
+    expect(scanner.id).toEqual('s1');
+    expect(scanner.entityType).toEqual('scanner');
+  });
+});
```
bjoernricks, line 93: Should be put into the model tests at gmp/models/__tests__/credential.js

bjoernricks (commented):
Besides moving the credential model into the correct file it looks very good 👍🏻

bjoernricks (approved):

PR #4341, opened 2025-02-20
Ensure name is string

bjoernricks:
## What

Ensure name properties of our models are always of type string.

## Why

Fix a crash of the group dialog when a user name only contains digits. Our XML to JS parsing library converts converts XML content and attributes which only contain digits as numbers. Sadly our MultiSelect component can't handle non-string values and labels anymore.

## References

https://forum.greenbone.net/t/error-in-new-group-dialog-when-there-are-users-with-login-names-formed-by-numbers-only/20465/3
https://jira.greenbone.net/browse/GEA-922

## Checklist

- [x] Tests

### DIFF
--- a/src/gmp/parser.js
+++ b/src/gmp/parser.js
@@ -224,3 +224,16 @@ export const parseBoolean = value => {
   }
   return value === true;
 };
+
+/**
+ * Parses the given value into a string if it is defined.
+ *
+ * @param {*} value - The value to be parsed into a string.
+ * @returns {string|undefined} The parsed string if the value is defined, otherwise undefined.
+ */
+export const parseToString = value => {
+  if (isDefined(value)) {
+    return String(value);
+  }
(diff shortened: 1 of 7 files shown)


src/gmp/parser.js
```diff
@@ -224,3 +224,16 @@ export const parseBoolean = value => {
...
   return value === true;
 };
+
+/**
+ * Parses the given value into a string if it is defined.
+ *
+ * @param {*} value - The value to be parsed into a string.
+ * @returns {string|undefined} The parsed string if the value is defined, otherwise undefined.
+ */
+export const parseString = value => {
```
daniele-mng, line 234: Suggestion: If `value` can be any type, and it's ok to we parse objects and arrays, then maybe:

```suggestion
export const parseToString = value => {
```

daniele-mng (approved):

PR #4352, opened 2025-02-25
Change: Move and improve entity goto functions

bjoernricks:
## What

Move and improve entity goto functions
## Why

* Move the goToDetails and goToList functions into an own module
* Extend them to throw errors if the necessary navigate function is not provided
* Allow to pass the navigate function directly or as part of an object
* Add tests for both functions

## References

Part of https://jira.greenbone.net/browse/GEA-935

## Checklist

- [x] Tests


src/web/entity/navigation.js
```diff
@@ -0,0 +1,56 @@
...
+    navigate = navigate.navigate;
+
+    if (!isFunction(navigate)) {
+      throw new Error('navigate function is required for goToDetails');
+    }
+  }
+  return ({data}) => {
+    return navigate('/' + type + '/' + data.id);
+  };
+};
```
daniele-mng, line 31:
```suggestion
  return ({data}) => navigate(`/${type}/${data.id}`);
```

src/web/entity/navigation.js
```diff
@@ -0,0 +1,56 @@
...
+    // we expect an object with a navigate function
+    navigate = navigate.navigate;
+
+    if (!isFunction(navigate)) {
+      throw new Error('navigate function is required for goToList');
+    }
+  }
+  return () => {
+    return navigate('/' + type);
+  };
```
daniele-mng, line 55:
```suggestion
return () => navigate(`/${type}`);
```

daniele-mng (approved):

PR #4371, opened 2025-03-04
Allow to pass a convert function to TextField

bjoernricks:
## What

Allow to pass a convert function to TextField

## Why
Add a convert function to TextField which is used for example in the port range dialog. It allows to change the value automatically to only valid values. This feature seems to have been removed when changing to the new UI.

## References

Use in the PortRangeDialog component.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/components/form/__tests__/TextField.test.jsx
+++ b/src/web/components/form/__tests__/TextField.test.jsx
@@ -7,7 +7,6 @@ import {describe, test, expect, testing} from '@gsa/testing';
 import TextField from 'web/components/form/TextField';
 import {render, fireEvent, screen} from 'web/utils/Testing';
 
-
 describe('TextField tests', () => {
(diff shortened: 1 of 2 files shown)


src/web/components/form/__tests__/TextField.test.jsx
```diff
@@ -70,4 +69,24 @@ describe('TextField tests', () => {
...
+    render(
+      <TextField
+        convert={convert}
+        data-testid="input"
+        value="foo"
+        onChange={onChange}
+      />,
+    );
+
+    const element = screen.getByTestId('input');
```
daniele-mng, line 86: Q: Since we have  `data-testid="form-input"` in `TextField` could we use it instead of adding a new one?

daniele-mng (approved):

PR #4377, opened 2025-03-06
Fix passing true as actions column to withEntitiesHeader

bjoernricks:
## What

Fix passing true as actions column to withEntitiesHeader

## Why

List pages that allow bulk selections but have no dedicated actions column like CVEs pass false for the action column when using withEntitiesHeader. This allows displaying the action column conditionally if the user selection is selected for the list table. Additionally the variables are changed to camelCase naming instead of underscore.

This changes removes proptype warnings in a bunch of tests and during development.

### DIFF
--- a/src/web/entities/Header.jsx
+++ b/src/web/entities/Header.jsx
@@ -31,9 +31,9 @@ const defaultActions = (
  * If actionsColumn is false, no actions (a null value in React) will be passed to
  * the Component.
  *
- * @param {Element}   actionsColumn   React element, undefined or boolean value.
- * @param {Object}    options         Default properties for Component.
- * @param {Component} Component       React component rendered as header
+ * @param {React.Element|undefined|boolean} actionsColumn - React element, undefined, or boolean value.
+ * @param {Object} options - Default properties for the Component.
+ * @param {React.Component} Component - React component rendered as header.
  *
  * @return {React.Component} A new EntitiesHeader component.
  */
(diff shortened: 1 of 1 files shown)


src/web/entities/Header.jsx
```diff
@@ -11,45 +11,47 @@ import TableRow from 'web/components/table/Row';
...
+ * If actionsColumn is true, a default table head column will be passed as actionsColumn
+ * to the Component if the current selectionType (passed via props) is SELECTION_USER.
  *
- * If actions is false no actions (a null value in react) will be passed to
- * Component.
+ * If actionsColumn is false, no actions (a null value in React) will be passed to
+ * the Component.
  *
- * @param {Element}   actions_column  React element, undefined or boolean value.
+ * @param {Element}   actionsColumn   React element, undefined or boolean value.
```
daniele-mng, line 34: I think this should be updated:

```suggestion
 * @param {Element | Boolean | undefined} actionsColumn React element, undefined or boolean value.
```

src/web/entities/Header.jsx
```diff
@@ -78,12 +80,12 @@ export const withEntitiesHeader =
  *                             }, {
  *                               ...
  *                             }, ... ]
- * @param {Element} actions_column   React element, undefined or boolean value.
+ * @param {Element} actionsColumn   React element, undefined or boolean value.
```
daniele-mng, line 83:
```suggestion
 * @param {Element | Boolean | undefined} actionsColumn React element, undefined or boolean value.
```

src/web/entities/Header.jsx
```diff
@@ -11,45 +11,47 @@ import TableRow from 'web/components/table/Row';
...
 
     const HeaderWrapper = props => {
       const {selectionType} = props;
-      let column = actions_column;
+      let column = actionsColumn;
 
-      if (actions_column && selectionType === SelectionType.SELECTION_USER) {
+      if (actionsColumn && selectionType === SelectionType.SELECTION_USER) {
         column = <TableHead width="6em">{_('Actions')}</TableHead>;
+      } else if (actionsColumn === true) {
```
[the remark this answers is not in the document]
bjoernricks: `true` has currently the meaning of some "do it automatically" kind. i don't really like it as it is currently but i hesitate to touch this old code. at least in this PR. just wanted to fix it, so we can use the expected behavior without getting warnings in the console.

src/web/entities/Header.jsx
bjoernricks: i didn't update the docs of the other HOC :-/

daniele-mng (approved):

PR #4383, opened 2025-03-07
Entity hooks

bjoernricks:
## What

Introduce useEntityXyz action hooks

## Why

Allow to replace the render props based EntityComponent with hooks.

## References

https://jira.greenbone.net/browse/GEA-935

## Checklist

- [x] Tests

### DIFF
--- a/src/web/entity/EntityComponent.jsx
+++ b/src/web/entity/EntityComponent.jsx
@@ -3,13 +3,10 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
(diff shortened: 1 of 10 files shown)


src/web/entity/hooks/utils.js
daniele-mng, line 1: Thought: Could we call the file `actionFunction.js` instead of `utils.js`? Since I think we have already too many `utils`, and there is only 1 fn in the file.

bjoernricks: of course!

daniele-mng (approved):

PR #4387, opened 2025-03-10
Replace withUserName by useUserName in task Row

bjoernricks:
## What

Replace withUserName by useUserName in task Row

## Why

Get rid of the withUserName HOC and instead use the useUserName hook.


daniele-mng:
Is this the last file with `withUserName` maybe we can delete it.

daniele-mng (approved):

PR #4394, opened 2025-03-11
fix: stacking modal overlay opacity

daniele-mng:
## What

- Ensure that only the top modal is applying the overlay.

## Why

- Previously, each modal would apply and overlay increasing opacity,

## References

GEA-962

## Checklist

- [ ] Tests


bjoernricks (approved):
Wow :flushed:

PR #4402, opened 2025-03-13
Fix: Ensure time is valid before settings

janlucasschwoerer-greenbone:
## What
Add check to ensure time is valid before updating state

## Why
A method call on an invalid moment throws an error

## References
https://jira.greenbone.net/browse/GEA-948

### DIFF
--- a/src/web/pages/schedules/Dialog.jsx
+++ b/src/web/pages/schedules/Dialog.jsx
@@ -215,12 +215,16 @@ const ScheduleDialog = ({
 
     if (type === 'startTime') {
       const newStartDate = startDate.clone().hour(hour).minute(minute);
-      setStartDate(newStartDate);
-      setStartTime(selectedTime);
+      if (newStartDate.isValid()) {
+        setStartDate(newStartDate);
+        setStartTime(selectedTime);
+      }
     } else if (type === 'endTime') {
       const newEndDate = endDate.clone().hour(hour).minute(minute);
-      setEndDate(newEndDate);
-      setEndTime(selectedTime);
+      if (newEndDate.isValid()) {
+        setEndDate(newEndDate);
+        setEndTime(selectedTime);
+      }
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/web/pages/schedules/Dialog.jsx
```diff
@@ -213,10 +213,18 @@ const ScheduleDialog = ({
 
     if (type === 'startTime') {
       const newStartDate = startDate.clone().hours(hour).minutes(minute);
+      if (!newStartDate.isValid) {
+        return
+      }
+
       setStartDate(newStartDate);
       setStartTime(selectedTime);
```
bjoernricks, line 221:
```suggestion
      if (newStartDate.isValid()) {
        setStartDate(newStartDate);
        setStartTime(selectedTime);
      }
```

src/web/pages/schedules/Dialog.jsx
```diff
@@ -213,10 +213,18 @@ const ScheduleDialog = ({
...
     } else if (type === 'endTime') {
       const newEndDate = endDate.clone().hours(hour).minutes(minute);
+      if (!newEndDate.isValid) {
+        return
+      }
+
       setEndDate(newEndDate);
       setEndTime(selectedTime);
```
bjoernricks, line 229:
```suggestion
      if (newEndDate.isValid()) {
        setEndDate(newEndDate);
        setEndTime(selectedTime);
      }
```

bjoernricks (changes requested):
`momenjs.isValid` is a function and therefore needs to be called. Also putting the setters in the if block makes the code a bit shorter. I am not sure if the indentation of my suggested changes is correct.

bjoernricks:
Adding a test to ensure the behavior would be nice too.

daniele-mng:
I guess this also happening here: https://github.com/greenbone/gsa/blob/804ea7c23efcb03233d7e9890dc8f584b705ee6e/src/web/pages/performance/startendtimeselection.jsx#L57

Does it happen with the other time pickers?

bjoernricks (approved):

PR #4405, opened 2025-03-13
Display errors during port list update and creation

bjoernricks:
## What

Display errors during port list update and creation

## Why

Ensure errors are displayed when saving or creating a port list fails.

## References

https://jira.greenbone.net/browse/GEA-935

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/portlists/PortListComponent.jsx
+++ b/src/web/pages/portlists/PortListComponent.jsx
@@ -183,7 +183,14 @@ class PortListComponent extends React.Component {
     return Promise.all(promises)
       .then(() => save(data))
       .then(() => this.closePortListDialog())
-      .catch(error => this.props.onSaveError(error));
+      .catch(error => {
+        if (data?.id && isDefined(this.props.onSaveError)) {
+          return this.props.onSaveError(error);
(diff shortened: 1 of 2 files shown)


src/web/pages/portlists/PortListComponent.jsx
```diff
@@ -183,7 +183,18 @@ class PortListComponent extends React.Component {
...
+          if (isDefined(this.props.onSaveError)) {
+            return this.props.onSaveError(error);
+          }
+        } else {
+          if (isDefined(this.props.onCreateError)) {
+            return this.props.onCreateError(error);
+          }
+        }
+        return Promise.reject(error);
+      });
```
daniele-mng, line 197: Q: could we reduce it to something like this? 
In any case, L:191 could become `else if`

```suggestion
      .catch(error => {
        if (data.id && isDefined(this.props.onSaveError)) {
          return this.props.onSaveError(error);
        } else if (!data.id && isDefined(this.props.onCreateError)) {
          return this.props.onCreateError(error);
        }
        return Promise.reject(error);
      });
```

bjoernricks: Of course

PR #4417, opened 2025-03-27
Types #2

bjoernricks:
## What

Use TypeScript for more and more component, functions and classes.

## Why

Catch issues already on compile time.

## Checklist

- [x] Tests

PR #4433, opened 2025-04-01
Configurable severity levels

bjoernricks:
## What

Implement a setting for the severity rating. The setting will allow to switch the used severity classes between the levels from CVSSv2 and CVSSv3.

## Why

Currently the CVSSv3 severity classes are not enabled by default in gvmd. Therefore it has to be configurable in GSA too.

## References
https://jira.greenbone.net/browse/GEA-999

Requires #4417
Requires #4438

## Checklist

- [x] Tests


src/web/components/bar/SeverityBar.tsx
```diff
@@ -18,18 +18,26 @@ import {
...
+  const gmp = useGmp();
+  const severityRating = gmp.settings.severityRating;
+  let cvss: number | undefined;
+  let threat: ResultSeverityRiskFactor | undefined;
+  let title: string;
 
   if (isDefined(severity)) {
     cvss = parseFloat(severity);
-    threat = resultSeverityRiskFactor(cvss);
+    // @ts-expect-error
```
daniele-mng, line 39: Any other options instead of suppressing the erro?

src/web/pages/hosts/dashboard/HostsVulnScoreDisplay.jsx
```diff
@@ -15,20 +17,22 @@ import DataTableDisplay from 'web/components/dashboard/display/DataTableDisplay'
...
 const ToolTip = styled.div`
   font-weight: normal;
   text-align: center;
   line-height: 1.2em;
 `;
 
-const transformVulnScoreData = (data = {}) => {
+const transformVulnScoreData = (
+  data = {},
+  {severityRating = DEFAULT_SEVERITY_RATING},
```
daniele-mng, line 34: Default parameter should be after.
```suggestion
  {severityRating = DEFAULT_SEVERITY_RATING},
  data = {},

```

src/web/pages/hosts/dashboard/HostsVulnScoreDisplay.jsx
```diff
@@ -69,65 +76,55 @@ const transformVulnScoreData = (data = {}) => {
...
-  constructor(...args) {
-    super(...args);
-
-    this.handleDataClick = this.handleDataClick.bind(this);
-  }
-
-  handleDataClick(data) {
-    const {navigate} = this.props;
-
+let HostsVulnScoreDisplay = ({filter, ...props}) => {
```
daniele-mng, line 79: Could this be a `const` instead of `let`. Since it gets exported?

src/web/pages/hosts/dashboard/HostsTopologyDisplay.jsx
```diff
@@ -99,64 +99,54 @@ const transformTopologyData = (data = []) => {
...
-  constructor(...args) {
-    super(...args);
-
-    this.handleDataClick = this.handleDataClick.bind(this);
-  }
-
-  handleDataClick(data) {
-    const {navigate} = this.props;
-
+let HostsTopologyDisplay = ({filter, ...props}) => {
```
daniele-mng, line 102: Could this be a const instead of let. Since it gets exported?

src/web/components/bar/SeverityBar.tsx
bjoernricks: Of course. It even shows a logic error.

src/web/pages/hosts/dashboard/HostsVulnScoreDisplay.jsx
bjoernricks: That doesn't work here. The function is generic. First argument is the data from the backend and second arg contains props passed trough DataDisplay into the function. I am not sure anymore why data is set to an empty object by default but i guess it's to avoid issues when the data is loaded currently.

src/web/components/dashboard/display/DataDisplay.tsx
```diff
@@ -315,10 +471,12 @@ class DataDisplay extends React.Component {
...
               </IconDivider>
             </IconBar>
-            {showFilterString && (
+            {showFilterString && isDefined(filter) && (
               <FilterString>
                 ({_('Applied filter: ')}
+                {/* @ts-ignore-error */}
                 <b>{filter.name}</b>&nbsp;
+                {/* @ts-ignore-error */}
                 <i>{filter.simple().toFilterString()}</i>)
```
daniele-mng, line 480:
```suggestion
                <i>{filter?.simple().toFilterString()}</i>)
```

src/web/components/dashboard/display/DataDisplay.tsx
```diff
@@ -315,10 +471,12 @@ class DataDisplay extends React.Component {
                   })}
               </IconDivider>
             </IconBar>
-            {showFilterString && (
+            {showFilterString && isDefined(filter) && (
               <FilterString>
                 ({_('Applied filter: ')}
+                {/* @ts-ignore-error */}
                 <b>{filter.name}</b>&nbsp;
```
daniele-mng, line 478: Not sure for this comment if we can remove it some how. Also `isDefined` on line 474 should ensure the values are available?

src/web/components/dashboard/display/DataDisplay.tsx
```diff
@@ -5,18 +5,101 @@
...
+  id: string;
+  width: number;
+  height: number;
+  svgRef: React.RefObject<SVGSVGElement>;
+  data: TData[];
+  state: TState;
+  setState: SetStateFunc<TState>;
+}
+
+type TransformFunc<TData, TTransformedData, TTransformProps = object> = (
```
daniele-mng, line 61: Nitpick: `object` and on line 70.

src/web/components/dashboard/display/DataDisplay.tsx
bjoernricks: sadly not. seems to be a bug in the type checker. that's why i had to add it. filter is defined here so it shouldn't complain.

src/web/components/dashboard/display/DataDisplay.tsx
bjoernricks: I would leave this at is is because the check above already ensured filter is defined and this is just a ts check bug.

daniele-mng (approved):

PR #4438, opened 2025-04-02
Use TypeScript for the testing module

bjoernricks:
## What
Use TypeScript for the testing module

## Why

Add types for all testing functions.

### DIFF
--- a/src/web/utils/Testing.tsx
+++ b/src/web/utils/Testing.tsx
@@ -22,10 +22,12 @@ import {
   renderHook as rtlRenderHook,
 } from '@testing-library/react/pure';
 import userEvent, {PointerEventsCheckLevel} from '@testing-library/user-event';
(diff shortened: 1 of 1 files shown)


src/web/utils/Testing.tsx
```diff
@@ -124,6 +126,16 @@ const TestingLicenseProvider = withProvider(
...
 )(LicenseProvider);
 
+interface RendererOptions {
+  capabilities?: object | true;
+  gmp?: object;
+  store?: object | boolean;
+  license?: object;
+  router?: boolean;
+  route?: string;
+  showLocation?: boolean;
```
daniele-mng, line 136: For the definitions could we use `Record<string, unknown>` instead of `object`?  It's more specific for JS objects. Otherwise, for TS `object`  could be a non-primitive like fn or array.

PR #4463, opened 2025-04-23
Refactor translations to context

daniele-mng:
## What

- Centralized locale in Context Provider, deprecate locale in redux store and `localeObserver`
- Improve Community Feed Notification / close notifications on `LoginPage`

## Why

- Improve locale management and UX switch between languages.

## References

GEA-1034

## Checklist

- [x] Tests


src/web/components/structure/LanguageSwitch.jsx
```diff
@@ -34,25 +37,34 @@ const LanguageSwitch = () => {
...
+      return;
+    }
+
     try {
-      changeLanguage(nextLanguage);
+      const newLanguage = nextLanguage;
+      setIsChangingLanguage(true);
 
-      await gmp.user.saveSetting(SETTING_ID_LOCALE, nextLanguage);
+      setLocale(newLanguage);
```
bjoernricks, line 48: Should not be called here and it is actually already called in setLanguage
```suggestion
```

src/web/components/structure/LanguageSwitch.jsx
```diff
@@ -34,25 +36,44 @@ const LanguageSwitch = () => {
...
+
+      await new Promise((resolve, reject) => {
+        i18n.changeLanguage(newLanguage, err => {
+          if (err) {
+            reject(err instanceof Error ? err : new Error(err));
+          } else {
+            resolve();
+          }
+        });
+      });
```
bjoernricks, line 55: I don't think this should be necessary. But ff we really need to wait until the change is done, I prefer to change `setLanguage` of `useLanguage` into a promise.
```suggestion
```

src/web/components/structure/LanguageSwitch.jsx
```diff
@@ -34,25 +36,44 @@ const LanguageSwitch = () => {
...
 
-      await gmp.user.saveSetting(SETTING_ID_LOCALE, nextLanguage);
+      gmp.setLocale(newLanguage);
```
bjoernricks, line 57:
```suggestion
```

src/web/hooks/useTranslation.ts
```diff
@@ -30,13 +32,16 @@ type TranslationHook = {
...
-  // provide same interface as useTranslation from i18next
-  // this will allow to use react-i18next in future if required
-  const ret = [_, i18next, true] as TranslationHook;
-  ret.t = _;
-  ret.i18n = i18next;
-  ret.ready = true;
-  return ret;
+  const [language] = useLanguage();
+
+  return useMemo(() => {
```
bjoernricks, line 37: Would be nice to keep the comment about the intention of the details of this hook.
```suggestion
  return useMemo(() => {
    // provide same interface as useTranslation from i18next
    // this will allow to use react-i18next in future if required
```

src/web/components/provider/LanguageProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const LanguageContext = createContext<LanguageContextProps>({
+  language: '',
+  setLanguage: () => {},
+});
+
+export const LanguageProvider: React.FC<LanguageProviderProps> = ({
+  children,
+}) => {
```
bjoernricks, line 29:
```suggestion
}) => {
  const gmp = useGmp();
```

src/web/components/provider/LanguageProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+        unsubscribe();
+      }
+    };
+  }, []);
+
+  const setLanguage = useCallback(
+    (newLang: string) => {
+      if (newLang !== languageState) {
+        setLanguageState(newLang);
+        changeGmpLocale(newLang);
```
bjoernricks, line 52:
```suggestion
        changeGmpLocale(newLang);
        gmp.setLocale(newLang);
```

src/web/components/provider/LanguageProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+    };
+  }, []);
+
+  const setLanguage = useCallback(
+    async (newLang: string) => {
+      if (newLang !== languageState) {
+        // Set locale in gmp
+        changeGmpLocale(newLang);
+        gmp.setLocale(newLang);
```
bjoernricks, line 57: I would love to get rid of this unnecessary extra API which internally calls changeGmpLocal (again) and sets the locale in the browser session via the GmpSettings. Therefore we should just use the following here
```suggestion
        // store language in the session
        gmp.settings.locale = newLang;
```

src/web/components/provider/LanguageProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+  language: '',
+  setLanguage: async () => {},
+});
+
+export const LanguageProvider: React.FC<LanguageProviderProps> = ({
+  children,
+}) => {
+  const gmp = useGmp();
+  const [languageState, setLanguageState] = useState<string>(
+    getLocale() ?? 'en',
```
bjoernricks, line 35: Nitpick, can we introduce a `const DEFAULT_LANGUAGE = 'en';` variable here (or in gmp/locale/lang)?

src/web/components/provider/LanguageProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  useEffect(() => {
+    const unsubscribe = onLanguageChange((newLang: string) => {
+      setLanguageState(currentLang =>
+        currentLang !== newLang ? newLang : currentLang,
```
bjoernricks, line 41: Hmm do we need to update the language in the user settings, dayjs and the browser session here too 🤔 ?

bjoernricks (approved):

bjoernricks (approved):

PR #4468, opened 2025-04-25
Refactor GMP

bjoernricks:
## What

Use TypeScript for the GMP class and base command classes.

## Why

With this change it should get easier to determine which data all the gmp commands like `gmp.portlists.getAll` and similar will return. A huge improvement is for example the type information of the `getAggregates` calls which are used in the dashboard display for all the chart data.

## References

https://jira.greenbone.net/browse/GEA-1047

## Checklist

- [x] Tests


src/gmp/collection/parser.ts
```diff
@@ -0,0 +1,259 @@
...
+
+interface ElementCounts {
+  page?: number;
+  __text?: number;
+  filtered?: number;
+}
+
+export function parseInfoEntities<TModel extends Model>(
+  response: InfoElement,
+  name: string,
```
daniele-mng, line 71: I think this is not used, maybe `_name` if we cannot remove it.

src/gmp/models/actionresult.ts
```diff
@@ -0,0 +1,30 @@
...
+interface ActionResultElement extends Element {
+  action_result: {
+    id: string;
+    action: string;
+    message: string;
+  };
+}
+
+class ActionResult {
+  readonly action: String;
```
daniele-mng, line 17:
```suggestion
  readonly action: string;
```

src/gmp/models/portlist.ts
```diff
@@ -0,0 +1,121 @@
...
+    const ret = super.parseElement(element) as PortListProperties;
+
+    const ranges =
+      isDefined(ret.port_ranges) && isDefined(element.port_ranges)
+        ? element.port_ranges.port_range
+        : [];
+
+    ret.port_ranges = map(ranges, (range: PortRangeElement) => {
+      range.port_list_id = ret.id;
+      return PortRange.fromElement(range as Element) as PortRange;
```
daniele-mng, line 101: May not need as PortRange
```suggestion
      return PortRange.fromElement(range as Element);
```

src/gmp/commands/entities.ts
```diff
@@ -0,0 +1,317 @@
...
+    this.clazz = clazz;
+    this.name = name;
+  }
+
+  getCollectionListFromRoot(root: TRoot): CollectionList<TModel> {
+    const response = this.getEntitiesResponse(root);
+    return parseCollectionList(response, this.name, this.clazz);
+  }
+
+  getEntitiesResponse(root: TRoot): TEntitiesResponse {
```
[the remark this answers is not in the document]
bjoernricks: The change was by intention but to make the purpose more obvious I changes the EntityCommand and EntitiesCommand class to being abstract. These are the base classes for the singular and plural command classes for example gmp.task and gmp.tasks.

daniele-mng (approved):

PR #4472, opened 2025-04-28
Change: file type preference styling

janlucasschwoerer-greenbone:
## What
Moved checkbox to enable credentials file upload and file input onto separate files

## Why
Originally the checkbox to enable uploads and the file input where on the same line.
This caused issues with actually being able to upload files due to the file input being too
small to click.
The checkbox to enable and the file input arenow on separate lines.

## References
https://jira.greenbone.net/browse/GEA-1036


bjoernricks (commented):
There is a linting warning `Warning:   17:8  warning  'Divider' is defined but never used  no-unused-vars` which should be resolved first. Thankfully it spotted an issue with the linting CI/CD job https://github.com/greenbone/gsa/pull/4475 which should fail in this case. To fix the linting warning you just need to remove the Divider import from the file.

bjoernricks (approved):

PR #4473, opened 2025-04-29
eslint rule i18n

daniele-mng:
This ESLint rule (`no-dynamic-i18n`) prevents the use of dynamic expressions inside internationalization (i18n) translation functions. It specifically targets:
- Non-string literals passed to the translation function `_()` 
- While allowing simple string literals and string concatenation with the `+` operator

## Why

When translatable strings contain dynamic expressions, they cannot be properly extracted by i18n tools. Issues:
- Translation extraction tools can only identify static strings, causing dynamic content to be missed
- Missing translations lead to untranslated text appearing in the UI

## Checklist

- [ ] Tests


src/web/pages/cves/Details.jsx
```diff
@@ -77,7 +77,7 @@ const CveDetails = ({entity}) => {
               .map(([name, title]) => (
                 <TableRow key={name}>
                   <TableData>{`${title}`}</TableData>
-                  <TableData>{_(entity[name])}</TableData>
+                  <TableData>{_l(entity[name])}</TableData>
```
bjoernricks, line 80: This change is actually not required here. But it is required that the value of `entity[name]`  is included in the locale file. And for extracting the value into the locale file it must be included somewhere in the javascript code as `_('Value of entity[name]')`. In this case this should be ensured by the [ENTITY_TYPES](https://github.com/greenbone/gsa/blob/main/src/gmp/utils/entitytype.ts#L67) variable already.

bjoernricks (approved):

PR #4496, opened 2025-04-30
Port List Improvements

bjoernricks:
## What

camelCase properties for all Port List and Port Range related classes and components.

## Why

Get a common coding style that uses camelCase.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/portlists/PortListComponent.tsx
+++ b/src/web/pages/portlists/PortListComponent.tsx
@@ -193,7 +193,7 @@ const PortListComponent = ({
     return response.data.id;
   };
 
-  const handleImportPortList = async (data: {xml_file: string}) => {
+  const handleImportPortList = async (data: {xmlFile: string}) => {
     handleInteraction();
     try {
       const response = await gmp.portlist.import(data);
--- a/src/web/pages/portlists/PortListComponent.tsx
+++ b/src/web/pages/portlists/PortListComponent.tsx
@@ -208,7 +208,7 @@ const PortListComponent = ({
     }
   };
 
-  const handleSavePortList = async (data: SavePortListData) => {
+  const handleSavePortList = async (data: SavePortListData<PortRange>) => {
     handleInteraction();
 
     if (isDefined(data.id)) {
--- a/src/web/pages/portlists/PortListComponent.tsx
+++ b/src/web/pages/portlists/PortListComponent.tsx
@@ -185,9 +185,9 @@ const PortListComponent = ({
 
   const handleSavePortRange = async (data: {
     id: string;
-    port_range_start: number;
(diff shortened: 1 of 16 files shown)


src/web/pages/portlists/PortListComponent.tsx
```diff
@@ -209,65 +208,62 @@ const PortListComponent = ({
...
-      } else if (!isDefined(data?.id) && isDefined(onCreateError)) {
-        return onCreateError(error);
+    if (isDefined(data.id)) {
+      // save existing port list
+      try {
+        const createdPromises = createdPortRanges.map(
+          async (range: PortRange) => {
+            // save temporary port ranges in the backend
+            const id = await handleSavePortRange({
+              id: range.id as string,
```
daniele-mng, line 221:
```suggestion
              id: range.id,
```

src/web/pages/portlists/PortRangesTable.tsx
```diff
@@ -45,16 +47,16 @@ const PortRangesTable: React.FC<PortRangesTableProps> = ({
       </TableHeader>
       <TableBody>
         {portRanges.map(range => (
-          <TableRow key={range.start + range.protocol_type}>
+          <TableRow key={`{range.start}-{range.protocolType}`}>
```
daniele-mng, line 50: Q: should it be interpolated?
```suggestion
          <TableRow key={`${range.start}-${range.protocolType}`}>
```

src/web/pages/portlists/PortListComponent.tsx
```diff
@@ -179,24 +179,23 @@ const PortListComponent = ({
   const handleDeletePortRange = async (range: PortRange) => {
     await gmp.portlist.deletePortRange({
       id: range.id as string,
```
daniele-mng, line 181:
```suggestion
      id: range.id,
```

src/web/pages/portlists/PortListComponent.tsx
daniele-mng, line 1: Line 345: download: handleDownload, and the others, I think there is a mismatch between being a promise and not expecting one.

src/web/pages/portlists/PortListComponent.tsx
bjoernricks: Thanks for catching this. it let me discover an issue with the meaning of this id. actually it is a port list id here and not a port range id!

src/web/pages/portlists/PortListComponent.tsx
bjoernricks: With the latest changes where the PortRange.id can be undefined now this code is actually required now.

daniele-mng (approved):

PR #4519, opened 2025-05-08
Add: cancellation of running CI jobs on new commit

henning:
## What

When workflows are triggered by a new commit, still running instances of the same workflow on the same git ref / branch will be cancelled to save resources.

## Why

We usually dont need to continue running old jobs when a new commit comes in that starts the same jobs again.

This saves us some running time and ressources.

## References

See DEVOPS-1538

## Checklist

- [ NA] Tests

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -3,6 +3,10 @@ name: Show changelog since last release
 on:
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true 
+
 jobs:
   changelog:
     name: Show changelog since last release
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -3,12 +3,6 @@ name: Show changelog since last release
 on:
   workflow_dispatch:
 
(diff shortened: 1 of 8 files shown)


.github/workflows/changelog.yml
```diff
@@ -3,6 +3,12 @@ name: Show changelog since last release
 on:
   workflow_dispatch:
 
+# cancel eventually still running jobs for the same branch and workflow
+# to save resources.
+#
+# NOTE:
+# This *can* lead to the effect that the first job that fails due to an
+# error introduced by a commit is *not* the run for the same commit but a later one.
```
bjoernricks, line 11: Sorry I don't get the meaning of the comment and it's consequences. Personally I also don't need these comments in the code and a note in the PR would be enough.

henning: thanks for the feedback.
i added these because I was worried that canbcelled runs can lkead to confusion about which commit really introduced a change that leads to an error. 
But  if the explanation is harder to understand than the corner case it should warn about, i can remove it.

henning: I have removed these comments

bjoernricks (approved):

PR #4524, opened 2025-05-09
Typescript commands

bjoernricks:
## What

Add type information for `AuthCommand`, `UserCommand` and `UserCommand` (and their dependent classes) which are available via `gmp.auth`, `gmp.user` and `gmp.users`.

## Why

Get fully typed Gmp class at the end.


src/web/pages/radius/RadiusPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+  // @ts-expect-error
+  const handleInteraction = () => dispatch(renewSessionTimeout(gmp)());
+
+  const loadRadiusAuthSettings = useCallback(async () => {
+    const response = await gmp.user.currentAuthSettings();
+    const {data: settings} = response;
+    // radius support is enabled in gvm-libs
+    const hasRadiusSupport = settings.has('method:radius_connect');
+    // @ts-expect-error
+    const {enabled, radiushost, radiuskey} = settings.get(
```
daniele-mng, line 69: Thought: We have `enabled, radiushost, radiuskey` and `enable, radiusHost, radiusKey`. Maybe we can have more specific names, since they are very similar and just the casing or a letter changing.

bjoernricks: If we want to change the data we get from `settings.has('method:radius_connect')` i would postpone that to another PR. Changing that data is more difficult because it's some data directly returned from the backend using some generic code.

bjoernricks: but i'll check if it can be improved in this PR first without changing the parsing code.

daniele-mng (approved):

PR #4560, opened 2025-06-03
Refactor Testing #2

bjoernricks:
## What

Move `web/components/testing` to `web/testing` to start a unified testing package

## Why

All UI testing code should be in `web/testing` in future. As a first step the code and modules from `web/components/testing` got moved to `web/testing` and the tests got adapted to this change.

## References
https://jira.greenbone.net/browse/GEA-1096

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/reportconfigs/__tests__/Dialog.test.jsx
+++ b/src/web/pages/reportconfigs/__tests__/Dialog.test.jsx
@@ -426,7 +426,7 @@ describe('New Report Config Dialog component tests', () => {
     });
 
     fireEvent.click(closeBtnElement);
-    const multiSelectMenuItems = await getSelectItemElementsForMultiSelect();
+    const multiSelectMenuItems = getSelectItemElementsForMultiSelect();
     fireEvent.click(multiSelectMenuItems[1]);
 
     const saveButton = screen.getDialogSaveButton();


src/web/testing/within.ts
```diff
@@ -0,0 +1,10 @@
+/* SPDX-FileCopyrightText: 2025 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import {getQueriesForElement} from '@testing-library/dom';
+import {allQueries} from 'web/testing/allQueries';
+
+export const within = (element: HTMLElement, queriesToBind = allQueries) =>
```
daniele-mng, line 9: Thought: since testing library already has already `within`, maybe we could find a custom name?

src/web/pages/reportconfigs/__tests__/Dialog.test.jsx
```diff
@@ -413,38 +398,38 @@ describe('New Report Config Dialog component tests', () => {
...
 
     const closeBtnElement = within(selectedItems[0]).getByRole('button', {
       hidden: true,
     });
 
-    await clickElement(closeBtnElement);
-    const multiSelectMenuItems = getSelectItemElementsForMultiSelect();
-    await clickElement(multiSelectMenuItems[1]);
+    fireEvent.click(closeBtnElement);
+    const multiSelectMenuItems = await getSelectItemElementsForMultiSelect();
```
daniele-mng: I don't think it's async.
```suggestion
    const multiSelectMenuItems = getSelectItemElementsForMultiSelect();
```

src/web/testing/within.ts
bjoernricks: IMHO we should keep the name because it overrides the within function from testing library and provides our custom queries too.

PR #4563, opened 2025-06-03
Refactor Testing #4

bjoernricks:
## What

Drop queries from render function and `web/utils/Testing`

## Why

We want to use screen everywhere.

## References
https://jira.greenbone.net/browse/GEA-1096

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/extras/__tests__/TrashCanPage.test.jsx
+++ b/src/web/pages/extras/__tests__/TrashCanPage.test.jsx
@@ -6,7 +6,8 @@
 import {describe, test, expect, testing} from '@gsa/testing';
 import Capabilities from 'gmp/capabilities/capabilities';
 import TrashcanPage from 'web/pages/extras/TrashCanPage';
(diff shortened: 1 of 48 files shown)


src/web/pages/extras/__tests__/TrashCanPage.test.jsx
```diff
@@ -39,39 +40,34 @@ const gmp = {
...
     });
 
-    const confirmButton = getByRole('button', {name: /Confirm/i});
-
-    await userEvent.click(confirmButton);
+    fireEvent.click(emptyTrashcanButton);
+    await wait();
+    expect(
+      screen.getByText('Are you sure you want to empty the trash?'),
+    ).toBeVisible();
```
daniele-mng, line 64: Out of the scope of the PR, but this should also work. Since `findByText` is a combo of `getBy` and `waitfor`.
```suggestion
    expect(
      await screen.findByText('Are you sure you want to empty the trash?'),
    ).toBeVisible();
```

PR #4581, opened 2025-06-10
Fix capabilities checks for displaying menu entries

bjoernricks:
## What

Fix capabilities checks for displaying menu entries

## Why

`scan_config` or `scan_config` is currently not a type that can be passed to the capabilities methods. It must be `config`, `configs`, `scanconfig` or `scanconfigs` at the moment. This is quite confusing and has to be improved in future.

Because I had the stunning idea to use *"better"* names for the entities in the UI then in the backend we introduced a translation mechanism between UI entity types and backend entity types. The backend entity types are used for the capabilities. Sadly we also  removed the underscore word separation for the UI entity names and this made the UI entity names even more confusing. As last resort of confusion, the capabilities methods accept both UI and backend entity types.

### DIFF
--- a/src/web/components/menu/__tests__/Menu.test.jsx
+++ b/src/web/components/menu/__tests__/Menu.test.jsx
@@ -6,6 +6,8 @@
 import React from 'react';
 import {describe, expect, test} from '@gsa/testing';
 import {rendererWith, screen} from 'web/testing';
+import Capabilities from 'gmp/capabilities/capabilities';
+import EverythingCapabilities from 'gmp/capabilities/everything';
 import Menu from 'web/components/menu/Menu';
 
 const renderMenuWith = ({capabilities, gmpSettings}) => {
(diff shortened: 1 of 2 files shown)


src/web/components/menu/__tests__/Menu.test.jsx
```diff
@@ -126,24 +116,20 @@ describe('Menu rendering', () => {
...
       'tags',
     ];
 
     renderMenuWith({
-      capabilities: {
-        mayAccess: feature => !configFeatures.includes(feature),
-        mayOp: () => true,
-        featureEnabled: () => true,
-      },
+      capabilities: new Capabilities(configFeatures),
```
a-h-abdelsalam: I may be wrong here since the test already passes but isn't the original test denying access to the features listed above ? Is it the same behavior passing `configFeatures` here ?

bjoernricks: Oh sorry, I wanted to rename the variable too. No the variable name `configFeatures` is misleading. It contains a list of capabilities. I'll change it now.

bjoernricks: Ok the list was completely unnecessary here and I've refactored the tests and merged two tests now

PR #4596, opened 2025-06-17
Improve Model Class

bjoernricks:
## What

Improve parsing and adapting data by improving the Model class

## Why

Allow passing known properties as arguments to constructors instead of adding them via `Object.defineProperty` inside of `Model.fromElement`.

Have a clear distinction between data from the backend and data that's available on a model. 

Currently the Model class still adds properties properties from the backend to a Model instance in `fromElement` to keep the backwards compatibility.

With this change applied we can differentiate between Models having an ID (instance of `Model`) and an optional ID (instance of `EntityModel`) in future by marking it required and throwing an exception in Model if the ID is missing in `fromElement`.

When all code is converted to TypeScript and we know all model properties, adding the extra properties from the backend can be removed in `Model.fromElement` too.

## References

https://jira.greenbone.net/browse/GEA-1122

## Checklist

- [x] Tests


src/gmp/models/__tests__/nvt.test.js
```diff
@@ -26,7 +26,7 @@ describe('nvt Model tests', () => {
     expect(nvt3.id).toEqual('1.2.3');
   });
 
-  test('should not allow to overwrite id', () => {
+  test.skip('should not allow to overwrite id', () => {
```
a-h-abdelsalam, line 29: Is skipping this intended ?

bjoernricks: Yes it is. Details are in my commit message at https://github.com/greenbone/gsa/pull/4596/commits/2cfa49bec956121ce4972e5afb4bfcfe7f3a3619

a-h-abdelsalam (approved):

PR #4609, opened 2025-06-20
Typescript for all SecInfo models

bjoernricks:
## What

Use TypeScript for all sec info related models like CVE, CPE, etc.

## Why

List and parse all know properties. Being able to validate the used properties.

## References

## Checklist

- [x] Tests


src/gmp/models/result.js
```diff
@@ -13,6 +12,14 @@ import {forEach, map} from 'gmp/utils/array';
...
 import {isEmpty} from 'gmp/utils/string';
 
+class CveResult {
+  constructor({name, epss}) {
+    this.name = name;
+    this.id = name;
+    this.epss = epss;
+  }
+}
+
```
daniele-mng: Could we use a function instead of a class, since it's just the constructor?

bjoernricks: Of course

daniele-mng (approved):

PR #4616, opened 2025-06-24
Fix creating Notes from Results and NVTs

bjoernricks:
## What

Fix creating Notes from Results and NVTs

## Why

The fields in the Note create dialog should be filled in automatically when creating a Note from a Result or NVT. This got broken while refactoring the NoteComponent into a function component.

## References

https://jira.greenbone.net/browse/GEA-1133

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/notes/Dialog.jsx
+++ b/src/web/pages/notes/Dialog.jsx
@@ -79,7 +79,7 @@ const NoteDialog = ({
     hosts,
     hosts_manual,
     id,
-    oid: isDefined(oid) ? oid : DEFAULT_OID_VALUE,
+    oid: oid ?? DEFAULT_OID_VALUE,
     port,
     port_manual,
     result_id,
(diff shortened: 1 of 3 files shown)


src/web/pages/notes/NoteComponent.jsx
```diff
@@ -108,28 +112,49 @@ const NoteComponent = ({
...
+      setId(id);
+      setNote(note);
+      setNvtName(nvt_name);
+      setOid(oid);
+      setPort(port);
+      setPortManual(port_manual);
+      setResultId(result_id);
+      setResultName(result_name);
+      setResultUuid(result_uuid);
+      setSeverity(severity);
```
daniele-mng: Maybe I am missing something, but it looks like we are setting state with the same state severity?

bjoernricks: Missed overriding severity indeed. Fixed that and didn't use de-structuring now. I think that's easier to read here.

src/web/pages/notes/Dialog.jsx
```diff
@@ -103,17 +103,21 @@ const NoteDialog = ({
         return (
           <>
             {state.fixed && isDefined(oid) && (
-              <FormGroup flex="column" title={_('NVT')}>
+              <FormGroup data-testid="group-nvt" flex="column" title={_('NVT')}>
```
daniele-mng: Q: we are using the same `data-testid='group-nvt` 3 times, and in the test we are only testing 1 of them, correct? Maybe we can change the name or remove the unsued ones?

bjoernricks: Yes the same `data-testid` is used three time, because they reference the element displaying the NVT information. But they are never rendered at the same time. Only one of them is rendered (or none).

bjoernricks: I slightly adjusted the ids now to be able to differentiate them. Of course the tests still need to be extended especially for the edit note case.

daniele-mng (approved):

PR #4623, opened 2025-06-26
Use TypeScript for TlsCertificate model

bjoernricks:
## What

Use TypeScript for TlsCertificate model

## References

## Checklist

- [x] Tests


src/web/pages/tlscertificates/Details.jsx
```diff
@@ -26,8 +26,12 @@ import withGmp from 'web/utils/withGmp';
 const TlsCertificateDetails = ({entity}) => {
   const [_] = useTranslation();
   return (
-    <Layout flex="column" grow="1">
-      <InfoTable>
+    <Layout
+      data-testid={`tls-certificate-details-${entity.id}`}
```
daniele-mng, line 30: Q: Maybe we could just have 1 `data-testid` since the look the same?

bjoernricks: Ups of course! The one for the InfoTable shouldn't be in the commit.

PR #4647, opened 2025-07-02
use Typescript for DefaultsPart and FilterPart

daniele-mng:
## What

use Typescript for DefaultsPart and FilterPart

## References

GEA-1144

## Checklist

- [x] Tests


src/web/pages/user-settings/FilterPart.tsx
```diff
@@ -0,0 +1,319 @@
...
+  UNSET_VALUE,
+  RenderSelectItemProps,
+} from 'web/utils/Render';
+
+export interface Filter {
+  id: string;
+  name: string;
+  filter_type: string;
+  [key: string]: unknown;
+}
```
bjoernricks: can we use the Filter model instead here?

PR #4648, opened 2025-07-02
Use TypeScript for Result model

bjoernricks:
## What

Use TypeScript for Result model

Also improve parsing of empty values, QoD values and the text elements. This also fixes an issue with Notes on Results introduced by the Note model TypeScript changes.

## References

## Checklist

- [x] Tests


src/gmp/models/result.ts
```diff
@@ -0,0 +1,376 @@
...
+  };
+  solution?: {
+    __text?: string;
+    _type?: string;
+  };
+  tags?: string;
+  type?: string;
+}
+
+type Compliance = string;
```
daniele-mng: Could we just use `string` directly instead of the alias?

bjoernricks: It's only a placeholder at the moment for the actual allowed values. But I couldn't figure out the actual values yet. If it's ok i would like to leave it as it is.

bjoernricks: I dropped the Compliance type now and we use string until we know the exact values.

PR #4651, opened 2025-07-02
Use TypeScript for Task model

bjoernricks:
## What

Use TypeScript for Task model

## References
Requires https://github.com/greenbone/gsa/pull/4649

https://jira.greenbone.net/browse/GEA-1098

## Checklist

- [x] Tests


src/gmp/models/task.ts
```diff
@@ -0,0 +1,528 @@
...
+                ? AUTO_DELETE_KEEP
+                : AUTO_DELETE_NO;
+            break;
+          case 'auto_delete_data':
+            const value = parseInt(pref.value);
+            copy.auto_delete_data =
+              value === 0
+                ? AUTO_DELETE_KEEP_DEFAULT_VALUE
+                : parseInt(pref.value);
+            break;
```
daniele-mng:
```suggestion
          case 'auto_delete_data': {
            const value = parseInt(pref.value);
            copy.auto_delete_data =
              value === 0
                ? AUTO_DELETE_KEEP_DEFAULT_VALUE
                : parseInt(pref.value);
            break;
          }
```

bjoernricks: I don't see any difference but I think I addressed the issue correctly now :-)

daniele-mng (approved):

PR #4657, opened 2025-07-04
Add: Add KDCs and realm display to CredentialDetails component

ozgen:
## What

Added support to show the realm and KDCs in the CredentialDetails component for Kerberos credentials.

## Why

These fields are important for users to view when working with Kerberos credentials.

## References

GEA-1129

## Checklist

- [ ] Tests

### DIFF
--- a/src/web/components/form/TagListDisplay.tsx
+++ b/src/web/components/form/TagListDisplay.tsx
@@ -1,0 +1,46 @@
+/* SPDX-FileCopyrightText: 2025 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import styled from 'styled-components';
+import tagStyles from 'web/components/form/tagStyles';
+
+interface TagListDisplayProps {
+  values: string[];
+  color?: keyof typeof tagStyles;
+}
+
+const Tag = styled.span<{bg: string; color: string}>`
+  background-color: ${({bg}) => bg};
+  color: ${({color}) => color};
+  padding: 2px 8px;
+  border-radius: 12px;
+  font-size: 0.85em;
+  font-weight: 500;
+  margin-right: 6px;
+`;
+
+const TagList = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+`;
+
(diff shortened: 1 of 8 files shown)


src/web/components/form/TagListDisplay.tsx
```diff
@@ -0,0 +1,46 @@
...
+  font-size: 0.85em;
+  font-weight: 500;
+  margin-right: 6px;
+`;
+
+const TagList = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+`;
```
daniele-mng, line 29: Suggestion: I think instead of using a `<span>` on L36, we could use `Layout` and in case adjust the margin in L22.

src/web/components/form/TagListDisplay.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const TagListDisplay = ({values, color = 'green'}: TagListDisplayProps) => {
+  const resolvedColor = tagStyles[color] ?? tagStyles.green;
+
+  return (
+    <TagList>
+      {values.map((val, idx) => (
+        <span key={idx} style={{display: 'flex', alignItems: 'center'}}>
```
daniele-mng: Suggestion: We should avoid using `idx` as key, better `val` or `val+idx`.

src/web/components/form/TagListDisplay.tsx
```diff
@@ -0,0 +1,47 @@
...
+  align-items: center;
+`;
+
+const TagListDisplay = ({values, color = 'green'}: TagListDisplayProps) => {
+  const resolvedColor = tagStyles[color] ?? tagStyles.green;
+
+  return (
+    <TagList>
+      {values.map((val, idx) => (
+        <Layout key={idx + val} style={{display: 'flex', alignItems: 'center'}}>
```
daniele-mng: Key should be a string. Also, I think we can use the prop `align` instead of `styles`.
```suggestion
        <Layout key={`${idx}-${val}`} style={{display: 'flex', alignItems: 'center'}}>
```

daniele-mng (approved):

PR #4666, opened 2025-07-11
add: user settings inline

daniele-mng:
## What

- Refactored all components in user-setting pages.
- Moved from editing settings inside the dialog to inline.
- Use gmp single setting cmd instead of batch, making changes more granular and manageable. 
- Errors are more granular and shown inside the row.

## Why

- User settings components were not in TS.
- User settings was using `user.saveSettings` for editing settings instead of single change cmd.

## References

GEA-1144
GEA-985

## Checklist

- [x] Tests

Screenshot from 2025-07-15 11-17-14


src/web/pages/user-settings/useSettingsSave.ts
```diff
@@ -0,0 +1,140 @@
...
+      settingName: string,
+      value: string | number,
+      setEditMode: (value: boolean) => void,
+    ): Promise<void> => {
+      try {
+        if (settingId) {
+          await gmp.user.saveSetting(settingId, String(value));
+        }
+        setEditMode(false);
+        loadSettings();
```
bjoernricks: Maybe just load the changed setting? or maybe just dispatch the new value to the store only?

src/web/pages/user-settings/UserSettingsPageHelpers.tsx
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Get the language name by its code
+ *
+ * @param {string} code - The language code to look up
+ * @returns {string|null} - The language name or null if not found
+ */
+export const getLangNameByCode = (code: string): string | null => {
+  const language = Languages[code];
+  return isDefined(language) ? `${language.name}` : null;
+};
```
bjoernricks:
```suggestion
export const getLangNameByCode = (code: string): string | undefined => {
  const language = Languages[code];
  return isDefined(language) ? String(language.name) : undefined;
};
```

PR #4671, opened 2025-07-14
Improve queryAllSelectElements by returning a HTMLSelectElement array

bjoernricks:
## What

Improve queryAllSelectElements by returning a HTMLSelectElement array

## Why

Allow for easier testing. We expect to query select elements here. Thus select elements should be returned.

### DIFF
--- a/src/web/testing/customQueries.ts
+++ b/src/web/testing/customQueries.ts
@@ -74,7 +74,7 @@ export const getSelectElement = (element?: HTMLElement) => {
  */
 export const queryAllSelectElements = (element?: HTMLElement) => {
   element = getElementOrReturnDocument(element);
-  return queryAllByTestId(element, 'form-select');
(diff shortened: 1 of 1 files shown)


src/web/testing/customQueries.ts
```diff
@@ -74,7 +74,7 @@ export const getSelectElement = (element?: HTMLElement) => {
  */
 export const queryAllSelectElements = (element?: HTMLElement) => {
   element = getElementOrReturnDocument(element);
-  return queryAllByTestId(element, 'form-select');
+  return queryAllByTestId(element, 'form-select') as HTMLSelectElement[];
```
daniele-mng, line 77: Q: Doesn't TS already return the correct type array?

bjoernricks: No sadly not. It assumes a `HTMLElement[]` here. That's the reason for my change.

PR #4682, opened 2025-07-16
fix: user setting browser language showing undefined

daniele-mng:
## What

fix: user setting browser language showing undefined

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/user-settings/helperFunctions.ts
+++ b/src/web/pages/user-settings/helperFunctions.ts
@@ -14,5 +14,5 @@ import Languages from 'web/utils/Languages';
  */
 export const getLangNameByCode = (code: string): string | undefined => {
   const language = Languages[code];
-  return isDefined(language) ? `${language.name}` : undefined;
+  return isDefined(language) ? String(language.name) : undefined;
 };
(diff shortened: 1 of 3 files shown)


src/web/utils/Languages.tsx
```diff
@@ -14,7 +14,7 @@ export {BROWSER_LANGUAGE};
 const Languages: Record<string, Language> = {
   ...GmpLanguages,
   [BROWSER_LANGUAGE]: {
-    name: String(_l('Browser Language')),
+    name: _l('Browser Language').toString(),
```
bjoernricks: Sadly both versions don't work. It needs to be converted to string where it is used actually.
```suggestion
    name: _l('Browser Language'),
```

src/web/pages/user-settings/helperFunctions.ts
```diff
@@ -13,6 +13,9 @@ import Languages from 'web/utils/Languages';
  * @returns {string|undefined} - The language name or null if not found
  */
 export const getLangNameByCode = (code: string): string | undefined => {
+  if (code === BROWSER_LANGUAGE) {
+    return Languages[BROWSER_LANGUAGE].name;
+  }
   const language = Languages[code];
   return isDefined(language) ? `${language.name}` : undefined;
```
bjoernricks: This should have worked already. Just replaced template literal with String now.
```suggestion
  const language = Languages[code];
  return isDefined(language) ? String(language.name) : undefined;
```

bjoernricks (approved):

PR #4694, opened 2025-07-17
fix:  typo in TagsPage component filter prop not populating filters

daniele-mng:
## What
fix:  typo in TagsPage component filter prop not populating filters

## Checklist

- [ ] Tests


bjoernricks (approved):
That bug was in there for at least over a year. Seems nobody uses the filters for tags ;-)

TypeScript to the rescue!

PR #4703, opened 2025-07-21
fix: ensure portlist page default filter is applied on initial load

daniele-mng:
## What

- Added the second value to the hook `usePageFilter` for loading the default filter
- Cleared and improved the useEffect logic for fetching.

## Why

 fix: ensure portlist page default filter is applied on initial load

## References

GEA-1166

## Checklist

- [x] Tests


src/web/pages/portlists/ListPage.jsx
```diff
@@ -145,23 +145,24 @@ const PortListsPage = () => {
...
+     */
+
+    const shouldFetch =
       isDefined(filter) &&
       !isLoadingFilter &&
-      !filter.equals(previousFilter)
-    ) {
+      (!isDefined(previousFilter) ||
+        filter.toFilterString() !== previousFilter?.toFilterString() ||
+        !filter.equals(previousFilter));
```
bjoernricks, line 160: I am a bit irritated about this change. It shouldn't behave differently then before.
Both `!isDefined(previousFilter)` and `filter.toFilterString() !== previousFilter?.toFilterString()` should be covered by `!filter.equals(previousFilter)` already.

bjoernricks: Note: a filter `foo=1 bar=2` is considered equal to `bar=2 foo=1` but their order of the filter terms differs. Therefore `filter.toFilterString` and `previousFilter.toFilterString` differs too. At the moment the `toFilterString` method is used to identify the data set loaded from the store. Therefore the data is not loaded from the backend but the store doesn't contain the data for the different identifier too.

bjoernricks (approved):

PR #4704, opened 2025-07-23
HOC Improvements

bjoernricks:
## What

Use TypeScript for DialogNotification and set displayName for all HOC wrapper components

## Why

Easier debugging.

## Checklist

- [x] Tests


src/web/components/chart/HostsTopologyChart.tsx
```diff
@@ -91,7 +92,8 @@ interface HostsTopologyChartProps {
   width: number;
   data: HostsTopologyChartData;
   svgRef: React.Ref<SVGSVGElement>;
-  _: (text: string, ...args: unknown[]) => string;
+  _: TranslateFunc;
+  i18n: I18n;
```
daniele-mng, line 96: Q: should it stay as `_`? `i18` is not used.
```suggestion
  _: I18n;
```

bjoernricks: Yes at least for now. it's required by the extractor currently https://github.com/greenbone/gsa/blob/main/i18next-parser.config.js#L6-L7

PR #4720, opened 2025-07-28
Improve useEntity hooks

bjoernricks:
## What

Improve useEntity hooks

* Split `useEntitySave` into `useEntitySave` and `useEntityCreate`
* Clarify and improve types passed to the useEntity hooks

## Why

While working on the PortList list page it turned out that the useEntity hooks didn't had the best or even an invalid interface.

## Checklist

- [x] Tests


src/web/entity/hooks/__tests__/useEntityCreate.ts
```diff
@@ -0,0 +1,62 @@
+/* SPDX-FileCopyrightText: 2025 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import {describe, test, expect, testing} from '@gsa/testing';
```
daniele-mng: File is missing the `.test.` suffix

PR #4753, opened 2025-08-04
Fix TrashCanPage

bjoernricks:
## What

Fix issues with the TrashCanPage component by refactoring it to TypeScript

## Why
The TrashCanPage got broken due to the Model refactoring and the differentiation between Audits and Tasks and Scan Configs and Policies. Before the refactoring of the Models audit responses could be parsed as Task instances and Policy models could be parsed as ScanConfig instances. This is not possible anymore and without having type information at the TrashCan page we didn't notice. Therefore the TrashCanPage and all its used code code refactored to TypeScript. The TrashCan command now returns audits, tasks, scan configs and policies parsed correctly already.

## References

https://jira.greenbone.net/browse/GEA-1194

## Checklist

- [x] Tests


src/web/pages/trashcan/TrashCanPage.tsx
```diff
@@ -0,0 +1,379 @@
...
+  // load the data on initial render
+  useEffect(() => {
+    loadTrash();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const tableProps: TrashCanTableProps = {
+    links: false,
+    onEntityRestore: handleRestore,
+    onEntityDelete: handleDelete,
+    actionsComponent: TrashActions,
```
daniele-mng, line 168: I think there is a mismatch in TS with onRestore/onDelete, the handlers are promises but the definitions aren't.

bjoernricks: You are right. I think it's also already mixed up everywhere :-/ During conversion to typescript it's not always obvious. The component may also just expect a void function but actually gets a `Promise<Something>`.

PR #4758, opened 2025-08-06
fix: scan config and policy dialog error

daniele-mng:
## What

- On opening of the scancofig and policy edit dialog, it would display error.

## Why

- Object structure was missing key/value

## References

GEA-1195

## Checklist

- [x] Tests

### DIFF
--- a/src/gmp/models/policy.ts
+++ b/src/gmp/models/policy.ts
@@ -73,7 +73,7 @@ class Policy extends Model {
   readonly family_list: PolicyFamily[];
   readonly nvts?: PolicyNvts;
   readonly predefined?: boolean;
-  readonly preferences?: PolicyPreferences;
+  readonly preferences: PolicyPreferences;
   readonly scanner?: Model;
   readonly audits: Model[];
 
@@ -83,7 +83,7 @@ class Policy extends Model {
     family_list = [],
     nvts,
     predefined,
-    preferences,
+    preferences = {nvt: [], scanner: []},
     scanner,
     audits = [],
     ...properties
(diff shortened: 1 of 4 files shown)


src/gmp/models/scanconfig.ts
```diff
@@ -26,6 +26,7 @@ export interface ScanConfigPreferenceElement {
   nvt?: {
     name?: string;
     _oid?: string;
+    oid?: string;
```
bjoernricks: Are you sure about that change? It's also not used in your changes as far as I can see.

src/gmp/models/policy.ts
```diff
@@ -170,11 +170,11 @@ class Policy extends Model {
           nvtPreferences.push(pref);
         }
       });
-      ret.preferences = {
-        scanner: scannerPreferences,
-        nvt: nvtPreferences,
-      };
     }
+    ret.preferences = {
```
bjoernricks, line 174: Could you adjust `PolicyProperties` too and also the Policy Model to make the preferences required (always available) now?

src/gmp/models/scanconfig.ts
```diff
@@ -242,12 +243,13 @@ class ScanConfig extends Model {
           nvtPreferences.push(pref);
         }
       });
-      ret.preferences = {
-        scanner: scannerPreferences,
-        nvt: nvtPreferences,
-      };
     }
 
+    ret.preferences = {
```
bjoernricks, line 247: Same for the ScanConfigProperties and the ScanConfig Model. preferences can be changed to be always available now.

PR #4763, opened 2025-08-12
Delta report sorting

daniele-mng:
## Checklist

- [ ] Tests


src/web/pages/reports/DeltaDetailsPage.tsx
```diff
@@ -0,0 +1,669 @@
...
+interface DeltaReportDetailsWrapperProps {
+  defaultFilter?: Filter;
+  reportFilter?: Filter;
+  entity?: Report;
+}
+
+interface DeltaReportDetailsProps {
+  deltaReportId: string;
+  entity: Report;
+  entityError?: unknown;
```
bjoernricks: I am currently defining errors as Error. Probably it can be a [Rejection instance](https://github.com/greenbone/gsa/blob/main/src/gmp/http/rejection.ts#L9) but the code handles that internally. Personally I would love if we could add some error type that is either `type ErrorType = Error | Rejection` or something like `interface ErrorType { message: string }`.
```suggestion
  entityError?: Error;
```

src/gmp/commands/reports.ts
```diff
@@ -0,0 +1,231 @@
...
+import DefaultTransform from 'gmp/http/transform/default';
+import {XmlResponseData} from 'gmp/http/transform/fastxml';
+import logger from 'gmp/log';
+import Filter, {ALL_FILTER} from 'gmp/models/filter';
+import Report, {ReportElement} from 'gmp/models/report';
+import {isDefined} from 'gmp/utils/identity';
+
+interface ReportCommandImportParams {
+  task_id: string;
+  in_assets?: number;
```
bjoernricks:
```suggestion
  in_assets?: YesNo;
```

src/gmp/commands/reports.ts
```diff
@@ -0,0 +1,231 @@
...
+  xml_file: string;
```
bjoernricks: For now while we don't have form validation in the frontend I would make it optional. When calling the method, undefined is allowed and only the backend checks if it is required.
```suggestion
  xml_file?: string;
```

src/gmp/commands/reports.ts
```diff
@@ -0,0 +1,236 @@
...
+  filter?: string;
+}
+
+interface ReportCommandAlertParams {
+  alert_id: string;
+  report_id: string;
+  filter: string;
+}
+
+interface ReportCommandGetDeltaParams {
```
bjoernricks: The interface seems to be unused

src/web/pages/reports/DeltaDetailsPage.tsx
```diff
@@ -0,0 +1,591 @@
...
+      updateFilter: f => {
+        return dispatch(setPageFilter(getReportPageName(params.id), f));
+      },
+    }),
+    [dispatch, gmp, params.id],
+  );
+};
+
+const useReportState = params => {
+  const {id, deltaid} = params || {};
```
bjoernricks: Maybe
```suggestion
const useReportState = ({id, deltaid}: {id?: string, deltaid?: string} = {}) => {
```

PR #4765, opened 2025-08-13
fix: Allow editing dashboard tab name when only one tab exist and convert to TypeScript

daniele-mng:
## What
 
fix: Allow editing dashboard tab name when only one tab exist and convert to TypeScript

## Why

It was not possible to rename a dashboard tab if there is only 1.
## References

GEA-1191

## Checklist

- [ ] Tests


src/web/pages/start/__tests__/Page.test.jsx
bjoernricks, line 1: Could you rename the test file and change it to TS too? Should be `StartPage.test.tsx`

PR #4834, opened 2025-09-16
Add: Agent Group Management Page

daniele-mng:
## References

GEA-975

## Checklist

- [ ] Tests

### DIFF
--- a/src/gmp/commands/trashcan.ts
+++ b/src/gmp/commands/trashcan.ts
@@ -6,7 +6,7 @@
 import HttpCommand from 'gmp/commands/http';
 import Response from 'gmp/http/response';
 import {XmlMeta, XmlResponseData} from 'gmp/http/transform/fastxml';
-import AgentGroup from 'gmp/models/agent-groups';
+import AgentGroup from 'gmp/models/agentgroup';
 import Alert from 'gmp/models/alert';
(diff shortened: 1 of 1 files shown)


src/gmp/commands/agent-groups.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    agentData.interval_in_seconds = intervalInSeconds;
+    agentData.miss_until_inactive = missUntilInactive;
+
+    log.debug('Prepared data for both create calls', {
+      agentGroupData,
+      agentData,
+    });
+
+    const [agentGroupResult, agentResult] = await Promise.allSettled([
```
daniele-mng: Thought:

-  reconsidering my approach, it makes more sense to create the agent group first since it defines the entity and then modify. 

- Modifying agents should be heavier/costly, so doing it second sequentially is safer.

- Using `Promise.allSettled` in parallel could lead to partial success and makes rollback harder.

- Sequential execution makes it easier to rollback the first call if the second fails.

- Error handling should remain similar to indicate exactly which step failed.



Suggestion:

- Run create_agent_group first, then modify_agent. Should consider rolling back first if second call fails. 

Todo:

- Refactor create() and save() to follow sequential execution.

bjoernricks (approved):

PR #4836, opened 2025-09-18
Entity, Capability and GMP API Types

bjoernricks:
## What

Introduce TS types for entity, capability and GMP API.

This PR also fixes the displayed permission names in the different permission dialogs.

## Why

Define which values are allowed and expected when an entity type is used. In GSA we have the EntityType that is used in every model class as some kind of an identifier. When making a request we need to use the ApiType to create a corresponding command for example `modify_task`. Therefore the EntityType can always be converted to an ApiType.
For checking permissions we currently allow both entity and api types for backwards compatibility. This may change in future when all code is converted to TypeScript.

In summary we have three new types:

* EntityType, which should always be used in GSA when we need a type of an entity
* ApiType, which should only be used within command class methods for creating the http request
* Capability, which lists all possible permissions of gvmd and may be used for `Capability.mayOp` only
* CapabilityEntityType, which is a combination of EntityType and ApiType. It can be used for all Capabilities methods other then `mayOp`. These methods are `mayAccess`, `mayCreate`, `mayDelete`, `mayEdit` and `mayClone`. It may be removed in future and be replaced with EntityType only when all code is converted to TypeScript.

## Checklist

- [x] Tests


src/gmp/utils/entitytype.ts
```diff
@@ -105,40 +112,86 @@ const ENTITY_TYPES = {
...
   portlist: 'port_list',
   portrange: 'port_range',
   reportconfig: 'report_config',
   reportformat: 'report_format',
+  scanconfig: 'config',
   tlscertificate: 'tls_certificate',
   vulnerability: 'vuln',
 } as const;
 
+export const API_TYPES = [
```
daniele-mng, line 145: Nice!

PR #4839, opened 2025-09-24
Scanner Improvements

bjoernricks:
## What

Cleanup and improve Scanner API

* Parse scanner port as number
* Display `Agent Controller` instead of `Agent Scanner` for the scanner type `7`
* Use TypeScript for ScannerCommand and ScannersCommand
* Cleanup ScannerCommand save method to behave accordingly to modify_scanner GMP command.
* Parse scanner type as string for easier usage in (selection) input fields

## Why

TypeScript everything and streamlined Scanner API.

## References

Requires latest gsad from main branch to work correctly. See https://github.com/greenbone/gsad/pull/231

https://jira.greenbone.net/browse/GEA-1282

## Checklist

- [x] Tests


src/gmp/models/scanner.ts
```diff
@@ -97,17 +97,18 @@ interface ScannerProperties extends ModelProperties {
...
 }
 
-export const OPENVAS_SCANNER_TYPE = 2;
-export const CVE_SCANNER_TYPE = 3;
-export const GREENBONE_SENSOR_SCANNER_TYPE = 5;
-export const OPENVASD_SCANNER_TYPE = 6;
-export const AGENT_CONTROLLER_SCANNER_TYPE = 7;
-export const OPENVASD_SENSOR_SCANNER_TYPE = 8;
-export const AGENT_CONTROLLER_SENSOR_SCANNER_TYPE = 9;
+export const OPENVAS_SCANNER_TYPE = '2';
```
ozgen, line 105: Since `scanner_type` is now a string instead of a number, we should check places where it’s used in filters.
I found one of the places: https://github.com/greenbone/gsa/blob/scanner-improvements/src/web/pages/performance/PerformancePage.tsx#L53

The value goes directly into the filter. Because gsad just inserts it into filter="%s", sending '7' instead of 7 might break the request.

ozgen (approved):

src/gmp/models/scanner.ts
bjoernricks: This should still work as expected. Before the GREENBONE_SENSOR_SCANNER_TYPE was implicitly converted to string. The actual code was
```js
const SENSOR_SCANNER_FILTER = Filter.fromString(
  `type=${String(GREENBONE_SENSOR_SCANNER_TYPE)}`,
);
```
It's still producing the same filter without the quotes.

PR #4847, opened 2025-09-29
Fix parsing and displaying certificate information

bjoernricks:
## What

Fix parsing and displaying certificate information

## Why

Parts of the certificate information of credentials were not parsed at all.

## Checklist

- [x] Tests

### DIFF
--- a/src/gmp/models/credential.ts
+++ b/src/gmp/models/credential.ts
@@ -20,10 +20,22 @@ type CredentialType =
   | typeof USERNAME_PASSWORD_CREDENTIAL_TYPE
   | typeof USERNAME_SSH_KEY_CREDENTIAL_TYPE;
 
+type TimeStatus =
+  | typeof CERTIFICATE_STATUS_EXPIRED
+  | typeof CERTIFICATE_STATUS_INACTIVE
+  | typeof CERTIFICATE_STATUS_VALID
+  | typeof CERTIFICATE_STATUS_UNKNOWN;
+
 interface CredentialElement extends ModelElement {
   certificate_info?: {
     activation_time?: string;
     expiration_time?: string;
+    issuer?: string;
+    md5_fingerprint?: string;
+    sha256_fingerprint?: string;
+    subject?: string;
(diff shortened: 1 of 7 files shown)


src/web/components/certinfo/CertInfo.tsx
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import React from 'react';
+import {Date} from 'gmp/models/date';
```
ozgen: `Date` confuses whether it is a native date or not, make it clear:

`import type { Date as GmpDate } from 'gmp/models/date';
`

src/gmp/models/credential.ts
```diff
@@ -195,6 +215,14 @@ class Credential extends Model {
       ret.certificate_info = {
         activationTime: parseDate(element.certificate_info.activation_time),
         expirationTime: parseDate(element.certificate_info.expiration_time),
+        issuer: element.certificate_info.issuer,
+        subject: element.certificate_info.subject,
+        serial: element.certificate_info.serial,
+        md5Fingerprint: element.certificate_info.md5_fingerprint,
+        sha256Fingerprint: element.certificate_info.sha256_fingerprint,
+        timeStatus: isDefined(element.certificate_info.time_status)
+          ? (element.certificate_info.time_status as TimeStatus)
```
ozgen: Instead of casting blindly, you could run the string through a small function that only accepts the values you actually know. 

like: 

```
const toTimeStatus = (s?: string): TimeStatus | undefined => {
  if (!s) return undefined;

  switch (s) {
    case CERTIFICATE_STATUS_VALID:
    case CERTIFICATE_STATUS_EXPIRED:
    case CERTIFICATE_STATUS_INACTIVE:
      return s;
    default:
      return CERTIFICATE_STATUS_UNKNOWN;
  }
};
```

If you think that it is not needed, that's also okay for me

ozgen (approved):

PR #4850, opened 2025-09-29
Convert EntityLink to TypeScript

bjoernricks:
## What

Convert EntityLink to TypeScript

## Why

Use TypeScript for EntityLink component and its tests.

## References

https://jira.greenbone.net/browse/GEA-1094

## Checklist

- [x] Tests


src/web/entity/Link.tsx
```diff
@@ -3,19 +3,24 @@
...
+
+const EntityLink = ({entity, textOnly, ...props}: EntityLinkProps) => {
   const [_] = useTranslation();
-  const {id, name, userCapabilities, deleted} = entity;
-  const type = normalizeType(getEntityType(entity));
+  const capabilities = useCapabilities();
+  const {id, name, userCapabilities} = entity;
+  const type = normalizeType(getEntityType(entity)) as EntityType;
 
   if (entity.isInTrash()) {
```
ozgen, line 25: Maybe, not all models implement trash?
Should we need to add another check?


`if (typeof e.isInTrash === 'function' && e.isInTrash()) {
`

ozgen (approved):

src/web/entity/Link.tsx
bjoernricks: At the moment all models implement this method with https://github.com/greenbone/gsa/blob/main/src/gmp/models/entitymodel.ts#L141

PR #4862, opened 2025-10-02
Allow to create Agent Controllers

bjoernricks:
## What

* Convert all scanner components to TypeScript
* Show create icon at scanner pages if the user is allowed to create scanners
* Add creating of agent controllers
* Allow for filtering by scanner type in the powerfilter dialog
* Only show scanners if the user has permissions to access them and the feature is enabled

## Why

Allow to create Agent Controllers

## References

https://jira.greenbone.net/browse/GEA-1282

## Checklist

- [x] Tests


src/web/pages/scanners/ScannerDialog.tsx
```diff
@@ -0,0 +1,245 @@
...
+                name="port"
+                value={state.port}
+                onChange={(value: number) =>
+                  onScannerPortChange && onScannerPortChange(value)
+                }
+              />
+            </FormGroup>
+
+            {showCertificateField && (
+              <FormGroup title={_('Certificate')}>
```
ozgen: The “New Scanner” dialog currently allows setting:

- Name
- Comment
- Type (Agent Controller)
- Host
- Port
- CA Certificate

However, to connect successfully to an Agent Controller (openvasd) using HTTPS/mTLS, we also need a client certificate and a client private key.

At the moment, it’s possible to create the scanner with only the CA certificate, but the connection will always fail.
A better approach would be to let users create credentials that store these certificates and then assign those credentials to the scanner during creation or update.

src/web/pages/scanners/ScannerDialog.tsx
```diff
@@ -0,0 +1,245 @@
...
+                  items={renderSelectItems(
+                    credentials as RenderSelectItemProps[],
+                  )}
+                  name="credentialId"
+                  value={credentialId}
+                  onChange={(value: string) =>
+                    onCredentialChange && onCredentialChange(value)
+                  }
+                />
+                <NewIcon
```
ozgen, line 207: When I edit an OpenVASD Sensor scanner and click “New Credential”, the Type field is empty.
I can’t choose any credential type.
Could you please check why the types are not showing?

ozgen: Also, the Credential field should be available for the Agent Controller type as well, since it also needs certificates for connection.

src/web/pages/scanners/ScannerDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+            <FormGroup title={_('Port')}>
+              <NumberField
+                disabled={scannerInUse}
+                name="port"
+                value={state.port}
+                onChange={(value: number) =>
+                  onScannerPortChange && onScannerPortChange(value)
+                }
+              />
+            </FormGroup>
```
ozgen, line 191: There seems to be an inconsistency in how we handle the CA certificate. The credentials UI allows CA, server, and client certificates, and the DB entries look partially correct; however, gvmd only persists server and client certificates on the credential. During connection creation, the CA is read from the scanner table rather than the credential which is not stored. To keep things consistent, the CA certificate should be configured in the scanner dialog and passed directly during connection setup.

PR #4869, opened 2025-10-09
Made edited scan configs keep their family trend.

jhelmold:
## What
Edited scan configs always switched their family trend to static when
they were saved. Now they keep their family trend.

## Why
This was a bug.

## References
GEA-1281

## Checklist

### DIFF
--- a/src/gmp/commands/scanconfigs.js
+++ b/src/gmp/commands/scanconfigs.js
@@ -101,7 +101,7 @@ export class ScanConfigCommand extends EntityCommand {
       id,
       comment,
       name,
-      familyTrend,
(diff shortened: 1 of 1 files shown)


src/gmp/commands/scanconfigs.js
```diff
@@ -93,6 +101,7 @@ export class ScanConfigCommand extends EntityCommand {
       id,
       comment,
       name,
+      familyTrend,
```
bjoernricks: This change should allow to use the existing trend parameter in gsad
```suggestion
      trend: familyTrend,
```

PR #4878, opened 2025-10-13
Improve FileField

bjoernricks:
## What

Improve FileField for issues found while working on client certificate credential.

## Why

* Use correct props for the FileField
* Add a label for the clear button
* Ensure the clear button is shown
* Use undefined for not having a value (internally it's null)
* Update the onChange handler to expect undefined too (it's used when the field in cleaned)

## Checklist

- [x] Tests

### DIFF
--- a/public/locales/gsa-de.json
+++ b/public/locales/gsa-de.json
@@ -397,6 +397,7 @@
   "Choose from common scheduling options above, or enable advanced mode for custom expressions.": "Wählen Sie aus den oben genannten Standardplanungsoptionen oder aktivieren Sie den erweiterten Modus für benutzerdefinierte Ausdrücke.",
   "Choose Tag": "Tag auswählen",
   "Clear": "Zurücksetzen",
+  "Clear input": "Eingabe löschen",
   "Client Certificate": "Benutzerzertifikat",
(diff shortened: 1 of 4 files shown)


src/web/components/form/FileField.tsx
```diff
@@ -4,30 +4,33 @@
  */
 
 import {useCallback} from 'react';
+import {FileInputProps} from '@mantine/core';
```
daniele-mng:
```suggestion
import type {FileInputProps} from '@mantine/core';
```

src/web/pages/credentials/CredentialDialog.tsx
```diff
@@ -193,13 +193,13 @@ const CredentialDialog = ({
...
+      setPublicKey(file);
     } catch (error) {
       setError((error as Error).message);
     }
   };
 
-  const handleClientCertificateChange = async (file: File | null) => {
+  const handleClientCertificateChange = async (file: File | undefined) => {
     try {
       if (hasValue(file)) {
```
daniele-mng: Question: it could be still be truthy and fail, instead of checking if it `hasValue` could we use? Eventually also valid for the other handlers.
```suggestion
    if (file instanceof File) {
```

PR #4929, opened 2025-11-04
Hide new agent task if flag is deactivated

daniele-mng:
## What

Pages >  Tasks > new task, shows agent even in flag is deactivated

## Checklist

- [ ] Tests


src/web/pages/tasks/TaskComponent.tsx
```diff
@@ -83,6 +83,8 @@ import ModifyTaskWizard, {
   type ModifyTaskWizardData,
 } from 'web/wizard/ModifyTaskWizard';
 import TaskWizard from 'web/wizard/TaskWizard';
+import Features from 'gmp/capabilities/features';
```
bjoernricks:
```suggestion
```

PR #4933, opened 2025-11-05
Fix: Added host and location to ticket model.

jhelmold:
## What
Added host and location to the ticket (gmp) model.

## Why
The values for the host and for the location were not (could not be) shown in the ticket details view. This was a bug.

## References
GEA-1265

## Checklist


src/gmp/models/ticket.ts
```diff
@@ -101,6 +104,8 @@ class Ticket extends Model {
   readonly openTime?: Date;
   readonly result?: Model;
   readonly report?: Model;
+  readonly host?: string;
```
daniele-mng, line 107: Maybe a good idea to update the unit test for these 2 new properties.

PR #4940, opened 2025-11-07
Expect a gmp export function for useEntityDownload hook

bjoernricks:
## What

Expect a gmp functions for useEntity hooks

## Why

Improve type checking by expecting a gmp functions as first argument to all useEntity hooks instead of the entity type name. This allows validating if the functions really exist and have correct types.

## References

Requires https://github.com/greenbone/gsa/pull/4942

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/roles/RoleComponent.tsx
+++ b/src/web/pages/roles/RoleComponent.tsx
@@ -54,9 +54,9 @@ interface DeletePermissionData {
 interface RoleComponentRenderProps {
   create: (role?: Role) => void;
   edit: (role: Role) => void;
-  clone: (role: Role) => void;
+  clone: (role: Role) => Promise<void>;
   delete: (role: Role) => Promise<void>;
-  download: (role: Role) => void;
+  download: (role: Role) => Promise<void>;
 }
 
 interface RoleComponentProps {


src/web/pages/roles/RoleComponent.tsx
```diff
@@ -237,54 +246,83 @@ const RoleComponent = ({
...
+          onClose={handleCloseRoleDialog}
+          onCreatePermission={handleCreatePermission}
+          onCreateSuperPermission={handleCreateSuperPermission}
+          onDeletePermission={handleDeletePermission}
+          onErrorClose={handleErrorClose}
+          onSave={async d => {
+            const promise = isDefined(d.id)
+              ? handleEntitySave({
+                  ...d,
+                  id: d.id as string,
```
daniele-mng, line 317:
```suggestion
                  id: d.id,
```

src/web/pages/roles/RoleComponent.tsx
```diff
@@ -237,54 +246,83 @@ const RoleComponent = ({
...
-              onSave={d => {
-                const promise = isDefined(d.id) ? save(d) : create(d);
-                return promise.then(() => closeRoleDialog());
-              }}
-            />
-          )}
-        </>
+    <>
+      {children({
+        clone: handleEntityClone,
```
daniele-mng, line 289: I think there might be a mismatch from what we are passing and what is expected in the interface. Here it's returning a `Promise` but the interface expects `return void`. Most likely also download on line 292.

src/web/pages/scanners/ScannerComponent.tsx
```diff
@@ -376,8 +385,15 @@ const ScannerComponent = ({
           onNewCredentialClick={openCredentialsDialog}
           onSave={async d => {
             const promise = isDefined(d.id)
-              ? handleScannerSave(d)
-              : handleScannerCreate(d);
+              ? handleScannerSave({
+                  ...d,
+                  id: d.id as string,
```
daniele-mng, line 390:
```suggestion
                  id: d.id,
```

src/web/pages/scanners/ScannerDialog.tsx
```diff
@@ -131,7 +131,7 @@ const ScannerDialog = ({
     },
   );
   const [userChangePort, setUserChangedPort] = useState<boolean>(false);
```
daniele-mng: Nitpick, not relevant to the PR: `userChangePort` and `setUserChange d Port` mismatch.

src/web/pages/roles/RoleComponent.tsx
bjoernricks: I changed this to return `Promise<void>` now.

src/web/pages/roles/RoleComponent.tsx
bjoernricks: Actually this is required because d.id has type `string | undefined` but actually isn't in the case of a save.

src/web/pages/scanners/ScannerComponent.tsx
bjoernricks: Same here. This is required because d.id has type `string | undefined` but actually is defined in the case of a save.

src/web/pages/scanners/ScannerDialog.tsx
bjoernricks: Adjusted the code to fix this typo.

daniele-mng (approved):

PR #4954, opened 2025-11-13
Add tests for Http class

bjoernricks:
## What

Add tests for Http class
## Why

Increase coverage for Http class and ensure it works as expected.

## Checklist

- [x] Tests


src/gmp/http/__tests__/http.test.ts
```diff
@@ -3,64 +3,206 @@
...
-}));
+// global.XMLHttpRequest = testing.fn(() => ({
+//   open: testing.fn(),
+//   send: testing.fn(),
+//   setRequestHeader: testing.fn(),
+//   status: 0,
+//   responseText: '',
+//   onreadystatechange: null,
+//   readyState: 0,
+// }));
```
daniele-mng: We want to keep this? Maybe comment explanation

bjoernricks: Sorry that was a leftover. Thanks for catching it!

PR #4967, opened 2025-11-19
Use TypeScript for Target Components

bjoernricks:
## What

Use TypeScript for Target Components

## Why

Convert all Target components to TypeScript, extend and provide new tests, convert module and component names to match our naming rules.

## References

https://jira.greenbone.net/browse/GEA-1376

## Checklist

- [x] Tests


src/web/pages/targets/TargetDetailsPage.tsx
```diff
@@ -28,71 +26,53 @@ import TableRow from 'web/components/table/TableRow';
...
+interface EntityProps {
+  entity: Target;
+}
 
-const Details = ({entity, ...props}) => {
-  const [_] = useTranslation();
+interface TargetDetailsPageProps {
+  entity: Target;
+  isLoading?: boolean;
+  links?: boolean;
```
daniele-mng: I don't think we use this prop
```suggestion
 ```

bjoernricks: removed

PR #4975, opened 2025-11-21
Rename Container Task to Import Task

bjoernricks:
## What
Rename Container Task to Import Task

## Why

With the introduction of the `Container Image Task` it will be confusing to the user to still have the `Container Task`. In our context the `Container Task` has nothing to do with OCI or docker containers. It's just a task that can import exported XML reports.

## References

https://jira.greenbone.net/browse/GEA-1387

## Checklist

- [x] Tests


src/gmp/models/audit.ts
```diff
@@ -212,7 +212,10 @@ class Audit extends Model {
     return this.alterable !== NO_VALUE;
   }
 
-  isContainer() {
+  /**
+   * Returns true if the audit is an import audit (formerly known as container task)
+   */
+  isImport() {
```
ozgen, line 218: We might be overlooking something in the audit logic. It currently only checks whether a target exists, but we may also need to include both the container image and the agent group

ozgen (approved):

PR #4985, opened 2025-11-27
Add package build workflow

henning:
**What**:
Add package build workflow

**Why**:

To enable automatic triggering of the package and qm deploy pipeline

**How**:

As announced/discussed in chat.

Tested with an example repo while developing the pipeline.

See DEVOPS-1736

### DIFF
--- a/.github/workflows/package-build-on-release.yml
+++ b/.github/workflows/package-build-on-release.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: "ubuntu-latest"
     steps:
       - name: Trigger development-packaging workflow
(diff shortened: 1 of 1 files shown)


.github/workflows/package-build-on-release.yml
```diff
@@ -0,0 +1,27 @@
...
+  id-token: write
+  pull-requests: write
+
+jobs:
+  call_packaging_workflow:
+    name: Call Packaging workflow
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Trigger development-packaging workflow
+        uses: greenbone/actions/trigger-workflow@487dab5354ace3dd575c8cf100a3e522c1dd3a79
```
bjoernricks:
```suggestion
        uses: greenbone/actions/trigger-workflow@v3
```

PR #5008, opened 2025-12-10
Fix Tagging of specific Entity/Resource Types

bjoernricks:
## What

It was not possible to tag Hosts, Audits, Audit Reports, Cert Bund Advisories, DFN Cert Advisories, Policies and Operating Systems.

## Why

Fix tagging these entity types and improve the codebase.

## References

https://jira.greenbone.net/browse/GEA-1425

## Checklist

- [x] Tests


src/web/entities/BulkTags.tsx
```diff
@@ -159,17 +155,16 @@ const BulkTags = <TEntity extends Model>({
...
-          active: YES_VALUE,
+          active: true,
           comment,
           filter: loadedFilter,
-          id,
-          name,
-          resource_ids: tagEntitiesIds,
-          resource_type: entitiesType,
-          resources_action: 'add',
+          id: id as string,
```
daniele-mng, line 163: Thought: Is `id` and `name` always present that we are casting it to overwrite the `undefined`? Valid also for L215, and some other files in the PR.

src/web/pages/tags/TagDialog.tsx
```diff
@@ -0,0 +1,287 @@
...
+      setIsLoading(true);
+      gmp.resourcenames
+        .getAll({resourceType: type})
+        .then(response => {
+          const data: ResourceOption[] = response.data;
+          let id: string | undefined = resourceIdText;
+          if (isEmpty(id)) {
+            id = undefined;
+          } else {
+            const idPresent = data.findIndex(res => res.id === id) !== -1;
```
daniele-mng:
```suggestion
            const idPresent = data.some(res => res.id === id);
```

src/web/entities/BulkTags.tsx
bjoernricks: Sadly not at the moment as `id` and `name` are allowed to be `undefined`. That's on my list for refactoring.

Some background, all entities are required to have a name and id besides `Filter` (and `PortRange`). Filter is a bit different then the others. A `Filter` can be an object created from a filter string (doesn't have an id or name) or from a response of the backend (has an id and name) like the other entities. Technically in the UI both filters behave the same but are different because only the latter exists in the backend. So we need to come up with a solution to distinguish both Filter "types" and then we can make id and name required.

bjoernricks: Thus casting id and name to string is correct in 99% of all cases. If we get an entity object from the backend, id and name are always available and defined.

daniele-mng (approved):

PR #5245, opened 2026-04-20
Change: Mark existing scanners for disabled features as disabled.

a-h-abdelsalam:
## What
Mark existing scanners for disabled features as disabled.

## Why
`get_scanners` does not hide existing scanners for disabled features. They are now marked unusable if the corresponding feature is disabled.

## References
GEA-1722

### DIFF
--- a/public/locales/gsa-de.json
+++ b/public/locales/gsa-de.json
@@ -691,6 +691,7 @@
   "Diffuse": "Diffus",
   "Directory depth for indexer": "Verzeichnistiefe für Indexer",
   "Disable Tag": "Tag deaktivieren",
+  "Disabled": "Deaktiviert",
   "Distance": "Entfernung",
   "Do not automatically delete reports": "Berichte nicht automatisch löschen",
   "Do not change": "Nicht ändern",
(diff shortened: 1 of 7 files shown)


src/gmp/models/entity-model.ts
```diff
@@ -68,6 +68,7 @@ export interface EntityModelProperties extends BaseModelProperties {
   userCapabilities?: Capabilities;
   userTags?: UserTag[];
   writable?: YesNo;
+  disabled?: boolean;
```
bjoernricks: Is disabled a generic property usable for all entity types (aka. resource types in gvmd)? If not it should be added to the specific models for the types having a disabled field only.

a-h-abdelsalam: Yes you are right, it is only used for scanners. The issue is I need to use `disabled` in `EntityNameTableData` and it raises a property does not exist error. `deprecated` is handled by simply suppressing the error. Do you see a better approach ?

bjoernricks: EntityNameTableData is a [generic](https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-types) therefore (if the using module is a TypeScript file) you should be able to specify the exact kind of model instance. for example`EntityNameTableData<Scanner>`

a-h-abdelsalam: OK thanks for the hint. Will check

bjoernricks (approved):

PR #5250, opened 2026-04-23
Change SesssionObserver into function component and use TypeScript

bjoernricks:
## What

Change SesssionObserver into function component and use TypeScript

## Why

Refactor the SessionObserver class component into a function component and also use TypeScript for the module.

Allows to use session hook instead of having to use the redux store. We want to get rid of the redux store at the end.


src/web/components/observer/SessionObserver.tsx
```diff
@@ -0,0 +1,87 @@
...
+    if (isDefined(timer)) {
+      return;
+    }
+
+    const timeout = sessionTimeout.diff(date()) + DELAY;
+
+    if (timeout > 0) {
+      const timer = globalThis.setTimeout(handlePing, timeout);
+      // currently the type of timer is NodeJS.Timeout in node and number in browser
+      setTimer(timer as unknown as number);
```
daniele-mng:
```suggestion
       const timer: ReturnType<typeof setTimeout> = globalThis.setTimeout(
        handlePing,
        timeout,
      );
      setTimer(timer);
```

src/web/components/observer/SessionObserver.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const handlePing = useCallback(async () => {
+    log.debug('pinging server to check session');
+    setTimer(undefined);
+    try {
+      await gmp.user.ping();
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (e) {
+      /* empty */
+    }
```
daniele-mng: Maybe:
```suggestion
    } catch (e) {
      log.debug('session ping failed', e);
    }
```
or if there is any reason why we should not log the `e` we could just remove it.

bjoernricks: The ping is just for getting a 401 in case the session has expired. which will redirect to the login page automatically and therefore we can ignore all error responses here.

src/web/components/observer/SessionObserver.tsx
bjoernricks: Thanks! I extracted this into a `type Timer =  ReturnType<typeof setTimeout>` :-)

daniele-mng (approved):

PR #5257, opened 2026-04-30
Session and Login

bjoernricks:
## What

* Don't logout on browser page reload
* Use token and username from session
* Remove username data from redux store

## Why

Move session related data into the session as only source of truth

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/tasks/TaskRow.tsx
+++ b/src/web/pages/tasks/TaskRow.tsx
@@ -89,7 +89,7 @@ const TaskRow = ({
   ...props
 }: TaskRowProps) => {
   const [_] = useTranslation();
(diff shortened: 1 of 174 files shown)


src/web/pages/tasks/TaskRow.tsx
```diff
@@ -89,7 +89,7 @@ const TaskRow = ({
   ...props
 }: TaskRowProps) => {
   const [_] = useTranslation();
-  const [username] = useUserName();
+  const username = useUserName() as string;
```
daniele-mng, line 92:
```suggestion
  const username = useUserName();
```

daniele-mng (approved):

PR #5330, opened 2026-06-04
Add: add dark mode theming

Chri6s:
## What
- Adds a global dark mode color scheme using the native Mantine `ThemeSwitch` inside the header.
- Aligns assets (SVGs, conditional animations like :hover) to the selected colorTheme.
- Adds safety checks to handle CSS variables during `d3-color` chart computations to prevent crashes.

## Why
- Better user experience.
## References
- N/A

## Checklist

- [x] Tests (aligned the ones that were not compliant to this change)


Chri6s:
@bjoernricks Can I request a review for this PR as there are new UI changes which could cause wrong coloring in specific views/icons etc.?

PR #5358, opened 2026-06-18
add: new os icon for n/a

daniele-mng:
## What

- Add new icon if there are information about the OS, but no icon for it.
- Add fallback for know OS, but no match because of version.

## References

GEA-1816

## Checklist

- [x] Tests


src/web/utils/__tests__/Os.test.tsx
bjoernricks, line 1: Should this file be in `web/utils/__tests__`?

bjoernricks (approved):

PR #5369, opened 2026-06-25
Typescript Overrides

bjoernricks:
## What

Use TypeScript for all Override components

## Why

With converting the components to TypeScript we can ensure the passed props are as expected.

## References

https://jira.greenbone.net/browse/GEA-1840

## Checklist

- [x] Tests


src/web/pages/overrides/OverrideTableRow.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const OverrideTableRow = ({
+  actionsComponent: ActionsComponent = OverrideTableActions,
+  entity,
+  onToggleDetailsClick,
+  ...props
+}: OverrideTableRowProps) => {
+  const [_] = useTranslation();
+
+  const renderSeverity = (severity: number): string => {
```
ozgen: the cast seems to hide that severity may be `undefined`, while `renderSeverity` already handles that case. Could we type the parameter as optional and pass `entity.severity` directly?

I mean like:

```
const renderSeverity = (severity?: number): string => {
  if (isDefined(severity)) {
    if (severity <= LOG_VALUE) {
      return translateRiskFactor(extraRiskFactor(severity));
    }

    return `> ${severityValue(severity - 0.1)}`;
  }

  return _('Any');
};

<TableData>{renderSeverity(entity.severity)}</TableData>
```

ozgen (dismissed):

ozgen (approved):

PR #5372, opened 2026-06-25
Use TypeScript for Group commands

bjoernricks:
## What

Use TypeScript for Group commands
## Why

Convert Group commands to TypeScript and add tests for both to ensure their behavior.

## References

https://jira.greenbone.net/browse/GEA-1098

## Checklist

- [x] Tests


src/gmp/commands/group.ts
```diff
@@ -0,0 +1,76 @@
...
+    comment = '',
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    grant_full,
+    users = [],
+  }: GroupCommandCreateParams) {
+    const data = {
+      cmd: 'create_group',
+      name,
+      comment,
+      grant_full: parseYesNo(grant_full),
```
ozgen: `grant_full` is optional, but `parseYesNo(undefined)` returns ` 0`.

Could this disable `grant_full` when the value is not provided? Maybe we should only convert it when it is defined.

ozgen (approved):

robindittmar (approved):

PR #5376, opened 2026-06-30
feat: add agent support bundle download

ozgen:
## What

- Add support bundle download for agents.
- Add a download action to the Agents page.
- Add tests for the new action and download flow.

## Why

- Users need an easy way to download agent support bundles.
- The bundles are required for troubleshooting and support.

## References

GEA-1834

## Checklist

- [x] Tests


src/web/pages/agents/AgentListPage.tsx
```diff
@@ -188,6 +192,35 @@ const AgentListPage = () => {
     bulkDisableUpdateToLatest,
   ]);
 
+  const downloadSupportBundleMutation = useDownloadAgentSupportBundle({
+    onError: error => {
+      showError(error);
```
bjoernricks: Could be simplified
```suggestion
  const downloadSupportBundleMutation = useDownloadAgentSupportBundle({
    onError: showError,
```

src/web/pages/agents/AgentListPage.tsx
```diff
@@ -188,6 +192,34 @@ const AgentListPage = () => {
...
+    },
+  });
+
+  const handleDownloadSupportBundle = (agent: Agent) => {
+    if (!agent.id) {
+      showError(new Error(_('Agent ID is required.')));
+      return;
+    }
+
+    downloadSupportBundleMutation.mutate(agent.id, {
```
bjoernricks: Just as a side note/information: Personally i just cast agent.id to string here and don't add a check. Because id being possibly undefined here isn't the case in the reality and the type is just based on some quirk in the requirements for the model class implementation (actually there is only one single model that may have an empty id, the filter). This quirk will also be resolved some day in the future :-)

bjoernricks (dismissed):

src/web/pages/agents/AgentListPage.tsx
ozgen: that makes sense 👍🏼

src/web/pages/agents/AgentListPage.tsx
```diff
@@ -188,6 +192,33 @@ const AgentListPage = () => {
...
+  const downloadSupportBundleMutation = useDownloadAgentSupportBundle({
+    onError: showError,
+  });
+
+  const handleDownloadSupportBundle = (agent: Agent) => {
+    if (!agent.id) {
+      showError(new Error(_('Agent ID is required.')));
+      return;
+    }
+
```
bjoernricks: With that change you can remove this check here :-)

bjoernricks (dismissed):

bjoernricks (approved):

PR #5400, opened 2026-07-03
Fix Delta Reports

bjoernricks:
## What

Fix Delta Reports

Parse and use the results from the `<get_report/>` response again for delta reports. This was removed while refactoring the report details for the multi requests.

## Why

Loading the results via `<get_results/>` doesn't work for delta reports. The returned data doesn't contain the delta information and won't display the expected data.

## References

https://jira.greenbone.net/browse/GEA-1903
https://jira.greenbone.net/browse/GEA-1856
https://jira.greenbone.net/browse/GEA-1904
https://jira.greenbone.net/browse/GEA-1906

## Checklist

- [ ] Tests


src/gmp/models/report/parser.ts
```diff
@@ -540,3 +554,53 @@ export const parseClosedCvesFromEndpoint = (
...
+  // due to the XML conversion to JS the result element can be a single element or an array
+  const length = isDefined(resultsElement.result)
+    ? isArray(resultsElement.result)
+      ? resultsElement.result.length
+      : 1
+    : 0;
+
+  return new CollectionCounts({
+    first: resultsElement._start,
+    rows: resultCountElement?.filtered,
```
ozgen: Should `rows` use `resultsElement._max` here?

`rows` is the page size, while `filtered` is the total number of matching results. Using `filtered` may cause the Previous button to stay disabled on later pages.

```suggestion
    rows: resultsElement._max,
```

ozgen (dismissed):

src/gmp/models/report/parser.ts
bjoernricks: I need to take a deeper look into that. I just restored the old code and it was already like this.

ozgen (approved):

PR #5401, opened 2026-07-09
Refactor create Scan Config Create Dialog

bjoernricks:
## What

Refactor create Scan Config Create Dialog to drop the radio selection and only show the existing scan configs as template to derive from.

Fix displaying the NVT preference counts when editing a Scan Config family.

Convert all corresponding components to typescript to ensure that all data is passed correctly and has the correct shape.

## Why

The current Scan Config Create Dialog shows a select input with the available scanners when the lowest radio is selected. This is a leftover from OSPd times and doesn't work at all.

## References

https://jira.greenbone.net/browse/GEA-693
https://jira.greenbone.net/browse/GEA-1918

## Checklist

- [x] Tests


public/locales/gsa-en.json
```diff
@@ -2086,6 +2082,7 @@
   "Tasks with most High Results per Host": "Tasks with most High Results per Host",
   "TCP": "TCP",
   "TCP Port Count": "TCP Port Count",
+  "Template": "",
```
ozgen: Do we need to add explanation instead of an empty string?

src/web/pages/scanconfigs/ScanConfigCreateDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+import SaveDialog from 'web/components/dialog/SaveDialog';
+import Select from 'web/components/form/Select';
+import TextField from 'web/components/form/TextField';
+import useTranslation from 'web/hooks/useTranslation';
+import PropTypes from 'web/utils/PropTypes';
+import {type RenderSelectItemProps, renderSelectItems} from 'web/utils/Render';
+
+interface ScanConfigCreateData {
+  baseScanConfig: string;
+  customScanConfig?: string;
```
ozgen: Is `customScanConfig` still needed? After the refactor, the dialog seems to only use `baseScanConfig`

src/web/pages/scanconfigs/ScanConfigCreateDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+              onChange={onValueChange}
+            />
+
+            <Select
+              disabled={scanConfigs.length === 0}
+              grow="1"
+              items={renderSelectItems(scanConfigs as RenderSelectItemProps[])}
+              label={_('Template')}
+              name="baseScanConfig"
+              value={state.customScanConfig}
```
ozgen: This looks wrong. It should probably be:

`value={state.baseScanConfig}`

public/locales/gsa-en.json
bjoernricks: Our translations files use the English term as the key and the translation as a value. if the key doesn't exist or the value is undefined, null or empty the translation mechanism falls back to using the key as the translation value. So this is not an issue but you spotted a bug in our extraction setup because it should just add the key as the value for the `en` translation.

src/web/pages/scanconfigs/ScanConfigCreateDialog.tsx
bjoernricks: removed the property

ozgen (approved):

PR #5406, opened 2026-07-10
Fix displaying errors when importing of a report fails

bjoernricks:
## What

Fix displaying errors when importing of a report fails

## Why

Don't use the onError handler and let the dialog handle the error cases
instead to show possible errors of the request in the dialog.

## References
https://jira.greenbone.net/browse/GEA-1751

### DIFF
--- a/src/web/pages/tasks/TaskComponent.tsx
+++ b/src/web/pages/tasks/TaskComponent.tsx
@@ -940,13 +940,10 @@ const TaskComponent = ({
   };
 
   const handleReportImport = (data: ReportImportDialogData) => {
-    return (
-      gmp.report
-        // @ts-expect-error
-        .import(data)
-        .then(onReportImported, onReportImportError)
-        .then(() => closeReportImportDialog())
-    );
+    return gmp.report
+      .import(data)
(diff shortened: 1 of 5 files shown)


src/web/pages/tasks/TaskComponent.tsx
```diff
@@ -940,13 +940,10 @@ const TaskComponent = ({
...
-    return (
-      gmp.report
-        // @ts-expect-error
-        .import(data)
-        .then(onReportImported, onReportImportError)
-        .then(() => closeReportImportDialog())
-    );
+    return gmp.report
+      .import(data)
+      .then(onReportImported, onReportImportError)
```
ozgen, line 945: Should we remove `onReportImportError` here as well?
it handles the error, so `SaveDialog` cannot display it, and the dialog may close even when the import fails.

bjoernricks: I am not sure about that. Currently onReportImportError is always undefined so we get the desired behavior. From an independent component perspective a user of the component might want to handle import errors by himself. Thus i think it should stay as it is.

ozgen (approved):

PR #5409, opened 2026-07-14
Immutable Filter Class

bjoernricks:
## What

Make Filter class in GSA immutable

## Why

Don't allow to change the internal data of a Filter instance. This avoids issues with frameworks depending on equality checks for comparing the identity. For example react decides to re-render if the identity has changed only. Also redux heavily relies on the identity.

## References

https://jira.greenbone.net/browse/GEA-848

## Checklist

- [x] Tests

### DIFF
--- a/src/gmp/models/filter.ts
+++ b/src/gmp/models/filter.ts
@@ -99,7 +99,7 @@ class Filter extends EntityModel {
 
   readonly alerts: Model[];
   readonly filter_type?: string;
-  readonly terms: FilterTerm[];
+  private terms: FilterTerm[];
 
   constructor({
     _type,
@@ -359,7 +359,7 @@ class Filter extends EntityModel {
     const index = this._getIndex(key);
     const hasKey = index !== -1;
     if (hasKey) {
-      this.terms.splice(index, 1);
+      this.terms = this.terms.filter((_, i) => i !== index);
       this._resetFilterId(); // filter has changed
     }
     return hasKey;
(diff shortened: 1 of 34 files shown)


src/gmp/models/filter.ts
```diff
@@ -310,6 +310,61 @@ class Filter extends EntityModel {
...
+   *
+   * @param key Filter term key to remove
+   *
+   * @return True if a FilterTerm with this key has been removed from this Filter.
+   */
+  private _delete(key: string) {
+    const index = this._getIndex(key);
+    const hasKey = index !== -1;
+    if (hasKey) {
+      this.terms.splice(index, 1);
```
daniele-mng: since we want to make it immutable and keep the original untouched, we can use `filter()` instead of `splice()`.
```suggestion
      this.terms = this.terms.filter((_, i) => !== index)
```

src/gmp/models/filter.ts
```diff
@@ -716,78 +763,79 @@ class Filter extends EntityModel {
...
   mergeKeywords(filter: Filter | undefined | null) {
-    if (hasValue(filter)) {
-      this._resetFilterId();
-
-      this._mergeNewKeywords(filter);
+    if (!hasValue(filter)) {
+      return this;
     }
-    return this;
+    const copy = this._copy();
```
daniele-mng, line 807: Would it make sense to check if new keywords are added before making a copy and resetting? Also valid for `mergeExtraKeywords`.

src/gmp/models/filter.ts
bjoernricks: Nice catch! I made the FilterTerm and FilterTerm[] also immutable now.

src/gmp/models/filter.ts
bjoernricks: Improved this behavior with the latest changed.

daniele-mng (approved):

PR #5424, opened 2026-07-15
Set scanner_id to NO_VALUE_ID.

jhelmold:
## What
Set the scanner_id to NO_VALUE_ID to avoid a misleading error message when saving a non-alterable container task.

## Why
This fixes a bug.

## References
GEA-1605

## Checklist


src/gmp/commands/task.ts
```diff
@@ -525,7 +525,7 @@ class TaskCommand extends EntityCommand<Task, TaskElement> {
     name,
     ociImageTargetId,
     registryAllowInsecure,
-    scannerId = NO_VALUE_ID,
+    scannerId,
```
bjoernricks: If scannerId is not read at all can we remove the parameter here?

bjoernricks (approved):

PR #5427, opened 2026-07-16
Fix: Show all filters in filter selection.

jhelmold:
## What
Show all filters instead of only the first 10 in selection of filters in the report view.

## Why
This was a small bug.

## References
GEA-1930

## Checklist

### DIFF
--- a/src/gmp/models/filter.ts
+++ b/src/gmp/models/filter.ts
@@ -1014,9 +1014,6 @@ export const REPORT_FORMATS_FILTER_FILTER =
   Filter.fromString('type=report_format');
 export const REPORTS_FILTER_FILTER = Filter.fromString('type=report');
 export const RESULTS_FILTER_FILTER = Filter.fromString('type=result');
-export const RESULTS_FILTER_ALL_FILTER = Filter.fromString(
-  'type=result rows=-1',
-);
 export const ROLES_FILTER_FILTER = Filter.fromString('type=role');
 export const SCANCONFIGS_FILTER_FILTER = Filter.fromString('type=config');
 export const SCANNERS_FILTER_FILTER = Filter.fromString('type=scanner');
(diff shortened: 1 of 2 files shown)


src/gmp/models/filter.ts
```diff
@@ -1010,6 +1010,9 @@ export const REPORT_FORMATS_FILTER_FILTER =
   Filter.fromString('type=report_format');
 export const REPORTS_FILTER_FILTER = Filter.fromString('type=report');
 export const RESULTS_FILTER_FILTER = Filter.fromString('type=result');
+export const RESULTS_FILTER_ALL_FILTER = Filter.fromString(
+  'type=result rows=-1',
+);
```
bjoernricks:
```suggestion
```

src/web/hooks/use-query/reports.ts
```diff
@@ -52,7 +52,7 @@ export const useGetResultsFilters = () => {
   return useGetEntities<Filter>({
     gmpMethod: gmp.filters.get.bind(gmp.filters),
     queryId: 'get_filters',
-    filter: RESULTS_FILTER_FILTER,
+    filter: RESULTS_FILTER_ALL_FILTER,
```
bjoernricks: Every filter has a `.all` method that sets the rows accordingly. There is not need to provide a new constant in the filter model for that.
```suggestion
    filter: RESULTS_FILTER_FILTER.all(),
```

bjoernricks: IMHO the `useGetResultsFilter` function should be renamed to `getAllResults` too

jhelmold: IMHO the name of the function useGetResultsFilters should stay as it is or be renamed to getAllResultsFilters, because it gets the filters for results and not the results itself. What do you think?

bjoernricks: You are right. we can leave it as it is.

bjoernricks (approved):

PR #5428, opened 2026-07-17
Extract FilterTerms handling into a FilterTerms class

bjoernricks:
## What

Extract FilterTerms handling into a FilterTerms class

## Why

This will allow to re-use most code for different filter classes.
## References

https://jira.greenbone.net/browse/GEA-1933

## Checklist

- [x] Tests

### DIFF
--- a/src/gmp/models/filter.ts
+++ b/src/gmp/models/filter.ts
@@ -155,7 +155,7 @@ const SORT_ORDER_DESC = 'sort-reverse';
  *
  * @param filterString  Filter representation as a string
  *
- * @return Array of parsed FilterTerms
+ * @returns Array of parsed FilterTerms
  */
 const parseFilterTermsFromString = (
   filterString: string | undefined,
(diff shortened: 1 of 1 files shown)


src/gmp/models/filter.ts
```diff
@@ -193,49 +193,14 @@ const parseFilterTermsFromString = (
...
-      userTags,
-      writable,
-    });
-
-    this.alerts = alerts;
-    this.filter_type = filter_type;
+  }: {
+    terms?: FilterTerm[];
+  } = {}) {
     this.terms = terms;
```
daniele-mng, line 204: Should we consider the following change to have always a fresh copy? Not sure if there are any risks of mutations.
```suggestion
    this.terms = [...terms];
```

src/gmp/models/filter.ts
```diff
@@ -944,29 +796,171 @@ class Filter extends EntityModel implements FilterType {
...
+  }
+
+  get length() {
+    return this.filterTerms.length;
+  }
+
+  /**
+   * Create a new Filter from the passed FilterTerms
+   */
+  private _fromFilterTerms(filterTerms: FilterTerms, keepId = false): Filter {
```
daniele-mng: Consideration: This is only used in `_delegate`, maybe we can just inline it as it's simple?

src/gmp/models/filter.ts
bjoernricks: No i don't think so here. The terms array is owned by the FilterTerms class after the creation. It just some kind of collection class.

src/gmp/models/filter.ts
```diff
@@ -977,9 +971,131 @@ class Filter extends EntityModel implements FilterType {
...
   static fromTerm(...term: FilterTerm[]) {
-    return new Filter({
+    const filterTerms = new FilterTerms({
       terms: term,
     });
+    return new Filter({
+      terms: [...filterTerms.getAllTerms()],
+    });
+  }
+
```
daniele-mng, line 971: Question: Are they copying the same data? Maybe it can be simplified?
```suggestion
  static fromTerm(...term: FilterTerm[]) {
    return new Filter({
    terms:term
    });
  }

```

bjoernricks: Yes you are right. Fixed.

src/gmp/models/filter.ts
bjoernricks: I guess the method was used before in some other place which i already dropped or refactored. Removed.

daniele-mng (approved):

PR #5434, opened 2026-07-20
feat: add Japanese (ja) language support

doracpphp:
## What

Add Japanese (`ja`) as a selectable UI language in GSA.

* Register `ja` in the language list with the native name "日本語".
* Add `ja` to the `i18next-cli` extract locales.
* Add `public/locales/gsa-ja.json` with all 2,373 keys translated.
* Update language count assertions in `languages.test.ts`.

**Verification:**
* Verified via `npm run type-check`, `eslint`, `prettier`, and `vitest`.
* Verified that running `npm run i18n-extract` reproduces the updated locale catalogue byte-for-byte.

## Why

8.x and later: The entire codebase was rewritten, causing Japanese support to be lost.

This PR restores Japanese support using the latest full key set. Terminology follows standard modern IT localization practices in Japanese (e.g., katakana with long vowel marks such as サーバー, ユーザー, フィルター), while carefully preserving all i18next placeholders (including `{{- name}}`).

## Checklist

- [x] Tests


bjoernricks:
@doracpphp sorry for the late reply. are you willing to maintain the JP translations? i think updating them once a month or once a quarter should be enough.

bjoernricks (approved):

doracpphp:
> @doracpphp sorry for the late reply. are you willing to maintain the JP translations? i think updating them once a month or once a quarter should be enough.

@bjoernricks 
No worries at all!

Yes, I'm happy to maintain the JP translations.

PR #5435, opened 2026-07-20
Split Filter into two classes

bjoernricks:
## What

Strip Filter model class into a model class and remove the additional parsing into another class.

## Why

All model classes should require an ID in future. In reality the ID is always present if a model is created from an XML element.

Currently, this is being prevented by the extra parsing stuff `fromString`, `fromTerm` and `fromResponeElement` which don't provide an ID. Therefore these methods were moved into the new `BaseFilter` class and the `Filter` model class only forwards the calls to this new class for backwards compatibility. In future these methods should be removed from `Filter` completely.

## References

https://jira.greenbone.net/browse/GEA-1933

## Checklist

- [x] Tests

### DIFF
--- a/src/gmp/models/filter.ts
+++ b/src/gmp/models/filter.ts
@@ -158,6 +158,8 @@ class Filter extends EntityModel implements FilterType {
   /**
    * Create a new Filter from the passed response element.
    *
+   * @deprecated Use `BaseFilter.fromResponseElement` instead.
+   *
    * @param element Response element to parse properties from.
    *
    * @returns A new Filter model instance.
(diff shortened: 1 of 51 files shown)


src/gmp/models/filter.ts
```diff
@@ -270,16 +112,13 @@ class Filter extends EntityModel implements FilterType {
    *
    * @returns A new Filter if the FilterTerms are different, otherwise the current Filter.
    */
-  private _delegate(terms: FilterTerms, keepId = false): Filter {
+  private _delegate(terms: FilterTerms, keepId = false) {
     return terms === this.filterTerms
       ? this
-      : new Filter({
+      : new BaseFilter({
```
daniele-mng, line 118: Passing `BaseFilter` instead of `Filter`, are we going to lose methods like alerts, `filter_type`, comment etc?

bjoernricks: Yes but these are only for interest when looking at the filter model (pages that call `<get_filters>`). It wasn't available before already. These changes just make that obvious.

daniele-mng (approved):

PR #5441, opened 2026-07-21
fix: trashcan agent groups delete/restore button actionable

daniele-mng:
## What

- On page trashcan delete and restore agent groups was not possible.

## Why

- Missing props

## References

GEA-1875

## Checklist

- [ ] Tests


bjoernricks (commented):
Can we make the props explicit? By just forwarding a props object it is difficult to understand in future which props are actually passed.

bjoernricks (approved):

PR #5451, opened 2026-07-28
GEA-1606

daniele-mng:
## What

- Refactor from JavaScript to Typescript.
- Replace `Redux` with `TanstackQuery`.
- Fix table sorting.
- Fix UI bug when batch deleting.
- Rename all files and components with specific name
- Improve test coverage

## References

GEA-1606

## Checklist

- [x] Tests


src/web/entity/QueryEntityComponent.tsx
```diff
@@ -0,0 +1,133 @@
...
+      TEntity,
+      TCreateData,
+      TCreateResponse,
+      TSaveData,
+      TSaveResponse
+    >,
+  ) => React.ReactNode;
+  createMutation: QueryMutation<TCreateVariables, TCreateResponse>;
+  saveMutation: QueryMutation<TSaveVariables, TSaveResponse>;
+  deleteMutation: QueryMutation<TDeleteVariables, unknown>;
```
bjoernricks: I think this should be void
```suggestion
  deleteMutation: QueryMutation<TDeleteVariables, void>;
```

src/web/entity/QueryEntityComponent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  cloneMutation: QueryMutation<TCloneVariables, unknown>;
```
bjoernricks: Should be `EntityActionData` from `gmp/commands/entity` and is just `{id: string}`
```suggestion
  cloneMutation: QueryMutation<TCloneVariables, EntityActionData>;
```

src/web/entity/QueryEntityComponent.tsx
```diff
@@ -0,0 +1,133 @@
...
+    return id;
+  };
+
+  const requireSaveId = (data: TSaveData): string => {
+    const id = getSaveId ? getSaveId(data) : (data as {id?: string}).id;
+    if (!isDefined(id)) {
+      throw new Error(`${entityName} ID is required for saving`);
+    }
+    return id;
+  };
```
bjoernricks: I am not sure if we need this anymore. With my model changes, all entities of a page are having an id now. It should not happen that there is no ID if an entity is cloned or deleted. Just for the differentiation between create and save we check if the data has an id or not. We could possibly split that too by providing a different save handler for the dialogs instead of a save handler that decides between save and create.

src/web/hooks/use-query/users.ts
```diff
@@ -0,0 +1,185 @@
...
+  enabled?: boolean;
+}
+
+interface UseGetUserParams {
+  id: string;
+  refetchInterval?: RefetchIntervalFn<User>;
+}
+
+interface UseUserMutationCallbacks {
+  onSuccess?: (response: unknown) => void;
```
bjoernricks: I think we can use `EntityActionData | void` here. It could be possible that mutation commands currently return a created entity or an ActionResult but both are `EntityActionData` which is actually `{id: string}`. IMHO it would be best to change all commands to return the same data for the different mutations and not return different things. And i guess (if i don't miss something) we just don't need the return data or we just need the id for querying the updated entity.

src/web/pages/users/__tests__/UserComponent.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+    _allow: '0',
+  },
+});
+
+const authSettings = new Settings();
+authSettings.set('method:ldap_connect', {enabled: false});
+authSettings.set('method:radius_connect', {enabled: false});
+
+const groups = [User.fromElement({_id: 'group1', name: 'Group 1'}) as Model];
+const roles = [User.fromElement({_id: 'role1', name: 'Admin'}) as Model];
```
bjoernricks: Shouldn't this be
```suggestion
const roles = [Role.fromElement({_id: 'role1', name: 'Admin'})];
```

src/web/pages/users/__tests__/UserComponent.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+    __text: '192.168.1.1',
+    _allow: '0',
+  },
+});
+
+const authSettings = new Settings();
+authSettings.set('method:ldap_connect', {enabled: false});
+authSettings.set('method:radius_connect', {enabled: false});
+
+const groups = [User.fromElement({_id: 'group1', name: 'Group 1'}) as Model];
```
bjoernricks: Shouldn't this be
```suggestion
const groups = [Group.fromElement({_id: 'group1', name: 'Group 1'})];
```

src/web/pages/users/__tests__/UserFilterDialog.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  changeInputValue,
+  fireEvent,
+  rendererWith,
+  screen,
+  waitFor,
+} from 'web/testing';
+import QueryFilter from 'gmp/models/filter/query-filter';
+import UserFilterDialog from 'web/pages/users/UserFilterDialog';
+
+const newFilter = new QueryFilter({
```
bjoernricks: To be correct a filter returned from the filter command methods is always a Filter model and not a QueryFilter, but it doesn't make a difference here in the tests.
```suggestion
const newFilter = new Filter({
```

src/web/pages/users/__tests__/UserFilterDialog.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  id: 'new-filter-id',
+  name: 'New Filter',
+});
+
+const gmp = {
+  filter: {
+    create: testing.fn().mockResolvedValue({
+      data: {id: 'new-filter-id', name: 'New Filter'},
```
bjoernricks:
```suggestion
      data: newFilter,
```

src/web/pages/users/__tests__/UsersHeader.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import {describe, expect, test, testing} from '@gsa/testing';
+import {rendererWith, screen} from 'web/testing';
+import UsersHeader from 'web/pages/users/UsersHeader';
+import SelectionType from 'web/utils/SelectionType';
+
+describe('users Header', () => {
```
bjoernricks:
```suggestion
describe('UsersHeader', () => {
```

src/web/pages/users/UsersTableHeader.tsx
bjoernricks, line 1: Maybe the component should be name `UsersTableHeader`?

src/web/pages/users/__tests__/UsersTableRow.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    __text: '192.168.1.1',
+    _allow: '0',
+  },
+});
+
+const gmp = {
+  session: createSession({username: 'admin'}),
+};
+
+describe('users Row', () => {
```
bjoernricks:
```suggestion
describe('UsersTableRow', () => {
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,213 @@
...
+import useGmp from 'web/hooks/useGmp';
+import useTranslation from 'web/hooks/useTranslation';
+import UserDialog, {type UserDialogSaveData} from 'web/pages/users/UsersDialog';
+
+interface UserComponentRenderProps {
+  clone: (entity: User) => Promise<void>;
+  create: (user?: User) => Promise<void>;
+  delete: (entity: User) => Promise<void>;
+  download: (entity: User) => Promise<void>;
+  edit: (user?: User) => Promise<void>;
```
bjoernricks:
```suggestion
  edit: (user: User) => Promise<void>;
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,213 @@
...
+  save?: (data: UserDialogSaveData) => Promise<void | unknown>;
```
bjoernricks: IMHO the return type should also be void for save.
```suggestion
  save: (data: UserDialogSaveData) => Promise<void>;
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,213 @@
...
+}
+
+interface UserComponentProps {
+  children: (props: UserComponentRenderProps) => ReactNode;
+  onCloneError?: (error: Error) => void;
+  onCloned?: (response: unknown) => void;
```
bjoernricks: I think it should be
```suggestion
  onCloned?: (response: Response<EntityActionData>) => void;
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,213 @@
...
+  onCreateError?: (error: Error) => void;
+  onCreated?: (response: unknown) => void;
```
bjoernricks:
```suggestion
  onCreated?: (response: Response<EntityActionData>) => void;
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,213 @@
...
+  onDeleteError?: (error: Error) => void;
+  onDeleted?: () => void;
+  onDownloadError?: (error: Error) => void;
+  onDownloaded?: OnDownloadedFunc;
+  onSaveError?: (error: Error) => void;
+  onSaved?: (response: unknown) => void;
```
bjoernricks: IMHO it should be
```suggestion
  onSaved?: () => void;
```

src/web/pages/users/UserDetails.tsx
```diff
@@ -72,6 +84,9 @@ const UserDetails = ({entity, links = true}) => {
             <TableData>
               <HorizontalSep>
                 {roles.map(role => {
+                  if (!role.id) {
```
bjoernricks: I don't think the can ever happen because the ID is required and available for all models returned from queries now.

src/web/pages/users/UserDetails.tsx
```diff
@@ -89,6 +104,9 @@ const UserDetails = ({entity, links = true}) => {
             <TableData>
               <HorizontalSep>
                 {groups.map(group => {
+                  if (!group.id) {
```
bjoernricks: Same here. ID should always be defined.

src/web/pages/users/UserDetailsPage.tsx
```diff
@@ -0,0 +1,252 @@
...
+  onUserEditClick: (user: User) => void | Promise<void>;
+}
+
+const permissionsSubjectFilter = (id: string): FilterType =>
+  QueryFilter.fromString(
+    'subject_uuid=' +
+      id +
+      ' and not resource_uuid=""' +
+      ' or resource_uuid=' +
+      id,
```
bjoernricks: Could be converted to a template literal for better readability.

src/web/pages/users/UserDetailsPage.tsx
```diff
@@ -0,0 +1,252 @@
...
+        entityType="user"
+        isLoading={isLoading}
+        sectionIcon={<UserIcon size="large" />}
+        title={_('User')}
+        toolBarIcons={<div />}
+      >
+        {() => null}
+      </EntityPage>
+    );
+  }
```
bjoernricks: Hmm how did we handle this case before? 🤔 looks a bit weird to me.

src/web/pages/users/UserDetailsPage.tsx
```diff
@@ -0,0 +1,252 @@
...
+  return (
+    <>
+      <DialogNotification {...dialogState} onCloseClick={closeDialog} />
+      <Download ref={downloadRef} />
+      <UserComponent
+        onCloneError={onError}
+        onCloned={(response: unknown) => {
+          const id = (response as {id: string}).id;
+          goToDetails('user', navigate)({data: {id}});
+        }}
```
bjoernricks: Can we fix this? Either it should be a `Response<EntityActionData>` or just `EntityActionData`

src/web/pages/users/UserDetailsPage.tsx
```diff
@@ -0,0 +1,252 @@
...
+        onCreated={(response: unknown) => {
+          const id = (response as {id: string}).id;
+          goToDetails('user', navigate)({data: {id}});
+        }}
```
bjoernricks: Same as above. IMHO we really should unify what's returned from create and save.

src/web/pages/users/UserDetailsPage.tsx
```diff
@@ -0,0 +1,252 @@
...
+              return (
+                <>
+                  <PageTitle
+                    title={_('User: {{name}}', {name: entity.name as string})}
+                  />
+                  <TabsContainer flex="column" grow="1">
+                    <TabLayout align={['start', 'end']} grow="1">
+                      <TabList align={['start', 'stretch']}>
+                        {tabs.map(({label, entities}) =>
+                          entities === undefined ? (
```
bjoernricks:
```suggestion
                          !isDefined(entities) ? (
```

src/web/pages/users/UsersDialog.tsx
```diff
@@ -262,7 +314,7 @@ const Dialog = ({
               )}
             </FormGroup>
           )}
-          {capabilities.mayAccess('roles') && (
+          {capabilities.mayAccess('roles' as CapabilitiesEntityType) && (
```
bjoernricks:
```suggestion
          {capabilities.mayAccess('role') && (
```

src/web/pages/users/UsersTableHeader.tsx
```diff
@@ -1,79 +1,90 @@
...
 import TableRow from 'web/components/table/TableRow';
-import withEntitiesHeader from 'web/entities/withEntitiesHeader';
 import useTranslation from 'web/hooks/useTranslation';
-import PropTypes from 'web/utils/PropTypes';
+import SelectionType, {type SelectionTypeType} from 'web/utils/SelectionType';
+import {type SortDirectionType} from 'web/utils/sort-direction';
 
-const Header = ({
+interface UsersHeaderProps {
+  actionsColumn?: ReactElement | null;
```
bjoernricks, line 15: Maybe (I am not sure about this)
```suggestion
  actionsColumn?: ReactNode;
```

src/web/pages/users/UsersHeader.tsx
```diff
@@ -1,79 +1,90 @@
...
+    );
+
   return (
     <TableHeader>
       <TableRow>
         <TableHead
           currentSortBy={currentSortBy}
           currentSortDir={currentSortDir}
-          sortBy={sort ? 'name' : false}
+          sortBy={sort ? 'name' : undefined}
```
bjoernricks:
```suggestion
          sortBy="name"
          sort={sort}
```

src/web/pages/users/UsersHeader.tsx
```diff
@@ -1,79 +1,90 @@
...
           title={_('Name')}
           width="20%"
           onSortChange={onSortChange}
         />
         <TableHead
           currentSortBy={currentSortBy}
           currentSortDir={currentSortDir}
-          sortBy={sort ? 'roles' : false}
+          sortBy={sort ? 'roles' : undefined}
```
bjoernricks: same as above

src/web/pages/users/UsersHeader.tsx
```diff
@@ -1,79 +1,90 @@
...
           title={_('Roles')}
           width="26%"
           onSortChange={onSortChange}
         />
         <TableHead
           currentSortBy={currentSortBy}
           currentSortDir={currentSortDir}
-          sortBy={sort ? 'groups' : false}
+          sortBy={sort ? 'groups' : undefined}
```
bjoernricks: same as above

src/web/pages/users/UsersHeader.tsx
```diff
@@ -1,79 +1,90 @@
...
           title={_('Groups')}
           width="26%"
           onSortChange={onSortChange}
         />
         <TableHead
           currentSortBy={currentSortBy}
           currentSortDir={currentSortDir}
-          sortBy={sort ? 'host_access' : false}
+          sortBy={sort ? 'host_access' : undefined}
```
bjoernricks: same as above

src/web/pages/users/UsersHeader.tsx
```diff
@@ -1,79 +1,90 @@
...
           title={_('Host Access')}
           width="10%"
           onSortChange={onSortChange}
         />
         <TableHead
           currentSortBy={currentSortBy}
           currentSortDir={currentSortDir}
-          sortBy={sort ? 'ldap' : false}
+          sortBy={sort ? 'ldap' : undefined}
```
bjoernricks: same as above

src/web/pages/users/UsersTableRow.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <DetailsLink key={group.id} id={group.id} textOnly={!links} type="group">
+        {group.name}
+      </DetailsLink>
+    ) : (
+      <span key={group.name}>{group.name}</span>
+    ),
+  );
+
+  const authMethod = convertAuthMethod(entity.authMethod, _);
+  const host_allow = convertAllow(
```
bjoernricks:
```suggestion
  const hostAllow = convertAllow(
```

src/web/pages/users/UsersTableRow.tsx
```diff
@@ -0,0 +1,168 @@
...
+        type="user"
+        onToggleDetailsClick={onToggleDetailsClick}
+      />
+      <TableData>
+        <HorizontalSeparator $wrap>{roles}</HorizontalSeparator>
+      </TableData>
+      <TableData>
+        <HorizontalSeparator $wrap>{groups}</HorizontalSeparator>
+      </TableData>
+      <TableData>{host_allow}</TableData>
```
bjoernricks:
```suggestion
      <TableData>{hostAllow}</TableData>
```

src/web/pages/users/UserDetailsPage.tsx
```diff
@@ -0,0 +1,252 @@
...
+const permissionsSubjectFilter = (id: string): FilterType =>
+  QueryFilter.fromString(
+    'subject_uuid=' +
+      id +
+      ' and not resource_uuid=""' +
+      ' or resource_uuid=' +
+      id,
+  ).all();
+
+const ToolBarIcons = ({
```
bjoernricks:
```suggestion
const UsersDetailsPageToolBarIcons = ({
```

src/web/pages/users/UsersListPage.tsx
```diff
@@ -0,0 +1,454 @@
...
+  selectionType: SelectionTypeType;
+  onChanged: () => void;
+  onDeleteBulk: () => void | Promise<void>;
+  onDownloadBulk: () => Promise<void>;
+  onDownloaded: OnDownloadedFunc;
+  onEntityDeselected: (entity: User) => void;
+  onEntitySelected: (entity: User) => void;
+  onError: (error: Error | Rejection) => void;
+  onFilterChanged: (newFilter: FilterType) => void;
+  onFilterCreated: (newFilter: FilterType) => void;
```
bjoernricks: When a filter is created you'll get a Filter model instance.
```suggestion
  onFilterCreated: (newFilter: Filter) => void;
```

src/web/pages/users/UsersListPage.tsx
```diff
@@ -0,0 +1,454 @@
...
+      onDownloadError={handleError}
+      onDownloaded={handleDownload}
+      onSaved={onChanged}
+    >
+      {({clone, create, download, edit, save}) => (
+        <>
+          <DialogNotification {...dialogState} onCloseClick={closeDialog} />
+          <Download ref={downloadRef} />
+          <PageTitle title={_('Users')} />
+          <EntitiesPage<User, UsersEntitiesPageProps>
```
bjoernricks: Would be nice to render the toolbaricons and table in this component directly and pass the elements into the entitiespage. See for example the <https://github.com/greenbone/gsa/blob/main/src/web/pages/overrides/OverrideListPage.tsx#L90-L114>. This allows to reduce the number of props passed to the EntitesPage (a lot of them are currently just forwarded). This improves the readability a lot.

src/web/pages/users/UserDetailsPage.tsx
daniele-mng: If we do not have this condition we get an error on first load for the details page

src/web/pages/users/UsersTableHeader.tsx
daniele-mng: It's more specific and narrow with `ReactElement`.

src/web/pages/users/UserDetailsPage.tsx
bjoernricks: Yes i understand the code. I am just not sure how we handle this case at the other pages currently. Are we only rendering a details page if something is loaded or an error occurred?

src/web/pages/users/__tests__/UsersTableHeader.test.tsx
bjoernricks, line 1: Should be named `UsersTableHeader` now too

src/web/pages/users/__tests__/UsersHeader.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import {describe, expect, test, testing} from '@gsa/testing';
+import {rendererWith, screen} from 'web/testing';
+import UsersHeader from 'web/pages/users/UsersTableHeader';
+import SelectionType from 'web/utils/selection-type';
+
+describe('UsersHeader', () => {
```
bjoernricks:
```suggestion
describe('UsersTableHeader', () => {
```

src/web/hooks/use-query/users.ts
```diff
@@ -0,0 +1,193 @@
...
+  id: string;
+  refetchInterval?: RefetchIntervalFn<User>;
+}
+
+interface UseUserMutationCallbacks<TResponse> {
+  onSuccess?: (response: TResponse) => void;
+  onError?: (error: Error) => void;
+}
+
+type UserIdsInput = string | string[];
```
bjoernricks: Would be good to rename the type because it's not only used for user ids.
```suggestion
type IdsInput = string | string[];
```

src/web/hooks/use-query/users.ts
```diff
@@ -0,0 +1,193 @@
...
+
+interface UserCreateInput {
+  access_hosts: string;
+  auth_method: string;
+  comment: string;
+  group_ids: UserIdsInput;
```
bjoernricks:
```suggestion
  group_ids: IdsInput;
```

src/web/hooks/use-query/users.ts
```diff
@@ -0,0 +1,193 @@
...
+  hosts_allow: string;
+  name: string;
+  password: string;
+  role_ids: UserIdsInput;
```
bjoernricks:
```suggestion
  role_ids: IdsInput;
```

src/web/hooks/use-query/users.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+interface BulkDeleteUsersInput {
+  users: User[];
+  options?: {inheritor_id?: string};
+}
+
+export type UserBulkInput = User[] | FilterType;
+
+const toUserIdsArgument = (value: UserIdsInput): string =>
```
bjoernricks: Reading through this code again, I think the `gmp.user.create` inputs are badly designed at the moment and it should be changed to accept only an array for group and role ids. but that's something for another pr.
```suggestion
const toIdsArgument = (value: IdsInput): string =>
```

src/web/hooks/use-query/users.ts
```diff
@@ -0,0 +1,193 @@
...
+  onSuccess,
+  onError,
+}: UseUserMutationCallbacks<EntityActionData> = {}) => {
+  const gmp = useGmp();
+  return useGmpMutation<UserCreateInput, EntityActionData>({
+    gmpMethod: async data => {
+      const response = await gmp.user.create({
+        ...data,
+        group_ids: toUserIdsArgument(data.group_ids),
+        role_ids: toUserIdsArgument(data.role_ids),
```
bjoernricks:
```suggestion
        group_ids: toIdsArgument(data.group_ids),
        role_ids: toIdsArgument(data.role_ids),
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,219 @@
...
+  useDeleteUser,
+  useSaveUser,
+} from 'web/hooks/use-query/users';
+import useGmp from 'web/hooks/useGmp';
+import useTranslation from 'web/hooks/useTranslation';
+import UserDialog, {type UserDialogSaveData} from 'web/pages/users/UsersDialog';
+
+interface UserComponentRenderProps {
+  clone: (entity: User) => Promise<void>;
+  create: (user?: User) => Promise<void>;
```
bjoernricks: I think it should be
```suggestion
  create: () => Promise<void>;
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const saveMutation = useSaveUser({
+    onSuccess: onSaved,
+    onError: onSaveError,
+  });
+  const cloneMutation = useCloneUser({
+    onSuccess: onCloned,
+    onError: onCloneError,
+  });
+  const deleteMutation = useDeleteUser({
+    onSuccess: () => onDeleted?.(),
```
bjoernricks: Any reason for not using this?
```suggestion
    onSuccess: onDeleted,
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,219 @@
...
+      setRoles(rolesResponse.data);
+
+      const settings = authSettingsResponse.data;
+      setSettings(settings);
+      setDialogVisible(true);
+
+      if (isDefined(user)) {
+        const newGroupIds = user.groups
+          .map(group => group.id)
+          .filter((id): id is string => isDefined(id));
```
bjoernricks: Should not be necessary. `group.id` should be always available now with my model changes.
```suggestion
          .map(group => group.id);
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,219 @@
...
+        const newRoleIds = user.roles
+          .map(role => role.id)
+          .filter((id): id is string => isDefined(id));
```
bjoernricks: Same. `role.id` should be always available.
```suggestion
          .map(role => role.id);
```

src/web/pages/users/UserComponent.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+        setAccessHosts(user.hosts?.addresses.join(', ') ?? '');
```
bjoernricks, line 132: Looks wrong to me too. Maybe we should change this to an multi select field in the dialog in future.

src/web/pages/users/UsersDialog.tsx
```diff
@@ -128,33 +176,37 @@ const Dialog = ({
...
-  const hasRadiusEnabled = settings.get('method:radius_connect').enabled;
+  const hasLdapEnabled =
+    (settings.get('method:ldap_connect') as AuthMethodSetting).enabled === true;
+  const hasRadiusEnabled =
+    (settings.get('method:radius_connect') as AuthMethodSetting).enabled ===
+    true;
 
   return (
-    <SaveDialog
+    <SaveDialog<ControlledDialogValues, UserDialogSaveData>
```
bjoernricks: I don't think `UserDialogSaveData` is correct here. It should only contain properties whose state is maintained within the SaveDialog (aka. uncontrolled values or initial values or default values). `UserDialogSaveData` is then the combination of the controlled (data whose state is handled within the UsersDialog) and uncontrolled values.

src/web/pages/users/UsersDialog.tsx
```diff
@@ -273,7 +325,7 @@ const Dialog = ({
             </FormGroup>
           )}
 
-          {capabilities.mayAccess('groups') && (
+          {capabilities.mayAccess('groups' as CapabilitiesEntityType) && (
```
bjoernricks:
```suggestion
          {capabilities.mayAccess('group') && (
```

src/web/pages/users/UsersTableRow.tsx
```diff
@@ -0,0 +1,168 @@
...
+    ) : (
+      <span key={group.name}>{group.name}</span>
+    ),
+  );
+
+  const authMethod = convertAuthMethod(entity.authMethod, _);
+  const hostAllow = convertAllow(entity.hosts ?? {addresses: []}, _).replaceAll(
+    '&#x2F;',
+    '/',
+  );
```
bjoernricks: Should we maybe put this replaceAll cleanup into the model class in the fromElement code as it is done at several places in the UI?

src/web/pages/users/UserComponent.tsx
bjoernricks: Added a ticket for this https://jira.greenbone.net/browse/GEA-1973

bjoernricks (approved):

PR #5458, opened 2026-07-29
Change SaveDialog state handling

bjoernricks:
## What

Change `SaveDialog` state handling

* Refactor internal state handling of `SaveDialog`
* Improve type information for passed `SaveDialog` values and change handler
* Fix a bug in the `ScanConfigEditDialog` for selecting the NVT family trends and all NVT families
* Add tests for the `SaveDialog`

## Why

Get rid of the `State` utility component that was only used in SaveDialog and predates react hooks.

## Checklist

- [x] Tests


src/web/utils/State.jsx
daniele-mng, line 1: 🚀

daniele-mng (approved):

PR #5460, opened 2026-07-30
Add a date picker next to the days field of an override

TheBeast85:
The *Active* group of the note and override dialogs offers a date picker and works the number of days out from it, but never shows the number. The days are what actually goes to the manager, and someone who wants to say "thirty days" has to reach for a calendar to find the date, or count.

This shows both. The row now reads **"yes, for the next `[30]` days"** with a calendar icon beside the field. Typing a number moves the calendar, and picking a day in the calendar fills in the number of days from today — so the value you send is visible and editable, and you no longer have to work it out by hand.

The label changed from *"yes, until"* to *"yes, for the next"*, because with a number of days in front of it the old one read wrong ("yes, until 30 days").

## What changed

- **`DaysDatePicker`** (new) is the calendar icon and the calendar it unfolds: a Mantine `Popover` around the `DatePicker` from `@mantine/dates`. `Popover.Target` needs a child it can attach a ref to and the icon is a plain function component, so it is wrapped in a `span`. The transition is off, so the calendar is there as soon as it opens rather than fading in.
- **`ActiveFormGroup`** gains a `days` prop and `computeEndDate`, the counterpart of the existing `computeDaysUntil`, so the field and the calendar always agree in both directions. Both dialogs pass their `days` through.

Notes get the same, because `NoteDialog` and `OverrideDialog` share the component.

## Tested

19 cases for `ActiveFormGroup`, including a real click in the calendar against a frozen clock (1 March 2026, clicking the 21st writes `20` into the field), that the calendar stays shut while *yes, for the next* is not selected, and that `computeEndDate` and `computeDaysUntil` are inverses.

End to end against the manager: 20 days produces an end time exactly 20 days ahead and reads back as such.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/web/components/form/DaysDatePicker.tsx
+++ b/src/web/components/form/DaysDatePicker.tsx
@@ -1,0 +1,96 @@
+/* SPDX-FileCopyrightText: 2026 Greenbone AG
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import {useCallback, useState} from 'react';
+import {Popover} from '@mantine/core';
+import {DatePicker} from '@mantine/dates';
+import date, {type Date} from 'gmp/models/date';
+import {isDefined} from 'gmp/utils/identity';
+import {CalendarIcon} from 'web/components/icon';
+import useLanguage from 'web/hooks/useLanguage';
+import useTranslation from 'web/hooks/useTranslation';
+
+interface DaysDatePickerProps {
+  disabled?: boolean;
+  maxDate?: Date;
+  minDate?: Date;
+  name?: string;
+  title?: string;
+  value?: Date;
+  onChange?: (value: Date, name?: string) => void;
+}
+
+/**
+ * A calendar icon that unfolds a calendar.
+ *
+ * Sits next to a field that holds a number of days and exists so that the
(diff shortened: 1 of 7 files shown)


bjoernricks:
Not sure what you want to solve here. These AI PR descriptions are horrible to read. Please just follow our PR template and write why you are changing something. IMHO this PR was already solved by https://github.com/greenbone/gsa/pull/5422

TheBeast85:
Sorry about the description, I will keep it to the template.

On #5422: it replaced the days field with the date picker, so the number itself is no longer shown anywhere. That number is what is sent as `<active>N</active>`, and right now it can neither be read back nor typed - a duration can only be expressed by finding the matching date in the calendar.

What this change does is keep the date picker and put the number next to it, kept in sync both ways, so either way works: type 30 directly, or pick a date and the days are filled in for you.

Would a PR along those lines be welcome? If so I will open it against the template.

bjoernricks:
I am open for a new PR that shows and allows to use both a date picker and a number input field. But I don't want to get a completely new form component like the `DaysDatePicker`. If you write a component that combines some existing form components that's fine but please don't invent a new one from scratch for a single purpose.

PR #5496, opened 2026-08-06
TypeScript for additional Dashboards components

bjoernricks:
## What

TypeScript for additional Dashboards components

* TypeScript
* Tests
* Class Components => Function Components

## Why

Use TypeScript for more and more dashboard components to ensure their behavior and expected props.
Use modern react to get rid of HOCs and use Hooks instead.

## References

https://jira.greenbone.net/browse/GEA-1997

## Checklist

- [x] Tests


src/web/components/dashboard/display/DataTable.tsx
```diff
@@ -38,7 +47,7 @@ const DataTable = ({dataTitles = [], data = [], dataRow: rowFunc}) => (
           return (
             <TableRow key={i}>
```
daniele-mng, line 48: Nitpick: Not part of the changes, but can we use a better key instead of index?

bjoernricks: Sadly I don't think so. We don't have a unique column for the data. Also it's not generic so we don't know the data shape in the DataTable component.

daniele-mng (approved):

PR #5520, opened 2026-08-14
TypeScript Dashboards

bjoernricks:
## What
Use TypeScript for Dashboard components and add tests for all of them

## Why

Ensure that the correct properties are passed to the components and ensure their behavior.

## References

https://jira.greenbone.net/browse/GEA-1997

## Checklist

- [x] Tests


src/web/components/dashboard/display/DataDisplay.tsx
```diff
@@ -59,14 +59,19 @@ interface DataDisplayRenderProps<TData, TState extends State> {
 type TransformFunc<
   TData,
   TTransformedData,
-  TTransformProps = Record<string, unknown>,
+  TTransformProps extends object = object,
```
daniele-mng, line 62: Q: is there a reason why we are loosening the typing here?  I see many `object` introduced.

src/web/components/dashboard/display/cvss/cvss-transform.ts
```diff
@@ -0,0 +1,183 @@
...
+  const transformedData = Object.keys(cvssData)
+    .sort((a, b) => {
+      return Number(a) - Number(b);
+    })
+    .map(key => {
+      const count = cvssData[Number(key)];
+      const percentValue = percent(count, sum);
+
+      const value = parseFloat(key) as number;
+
```
daniele-mng:
```suggestion
      const value = Number(key);

```

src/web/components/dashboard/display/severity/SeverityClassDisplay.tsx
```diff
@@ -40,7 +41,7 @@ interface SeverityClassDataDisplayProps extends DataDisplayProps<
   SeverityClassData,
   TransformSeverityDataProps
 > {
```
daniele-mng: Can we reuse this with `SeverityClassDisplayProps` they partially overlap.

src/web/components/dashboard/display/cvss/CvssDisplay.tsx
```diff
@@ -3,16 +3,47 @@
...
+  type DataDisplayProps,
+  type State,
+} from 'web/components/dashboard/display/DataDisplay';
 import useGmp from 'web/hooks/useGmp';
 import useTranslation from 'web/hooks/useTranslation';
-import PropTypes from 'web/utils/prop-types';
+
+type CvssDisplayState = State;
+
+interface CvssDisplayProps extends DataDisplayProps<
```
daniele-mng: CvssDisplayProps and CvssDataDisplayProps partially overlap.

src/web/pages/vulnerabilities/dashboard/hosts-transform.ts
```diff
@@ -0,0 +1,103 @@
...
+      (prev, current) => prev + (parseFloat(current.count) ?? 0),
+      0,
+    );
+    const yValue = sumOfBinMembers;
+    const percentValue = percent(yValue, totalVulns);
+    const filterValue = {start: min, end: max};
+    return {
+      x: binWidth > 1 ? `${min}-${max}` : String(min),
+      y: yValue,
+      label: 'label',
```
daniele-mng, line 94: Is this correct returning a string `label`?

src/web/components/dashboard/display/cvss/cvss-transform.ts
bjoernricks: ok that code is a bit strange.

src/web/components/dashboard/display/DataDisplay.tsx
bjoernricks: AI ... sadly I didn't question this change

bjoernricks: I am not sure about the best shape here. At least using `TTransformProps extends Record<string, unknown>` is not a good choice.

With that change we get several errors like
```
Type 'TransformSeverityDataProps' does not satisfy the constraint 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type 'TransformSeverityDataProps'.
```

bjoernricks: I played a bit around with this code and came to the conclusion `T extends object = object` is what we want here. `T extends object = Record<string, unknown>` would also be ok but doesn't make a real difference. `T extends Record<string, unknown>` is wrong because that requires all `T` to have an index signature which we don't want.

src/web/pages/vulnerabilities/dashboard/hosts-transform.ts
bjoernricks: At least i didn't change it. It was `label: 'label',` before already. let me check. maybe it's not used at all.

bjoernricks: Ok i think it's unused but currently required because of depending on `LegendData`. This will be addressed in another PR which i am already working on locally.

daniele-mng (approved):

PR #5541, opened 2026-08-18
Change: allow editing a target that a task refers to

TheBeast85:
## What

Drop the `inUse` handling from the target dialog. The fields are no longer disabled, and `handleSaveClick` no longer reduces the request to name, comment and alive tests.

## Why

Once a task referred to a target, the dialog disabled hosts, port list and every credential, so the scan target of a task could not be corrected here. That mirrored a restriction in gvmd which greenbone/gvmd#3098 removes; with it gone the dialog no longer has to anticipate what gvmd will accept, and any remaining error comes back from gvmd.

## Testing

The existing target tests, with the case that asserted the fields are disabled for a target in use turned around to assert they stay editable.

## References

Needs greenbone/gvmd#3098.


bjoernricks:
Hi, this PR will not be applied. The in use feature is an important concept in our software and removing it has several consequences. Therefore I'll close this PR.

@TheBeast85 please create issues for such big changes first and discuss the possibility of getting the desired changes into the software before creating a PR.

PR #5544, opened 2026-08-19
Change: allow editing a scan config in use, and do not deselect every NVT when the families are not loaded

TheBeast85:
## What

- Fix: build `trend`/`select` only once the config **and** the family list have arrived, and leave the family part out of the request while the selection is unknown
- Change: show the full edit dialog for a scan config a task refers to, and for a policy an audit refers to, instead of the "only name and comment" notice

## Why

The edit dialog fetches the config and the list of NVT families in parallel and builds `trend`/`select` once, as soon as the config has arrived. When the config arrives first, `families` is still undefined, `createTrendAndSelect` returns empty objects, and the guard keeps them that way. Saving then sends only the bare `trend` parameter, gsad turns that into a `family_selection` element without a single `family`, and gvmd reads that as "no family selected" and clears every fully selected family of the config. GEA-1698 fixed the case where the parent dialog held a stale value for one family; this is the case where it holds no family at all.

The in-use notice replaced families, scanner preferences and NVT preferences, and the component cut the request down to name, comment and id. gvmd no longer refuses those changes (greenbone/gvmd#3099), so the dialog can show the full form.

## References

GEA-1698, greenbone/gvmd#3099

## Checklist

- [x] Tests


bjoernricks:
Please split such things into two different PRs. As I mentioned in your other PRs the "in use" changes have no chance of being applied at the moment. This is nothing we are going to change anytime soon. It has a serious use case.

If you want the fix for the trend/select stuff merged feel free to adapt this PR and force push or to close it and create a new PR.

bjoernricks:
@TheBeast85 The `allow editing a scan config in use` has no chance of being merged. Please read my comment above.

TheBeast85:
Thanks for the clear answer, and sorry for mixing two unrelated things into one PR.

I have split them as you suggested: the trend/select fix is now on its own in #5574, on top of current main and without any "in use" change. I am dropping the scan config "in use" part and closing this PR — understood that this is a deliberate concept and not something to change from the outside.

If it would still be useful to describe the underlying issue, I am happy to open an issue for it first.

PR #5574, opened 2026-09-01
Fix: do not deselect every NVT when the families are not loaded yet

TheBeast85:
## What

- Build `trend` and `select` in `ScanConfigEditDialog` only once the scan config **and** the NVT family list have both arrived.
- Leave `select`, `trend` and `familyTrend` out of the save request while that selection is still unknown.

## Why

Both requests run in parallel. If the config arrives first, `families` is still undefined, the selection is built empty and never rebuilt. Saving then sends a family selection that covers no family at all, which gvmd reads as "deselect everything", so every fully selected family of the config is cleared.

## References

Split out of #5544, which mixed this fix with an unrelated change.

## Checklist

- [x] Tests

### DIFF
--- a/src/web/pages/scanconfigs/ScanConfigEditDialog.tsx
+++ b/src/web/pages/scanconfigs/ScanConfigEditDialog.tsx
@@ -296,8 +296,16 @@ const ScanConfigEditDialog = ({
     );
   }, [familySelectionUpdate]);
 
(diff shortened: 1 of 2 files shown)


bjoernricks:
Thanks for your PR. While reviewing it I found additional related things to be improved and created https://github.com/greenbone/gsa/pull/5583 to fix your issue.

PR #5583, opened 2026-09-07
Fix scan config dialog for loading families

bjoernricks:
## What

Fix scan config dialog for loading families

## Why

Trend and selected families weren't updated when loading the families did take longer then loading the scan config when opening the dialog for editing a scan config. It was also possible to save partially loaded data.

## References
https://github.com/greenbone/gsa/pull/5574

## Checklist

- [x] Tests


src/web/components/dialog/DialogTwoButtonFooter.tsx
```diff
@@ -35,15 +35,15 @@ const DialogTwoButtonFooter = ({
     <DialogFooterLayout align={['space-between', 'center']} shrink="0">
       <Button
         data-testid="dialog-close-button"
-        disabled={isLoading}
+        disabled={isSaving}
```
ozgen, line 38: Previously loading disabled both buttons, but with `isLoading` the Cancel button stays enabled. Is this behavior change intentional?

ozgen (approved):

src/web/components/dialog/DialogTwoButtonFooter.tsx
bjoernricks: yes it is. When loading data the user should still be able to cancel the dialog. Maybe because he didn't wanted to open it initially but clicked on the wrong button. When saving it's already to late for cancelling so both buttons are disabled.

PR #5623, opened 2026-09-15
Improve and extend linter rules

bjoernricks:
## What

Improve and extend oxlint rules:

* Enable eslint oxlint plugin
* Raise warning if console.log is used
* Set a consistent style for the type imports (and adapt code for that style)
* Require `test` function in unittests

## Why

Improve code quality by defining and extending standards.


src/web/pages/tags/__tests__/TagComponent.test.tsx
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+// oxlint-disable typescript/no-useless-default-assignment
```
daniele-mng, line 6: Is this comment needed?

bjoernricks: yes sadly. Removing the undefined will cause the type-checker to complain. So either I add a full typed version of the gmp thing or we add the comment. I did choose the comment.

daniele-mng (approved):

